PR #4, opened 2017-06-24
run gc after each request

grunxen:
Run garbage collector after each server request for decreasing memory usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,7 @@
 Gemfile.lock
+
+.idea
+.vscode
+.yardoc
+doc
+coverage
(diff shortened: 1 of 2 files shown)


castwide:
Thanks for the contribution. Is memory usage a major issue on your setup?

The GC call seemed to have minimal impact for me (consistently less than 10% difference), but it didn't appear to have any detrimental side effects either, so I went ahead and merged it.

grunxen:
> Is memory usage a major issue on your setup?
> 
Yes. Without gc calls it could use ~800mb of memory, now ~250mb.

castwide:
Wow, that's huge. Thanks again.

PR #46, opened 2018-04-21
Use pure-ruby kramdown instead of redcarpet

grddev:
I wanted to try out solargraph (and vscode-ruby) in a JRuby project, but couldn't even install it as redcarpet requires native extensions which cannot be installed in JRuby.

This replaces the markdown generation with the kramdown gem, which is implemented in pure ruby, and thus should work for JRuby as well as MRI. I tried to figure out reasonable options to get a somewhat similar behavior to the redcarpet version, but there didn't seem to be an automated test for this, so I wasn't really sure what would be a good test case.

### DIFF
--- a/lib/solargraph/page.rb
+++ b/lib/solargraph/page.rb
@@ -1,21 +1,11 @@
 require 'ostruct'
 require 'tilt'
-require 'redcarpet'
+require 'kramdown'
 require 'htmlentities'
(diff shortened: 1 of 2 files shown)


castwide:
If it improves platform independence, I'm willing to give it a try. I'll set up some basic specs to help make sure it works as the library expects.

PR #48, opened 2018-04-22
Rescue Encoding errors during parsing to prevent crashing the whole service

Sam-Killgallon:
Solargraph kept crashing with 

```
EncodingError: invalid byte sequence in UTF-8
     # /Users/sam/.gem/ruby/2.5.1/gems/parser-2.5.1.0/lib/parser/source/buffer.rb:166:in `source='
     # ./lib/solargraph/source.rb:339:in `inner_parse'
     # ./lib/solargraph/source.rb:318:in `parse'
     # ./lib/solargraph/source.rb:60:in `initialize'
     # ./lib/solargraph/source.rb:410:in `new'
     # ./lib/solargraph/source.rb:410:in `load_string'
     # ./lib/solargraph/source.rb:405:in `load'
     # ./lib/solargraph/workspace.rb:126:in `block in load_sources'
     # ./lib/solargraph/workspace.rb:125:in `each'
```

When opening one of my projects. It turned out to be an invalid character in one of the comments which the parser is unable to handle https://github.com/whitequark/parser#invalid-characters-inside-comments-and-literals .

This ensures that it will keep parsing the remaining files and just skip the dodgy one, instead of crashing the whole service. 

I included a minimal fixture to reproduce the problem. The file is runnable by ruby, but won't work with the parser.


castwide:
Argh, this again!

The Source class uses a custom Builder (Solargraph::Source::FlawedBuilder) and enforces UTF-8 encoding to work around encoding errors. Apparently it still misses some cases. It might be possible to fix this one too, but either way, not crashing the service is a good thing.

castwide:
Thanks. I think it's good to merge for now. I might move the rescue from Workspace to Source, but the end goal of gracefully handling encoding errors (and the spec that covers it) should remain the same.

castwide:
@Sam-Killgallon I pushed some changes that correct for invalid byte sequences in UTF-8 before the parse by converting them to whitespace. This has the benefits of mapping the readable parts of the file and avoiding library errors when the offending file is opened in an editor. It works with the invalid_byte.rb file you provided for the specs. Just letting you know in case the change presents some other problem.

Sam-Killgallon:
Nice thanks. I've been running the version from master for a couple of hours and all seems good so far 👍

PR #57, opened 2018-05-14
Fix in-memory cache issue and add some nice minor utilities

douglascamata:
With this PR Solargraph does a complete parse of the codebase at its initialization. 

I was able to achieve almost instantaneous `Go To Definition` **on different constants** after using this feature for the first time.

It also kept working all when I moved definitions from file to file or removed some.

### DIFF
--- a/lib/solargraph/workspace/config.rb
+++ b/lib/solargraph/workspace/config.rb
@@ -13,7 +13,7 @@ module Solargraph
       def initialize workspace = nil
         @workspace = workspace
         include_globs = ['**/*.rb']
-        exclude_globs = ['spec/**/*', 'test/**/*', 'vendor/**/*']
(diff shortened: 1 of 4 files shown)


lib/solargraph/workspace/config.rb
```diff
@@ -13,7 +13,7 @@ class Config
       def initialize workspace = nil
         @workspace = workspace
         include_globs = ['**/*.rb']
-        exclude_globs = ['spec/**/*', 'test/**/*', 'vendor/**/*']
+        exclude_globs = ['spec/**/*', 'test/**/*', 'vendor/**/*', '.bundle/**/*']
```
douglascamata, line 16: This is a bit of a personal touch. WDYT, @castwide?

castwide: Looks good to me. As a general rule, Solargraph should exclude folders that contain gem caches.

Issue #14, opened 2017-11-05
Extensions


castwide:
Allow extensions to add their own suggestions to completion requests.

The master branch currently contains an experimental architecture for registering extensions through LiveMap.

The [solargraph-rails-ext](https://github.com/castwide/solargraph-rails-ext) extension uses this architecture to augment method suggestions with methods defined at runtime.

castwide:
Version 0.15.0 includes a new plugin architecture. The solargraph-rails-ext plugin is currently deprecated. The gem ships with a Runtime plugin that handles the Rails plugin's features in a more generic manner.

Plugins are considered experimental and prone to change.

castwide:
Extensions are on hold for the time being.

1. Many of the issues with missing core documentation that were previously solved by the `runtime` plugin have been solved by the new process for generating the docs (#155).
2. A lot of the functionality that plugins would have provided is now possible through other means, particularly YARD directives.

PR #76, opened 2018-07-25
straw man proposal for adding sourcegraph file extension support

warncke:
This PR has a working implementation of sourcegraph's files extension: https://sourcegraph.com/github.com/sourcegraph/language-server-protocol/-/blob/extension-files.md

The implementation method is to:
* enter "remote" mode using files extension if client advertises support
* set a global initialized flag on host and defer all requests until initialized
* defer some operations (e.g. Library api_map) until files are loaded
* modify Config to take list of files along with directory and load those files from the LSP client

Not dealt with:
* anything dealing with editing files

Pending discussion on whether or not this is something that could be merged upstream and whether or not the implementation approach makes sense for the code base I can modify and flesh out as needed.

### DIFF
--- a/lib/solargraph.rb
+++ b/lib/solargraph.rb
@@ -24,6 +24,7 @@ module Solargraph
 
   autoload :Shell,          'solargraph/shell'
   autoload :Source,         'solargraph/source'
+  autoload :SourceRemote,   'solargraph/source_remote'
   autoload :ApiMap,         'solargraph/api_map'
   autoload :YardMap,        'solargraph/yard_map'
   autoload :Pin,            'solargraph/pin'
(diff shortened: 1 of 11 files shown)


castwide:
Thanks for the PR. I'm open to supporting the files extension, although I might want to suggest some changes to the implementation. It's possible that I'll want to implement it as a Solargraph extension (castwide/solargraph#14), but that feature is currently incomplete and might not be robust enough for this kind of modification.

warncke:
Cool. I am definitely open to modifying the implementation as needed to fit
it into the project architecture in the best way.

castwide:
Closing as stale.

Issue #81, opened 2018-08-14
Pegged at 100% CPU


AndrewRayCode:
The solargraph server is pegged at 100% CPU always for me. I'm using the [`vscode-ruby`](https://github.com/rubyide/vscode-ruby) plugin

I also regularly get these errors spat out from the "Ruby Language Server":

```
[Error - 6:19:10 PM] Request textDocument/definition failed.
  Message: [TypeError] no implicit conversion of nil into String
  Code: -32603 
```

This is on a very large codebase, so it's probably impossible for me to narrow this down to a specific case.

blackst0ne:
My solagraph instance eats 100% CPU and a lot of RAM which makes it unusable.

castwide:
I'm in the middle of two sets of changes that should have a significant impact on performance and memory.

The first set of changes involve the ApiMap and are currently in the master branch:
* It avoids remapping the workspace for edits that do not change the API surface.
* Updates require less YARD parsing during runtime.
* The ApiMap caches more data, resulting in more simple key-value lookups vs. intensive queries.

The second set of changes involve the YardMap. Solargraph uses YARD docs to provide data from the Ruby core, the stdlib, and any gems required by the codebase. I've known for a while that YARD queries were a good candidate for optimization, but when I profiled them recently, they turned out to be even less efficient than I expected. I've figured out a few ways to make them faster, most notably by storing the pins generated by the YardMap directly in the ApiMap store. Testing against the rails/rails repo, completion items are noticeably faster and the server process uses nearly 70% less memory.

I expect to publish these updates in gem version 0.25.0.

@AndrewRayCode Do you get those errors after running Go to Definition? I can't think of any other reason the client would send a `textDocument/definition` method. Are there any more detailed error messages in the developer console?

castwide:
All the ApiMap and YardMap enhancements are pushed to master.

blackst0ne:
@castwide what's the ETA of the next release with these improvements? :slightly_smiling_face:

castwide:
@blackst0ne Most likely tomorrow.

castwide:
Gem version 0.25.0 is released. I tested it on 2 machines: a desktop with 32GB RAM and 8x4GHz CPU, and a laptop with 4 GB RAM and 2x2.6 GHz CPU. Both showed notable improvements in performance and memory usage.

[...]

PR #82, opened 2018-08-19
FIX: Request textDocument/hover failed

piktur:
Great news from #81! Looking forward to the improvement. In the meantime this PR attempts to ensure correct type for String addition:

VSCode log output
```
[Error] Request textDocument/definition failed.
  Message: [TypeError] no implicit conversion of nil into String
  Code: -32603 
```

and backtrace from VSCode DevTools
screen shot 2018-08-19 at 22 55 36

Cheers

### DIFF
--- a/lib/solargraph/source/fragment.rb
+++ b/lib/solargraph/source/fragment.rb
@@ -155,7 +155,7 @@ module Solargraph
       #
       # @return [String]
       def whole_word
-        @whole_word ||= word + remainder
+        @whole_word ||= word + remainder.to_s
       end
 
       # Get the whole signature at the current offset, including the final


castwide:
Thanks for tracking this one down. It'll be included in version 0.25.1.

castwide:
Writing a spec for this case, I noticed that similar problems can happen with a few other Fragment methods, so I added some more changes to ensure that `word`, `remainder`, `signature`, `base`, and `chain` (and methods that compose from them like `whole_word`) always return strings. Thanks again for the catch.

piktur:
Brilliant, thanks again for your work.

PR #92, opened 2018-10-08
Fix rspec deprecation warnings.

jonknapp:
This initializes the array to an object that `filename` can be called on. This will probably make the test run a bit slower since it is initializing data in the `Array`, though I'm not sure if there is a performance hit or not.

It could be replaced by making another `double` if desired, but I try to use as few of them as possible:

```
calculated = double(:calculated, each: [], length: Solargraph::Workspace::Config::MAX_FILES + 1)
config = double(:config, calculated: calculated, max_files: 0)
```

### DIFF
--- a/spec/library_spec.rb
+++ b/spec/library_spec.rb
@@ -101,7 +101,7 @@ describe Solargraph::Library do
     library.close file
     expect {
       library.checkout file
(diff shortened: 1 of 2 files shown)


castwide:
Looks good. Thanks a lot.

Issue #93, opened 2018-10-09
supporting rubocop's --force-exclusion option


jonknapp:
Rubocop has [an option called `--force-exclusion`](https://docs.rubocop.org/en/latest/basic_usage/#other-useful-command-line-flags) that will ignore any linter errors for files that a user's configuration has explicitly ignored. I'm not quite sure why it's not the default, but that's another story 😄 . I imagine someone _may_ enjoy being able to lint a file their config excludes for some reason.

I would either like to update the default Rubocop config to add the `--force-exclusion` flag or add a way to specify it through an extension setting and/or `.solargraph` config file. My two cents are to set it as a default configuration option so that adding another dot file to a project is a requirement to run rubocop with that project's config.

--------

I wanted to start the conversation as an issue but I'm more than willing to create some additional tests and a PR to add the support if desired.

castwide:
RuboCop assumes that if the file was passed as an explicit argument, as Solargraph does, the user expects it to be linted, regardless of the exclusion settings. I can see how that's not always the case, though. I'm open to a PR for this.

jonknapp:
Do you have a preference on "default" argument vs VS Code config vs solargraph config vs something else?

I would assume it depends if users of solargraph, rubocop, and/or a solargraph consumer (like the VS Code plugin) would use both versions of the flag and under what circumstances.

castwide:
I'm tempted to say --force-exclusion should not be the default. I'd prefer to get diagnostics for any file that I open in an editor, while the exclusion rules should apply when you're linting the entire workspace.

The .solargraph.yml file seems like a reasonable place for the option, since that's where the reporters are configured. I'm not sure what the convention should be, though. Maybe it could look something like this:

```
reporters:
- rubocop:
  - force-exclusion
```

If it's an option in both .solargraph.yml and the VS Code settings, I'd expect to the VS Code setting to take precedence.

jonknapp:
Yea, I tend to lean towards not making it the default as well.

Maybe a way to pass cli options with a single yaml config line? I'm not sure if there's a desire to tweak other things as well, but I've seen other libs take that approach. Something like:

```
reporters:
- rubocop:
  additional_cli_options: "--force-exclusion"
```

It would future proof things a bit at least. Not a fan of my long option name though.

castwide:
That could work. For a shorter name, we could call it `arguments`.

jonknapp:
[...]

PR #95, opened 2018-10-15
Allow configs for reporters; Rubocop can be passed additional args

jonknapp:
This allows solargraph reporters to have configurations passed to them from a `.solargraph` config file. They are passed a hash of the options, which should all be empty hashes except for Rubocop which can now take an `arguments` param for additional parameters to add to commands ran.

It allows support for https://github.com/castwide/solargraph/issues/93.

For example:

```
---
reporters:
- require_not_found
- rubocop:
    arguments:
    - --force-exclusion
```

would run Rubocop with the [`--force-exclusion` option](https://docs.rubocop.org/en/latest/basic_usage/#other-useful-command-line-flags). The `arguments` portion of the reporter config should be an array of strings.

### DIFF
--- a/LANGUAGE_SERVER.md
+++ b/LANGUAGE_SERVER.md
@@ -49,3 +49,21 @@ reporters:
 necessarily mean that the path is incorrect; only that Solargraph was unable to recognize it.
 
 Run `solargraph reporters` for a list of available reporters.
+
+### RuboCop Configuration
+
+Additional arguments can be passed to RuboCop commands by using a slightly
+different configuration. For example, the following will ignore reporting diagnostics
+on files that your RuboCop configuration excludes and only perform linting rules:
+
+```
+reporters:
+- rubocop:
+    arguments:
+    - --force-exclusion
+    - --lint
+- require_not_found
+```
+
+The json formatter and a path to the closest configuration file discovered are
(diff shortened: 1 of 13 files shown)


castwide:
Thanks for the quick work. Looks solid to me at first glance. My only concern is that passing arguments as a single string could lead to mangled escape sequences. That's the main reason for passing them as an array.

jonknapp:
It's not a problem to change `arguments` to be an array if you feel that will work better. Feel free to comment on anything else if you get some time to look it over again.

I'll add info to the README and update the code to handle an array of arguments instead of a string.

jonknapp:
I added some documentation and changed the `arguments` part of the reporter config to be an array of strings instead of a single string that we parse. If anything else is needed, let me know. Otherwise, I'm going to look at the VSCode plugin and see what needs to be done there to support this new feature.

castwide:
Thanks. I merged the PR into a new branch `rubocop-custom-arguments`, since there's a patch update I expect to release in the next few days. The next release will be 0.28.3, and I expect this PR to be released in 0.29.0.

The update to VSCode will probably require a change to pass client configuration data to the reporters, most likely in `LanguageServer::Host#diagnose`.

PiotrKolaczRM:
Will this be merged into `master`?

PR #100, opened 2018-10-25
Don't die if we have trouble eval-ing a gemspec

jhstatewide:


castwide:
Good call, thanks. I'd like to concoct a way to analyze the specification without evaluating the gemspec, but as long as it's necessary, this is a step in the right direction. I'll include this merge in the next patch release.

PR #101, opened 2018-10-28
We cannot trust that nil has undefined?

jhstatewide:
First can possibly be nil here. If so, we don't want to pass any messages to it, including undefined?

### DIFF
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -28,6 +28,7 @@ module Solargraph
     end
 
     def method_missing name, *args, &block
+      return if first == nil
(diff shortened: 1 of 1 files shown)


castwide:
Do you have an example where this happens? The elements in ComplexType should never be nil, so we might need to track down the root cause.

jhstatewide:
I am hitting this on one of my projects. I will see if I can create a minimal reproducible example for you.

castwide:
@jhstatewide Confirmed. Thanks a lot, that's super helpful.

I'll try to track down the root cause. In the meantime, I think we're good to merge your fix.

PR #132, opened 2019-01-02
Make SourceMap#query_symbols use fuzzy matching.

### LINKED ISSUE #44: Searching for symbols in workspace isn't using fuzzy matching like searching for symbols in current file
If I search for a symbol in the current file, then searching for `foobar` matches `foo_bar`, but searching for symbols across a workspace for `foobar` doesn't match `foo_bar` in another file, yet `foo_b` matches `foo_bar`.

Please enable fuzzy searching for symbols across workspaces too.

gmcabrita:
Fixes #44.

I made it fuzzy match either on the path or the name, since trying to fuzzy match a method name on a long path will typically fail.

### DIFF
--- a/lib/solargraph/source_map.rb
+++ b/lib/solargraph/source_map.rb
@@ -1,3 +1,5 @@
+require 'jaro_winkler'
+
 module Solargraph
   # An index of pins and other ApiMap-related data for a Source.
   #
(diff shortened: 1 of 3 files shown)


castwide:
Awesome, thanks! This looks like a good addition for v0.31.

PR #147, opened 2019-01-25
Update kramdown dependency to latest 2.0.0 release

mnussbaum:
:wave: In order to continue to install solargraph with my system package manager on Arch linux solargraph needs to be on the latest version of kramdown, as [ruby-kramdown](https://www.archlinux.org/packages/community/any/ruby-kramdown/) has already been updated to 2.0.0 and solargraph can no longer find an installed version it accepts.

Solargraph's tests all pass the same with the new kramdown version as with the old once the now-separate kramdown-parser-gfm gem was installed as well.


castwide:
This update breaks backwards compatibility with Ruby < 2.3. Based on my rough estimates, this will impact about 5% of Solargraph users. That doesn't mean we won't apply the change, especially since 2.2 is already EOL, but we'll need to provide some kind of notification, most likely in the README.

I'm going to do some research to see if compatibility back to 2.1 is still feasible. If not, we'll update the minimum Ruby requirement.

castwide:
This should no longer be an issue in v0.36.0. The `kramdown` gem has been replaced with `maruku`. If this raises any other problems, please let me know.

PR #151, opened 2019-01-30
Synchronized didChange

mattn:
When received completion after didchange immediately, completion got empty result since didchange call async_update.

### DIFF
--- a/lib/solargraph/language_server/host.rb
+++ b/lib/solargraph/language_server/host.rb
@@ -215,7 +215,7 @@ module Solargraph
       # @return [void]
       def change params
         updater = generate_updater(params)
-        sources.async_update params['textDocument']['uri'], updater
+        sources.update params['textDocument']['uri'], updater
       end
 
       # Queue a message to be sent to the client.


castwide:
Can you provide more information about the problem this solves, preferably with a reproducible example?

The async_update is necessary due to lag that can occur with large files and/or workspaces. Completion results should still (mostly) work while the host completes the synchronization in the background. If there are cases where it fails, we might be able to fix it somewhere in the async process instead of making it synchronous.

mattn:
Please try vim-lsp and solargraph.

https://github.com/prabirshrestha/vim-lsp/wiki/Servers-Ruby

vim-lsp flush to requests that must be sent before completion. For example, inserting some text before cmopletion, and type trigger key for completion immediately, vim-lsp send the didChange and completion. I mean that solargraph must NOT do async updating for didChange. If LSP client is heavy to update content, it is problem of clients. The client should send diffs instead of whole content.

castwide:
Async updates still apply the document change immediately. Only the rebuild of the API map is asynchronous. Using gVim on Windows 10, I was able to get the expected completion result:

image

(Note that I had to fix the escaping issue you reported in #150 first.)

mattn:
I'm contributor of vim-lsp. I got some reports that solargraph does not work. Most of them are caused by this reason. Completion request often overtake async update.

mattn:
Here is gif that completion overtake async update.
terminal6

mattn:
Hmm, it seems that solargraph still have another issue similar to this.

mattn:
Another gif.

terminal6

I'm trying this with following configuration.

```
function! OpenCompletion()
    if &omnifunc != '' && !pumvisible() && (v:char == '.' || v:char == '>')
        call feedkeys("\<C-x>\<C-o>", "n")
    endif
endfunction

autocmd InsertCharPre * call OpenCompletion()
```

castwide:
This is the configuration that works for me:

```
imap <c-space> <Plug>(asyncomplete_force_refresh)
inoremap <expr> <Tab> pumvisible() ? "\<C-n>" : "\<Tab>"
inoremap <expr> <S-Tab> pumvisible() ? "\<C-p>" : "\<S-Tab>"
inoremap <expr> <cr> pumvisible() ? "\<C-y>\<cr>" : "\<cr>"

if executable('solargraph')
    " gem install solargraph
    au User lsp_setup call lsp#register_server({
        \ 'name': 'solargraph',
        \ 'cmd': {server_info->[&shell, &shellcmdflag, 'D:\Users\fsnyd\Documents\code\solargraph-shim\latest.bat stdio']},
        \ 'initialization_options': {"diagnostics": "true"},
        \ 'whitelist': ['ruby'],
        \ })
endif
```

...but I think I may have found another issue.

You might not get completion results right away because Solargraph is still initializing the first time you try. However, that points to a possible secondary problem: initialization for a single file should only take 1-2 seconds. If your initialization time is significantly longer, it might be due to the `rootUri` parameter being passed to the `initialize` method.

Solargraph assumes that `rootUri` is a workspace and attempts to add all its .rb files to the map. If a `rootUri` is not specified, vim-lsp sets it to the user's home directory. Obviously, this can be extremely expensive.

If you want to start Solargraph without a workspace, you can initialize it with a null `rootUri` parameter. Unfortunately, vim-lsp will not allow it.

One possible workaround is to tell the client to look for a `.solargraph.yml` file, similar to how the [README's typescript example](https://github.com/prabirshrestha/vim-lsp#registering-servers) looks for `tsconfig.json`:

```
\ 'root_uri':{server_info->lsp#utils#path_to_uri(lsp#utils#find_nearest_parent_file_directory(lsp#utils#get_buffer_path(), '.solargraph.yml'))},
```

This will stop the server from inadvertently mapping the entire home directory, but it also means a .solargraph.yml is required for initialization.

A barebones .solargraph.yml file that initializes an empty workspace only needs this:

```
include: []
```

Open files will get mapped individually, but other files in the directory will be ignored.

Let me know if any of this helps or if you have any recommendations for how else to handle workspaces. If you still have problems that appear to be solved using update instead of async_update, we can also try making it an initialization option. Async has proven to be useful in other clients, so I don't want to disable it completely.

mattn:
I understand that there is a problem in vim-lsp that rootUri can not be set null, regardless of whether this issue is a bug or not. Thank you very much. I'll fix it.

castwide:
I found an issue with local variable detection that might explain why you had trouble completing `a.` in your examples. The fix is in the master branch.

mattn:
Thank you. I'm trying master branch but still have an issue. Could you handle this?
terminal6

castwide:
This is what it looks like for me.

diiv9cir8o

I waited a few seconds after opening the file to make sure the server was initialized. I'm also not sure what to make of the completion popup after I typed `b = [](enter)`. I didn't type anything else to request it. Maybe it was somehow left over from typing the `b`?

mattn:
Could you please try with disabling asyncomplete.vim ?

castwide:
This is the only configuration I can get to work at all:

```
call plug#begin('~/vimfiles/plugged')
Plug 'prabirshrestha/async.vim'
Plug 'prabirshrestha/vim-lsp'
Plug 'prabirshrestha/asyncomplete.vim'
Plug 'prabirshrestha/asyncomplete-lsp.vim'
call plug#end()

imap <c-space> <Plug>(asyncomplete_force_refresh)
inoremap <expr> <Tab> pumvisible() ? "\<C-n>" : "\<Tab>"
inoremap <expr> <S-Tab> pumvisible() ? "\<C-p>" : "\<S-Tab>"
inoremap <expr> <cr> pumvisible() ? "\<C-y>" : "\<cr>"

if executable('solargraph')
    " gem install solargraph
    au User lsp_setup call lsp#register_server({
        \ 'name': 'solargraph',
        \ 'cmd': {server_info->[&shell, &shellcmdflag, 'D:\Users\fsnyd\Documents\code\solargraph-shim\latest.bat stdio']},
        \ 'initialization_options': {"diagnostics": "true"},
        \ 'whitelist': ['ruby'],
		\ 'root_uri':{server_info->lsp#utils#path_to_uri(lsp#utils#find_nearest_parent_file_directory(lsp#utils#get_buffer_path(), '.solargraph.yml'))},
        \ })
endif
```

When I added `let g:asyncomplete_auto_popup = 0`, even `ctrl+space` stopped working.

Sorry if I'm misunderstanding anything. I'm not familiar with vim outside of basic editing in a shell terminal.

mattn:
Could you please try this.

1. put following file on somewhere.

`foo.rb`
```
a = []
a.
```

`test.vim`
```
set nocp
call plug#begin('~/vimfiles/plugged')
Plug 'prabirshrestha/async.vim'
Plug 'prabirshrestha/vim-lsp'
"Plug 'prabirshrestha/asyncomplete.vim'
"Plug 'prabirshrestha/asyncomplete-lsp.vim'
call plug#end()

"imap <c-space> <Plug>(asyncomplete_force_refresh)
"inoremap <expr> <Tab> pumvisible() ? "\<C-n>" : "\<Tab>"
"inoremap <expr> <S-Tab> pumvisible() ? "\<C-p>" : "\<S-Tab>"
"inoremap <expr> <cr> pumvisible() ? "\<C-y>" : "\<cr>"

if executable('solargraph')
    " gem install solargraph
    au User lsp_setup call lsp#register_server({
        \ 'name': 'solargraph',
        \ 'cmd': {server_info->[&shell, &shellcmdflag, 'D:\Users\fsnyd\Documents\code\solargraph-shim\latest.bat stdio']},
        \ 'initialization_options': {"diagnostics": "true"},
        \ 'whitelist': ['ruby'],
        \ 'root_uri':{server_info->lsp#utils#path_to_uri(lsp#utils#find_nearest_parent_file_directory(lsp#utils#get_buffer_path(), '.solargraph.yml'))},
        \ })
   set omnifunc=lsp#complete
endif
```

2. Start vim with below's command

```
gvim -u test.vim foo.rb
```

3. type some keys

```
GA<c-x><c-o>
```
Move last line (G), Go next position of `.` (A), And type keys for omnifunc: <kbd>CTRL</kbd>-<kbd>x</kbd> <kbd>CTRL</kbd>-<kbd>o</kbd>

mattn:
On my environment, I got this with the command above.

image

This is not candidates for Array `a`.

mattn:
```
set nocp
call plug#begin('~/vimfiles/plugged')
Plug 'prabirshrestha/async.vim'
Plug 'prabirshrestha/vim-lsp'
"Plug 'prabirshrestha/asyncomplete.vim'
"Plug 'prabirshrestha/asyncomplete-lsp.vim'
call plug#end()

"imap <c-space> <Plug>(asyncomplete_force_refresh)
"inoremap <expr> <Tab> pumvisible() ? "\<C-n>" : "\<Tab>"
"inoremap <expr> <S-Tab> pumvisible() ? "\<C-p>" : "\<S-Tab>"
"inoremap <expr> <cr> pumvisible() ? "\<C-y>" : "\<cr>"

if executable('solargraph')
    " gem install solargraph
    au User lsp_setup call lsp#register_server({
        \ 'name': 'solargraph',
        \ 'cmd': {server_info->[&shell, &shellcmdflag, 'D:\Users\fsnyd\Documents\code\solargraph-shim\latest.bat stdio']},
        \ 'initialization_options': {"diagnostics": "true"},
        \ 'whitelist': ['ruby'],
        \ 'root_uri':{server_info->lsp#utils#path_to_uri(lsp#utils#find_nearest_parent_file_directory(lsp#utils#get_buffer_path(), '.solargraph.yml'))},
        \ })
   set omnifunc=lsp#complete
endif

function! OpenCompletion()
    if &omnifunc != '' && !pumvisible() && (v:char == '.' || v:char == '>')
        call feedkeys("\<C-x>\<C-o>", "n")
    endif
endfunction

autocmd InsertCharPre * call OpenCompletion()
```

If you use this instead of above, you can show popup with dot.

castwide:
Thanks. Those configurations get me further than I was.

One obvious problem is that `foo.rb` starts with a syntax error, the `a.` on the last line. If the file can't be parsed when it first gets opened, its code won't get mapped. That would explain why Solargraph can't detect the `a` variable. It should start working again after you delete the `.` and the mapper starts working. The parser can correct for (most) syntax errors while the file's being edited, but not on the initial parse.

This is what I get:

image

Is it possible that you have some other completion plugin running? Solargraph should never include operators in method suggestions, and the detail on the right looks different.

Other than that, I still have occasional trouble getting the expected completion results, but I suspect that part of the problem is my unfamiliarity with vim. I'll keep looking into it.

castwide:
@mattn In my most recent tests with gVim and gem v0.31.3, I've been getting the completion results I expect. Are you still experiencing problems?

mattn:
Yeah, I get good result now. Thanks.

PR #161, opened 2019-02-24
fix: multi-bytes encoding

SolaWing:
currently all utf-8 comment will be convert to _. I found that Utf-8 string was mistaken as binary. I think it is a good style to let the passed parameters guarantee that their encoding is correct.


castwide:
Thanks. I think we might need to add/modify a couple other specs, but overall this looks solid.

PR #168, opened 2019-03-20
Add .solargraph.yml to gitignore

h-michael:
I want to customize .solargraph.yml with developing solargraph itself.
So I want to add .solargraph.yml to .gitignore.


castwide:
This seems reasonable. In the future I might set up an optional configuration file for analytics, but it makes sense to let users choose their own configurations for their language clients.

PR #185, opened 2019-05-15
Merge closure into master

castwide:
* Deprecated plugin architecture
* Closure pins for enhanced context and scope detection
* Block resolution
* Major Pin refactoring
* Single parameter pin for blocks and methods
* Major NodeProcessor refactoring
* Block rebinding
* Resolve method aliases
* Namespace scope gates (WIP)
* Improved ApiMap::Store indexing
* ApiMap block recipient cache
* Refactored pin and local mapping
* Host synchronization fixes
* Rebind instance_eval, class_eval, and module_eval blocks
* Improved string detection
* Use `@param` tags for parameter pin documentation
* Go To Definition works on `require` paths (castwide/vscode-solagraph#104)
* Mapper processes singleton method directives
* Resolve `self` return types based on current context
* Exclude inner node comments from documentation


minkir014:
Do you mean that after the last 4 commits. We can go to definition for require statements and bundled gems.

castwide:
Yes. It works with both local paths and gems.

minkir014:
Very good. Thanks.

Issue #197, opened 2019-06-21
Global configuration support


taylorthurlow:
Is there any chance we could get support for a global config at `~/.config/solargraph/config.yml` or something similar? Overridden by the project configuration obviously.

castwide:
Good idea. I'll add it to the roadmap.

I might not get to this for a while, so if anyone wants to submit a PR, just let me know and I'll assign it. I don't expect it to be too difficult, but my plate is crazy full.

castwide:
This is released in v0.35.0. The default location for the global config is `~/.config/solargraph/config.yml`, and it can be changed with the `SOLARGRAPH_GLOBAL_CONFIG` environment variable.

nomasprime:
@castwide great, thanks.

Does this merge with existing defaults or do we need to specify defaults? Personally I'd prefer merge but if not, where do we find the existing defaults?

nomasprime:
@castwide great, thanks.

Looks like defaults overwrite built-in defaults. What's the best way of finding out existing defaults, is it [this](https://solargraph.org/guides/configuration)?

castwide:
@nomasprime Yes, when you use `solargraph config` to generate the file, it starts with everything set to the built-in defaults.

PR #210, opened 2019-07-16
Add ability to read from a global config file

drose-shopify:
Allows the use of a global config file defined at ~/.solargraph.yml

https://github.com/castwide/solargraph/issues/197

### DIFF
--- a/lib/solargraph/workspace/config.rb
+++ b/lib/solargraph/workspace/config.rb
@@ -11,9 +11,6 @@ module Solargraph
       # The workspace's .solargraph.yml can override this value.
       MAX_FILES = 5000
         
-      # The directory that contains the global config file
-      GLOBAL_CONFIG_DIR = File.expand_path('~')
-
       # @return [String]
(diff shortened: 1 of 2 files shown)


drose-shopify:
I noticed the linked issue asks for the global config to be in `~/.config/solargraph/config.yml`. Is there a preference to where the global config should live?

castwide:
Thanks, @drose-shopify!

Using `~/.solargraph.yml` is risky because it could make the language server treat the entire home directory as a single workspace, which is almost certainly not the user's intent. I'd prefer either `~/.config/solargraph/config.yml` (common convention) or `~/.solargraph/config.yml` (reuse of existing data directory). Maybe it could also be configurable with an environment variable, like #125 did with the document cache.

drose-shopify:
A fair enough, I didn't think about that. Also <3 the idea of using an environment variable to set the path.

castwide:
Awesome, thanks again! Merging to include in version 0.35.

PR #232, opened 2019-10-04
Speedup Solargraph::ApiMap::Store#fqns_pin

ojab:
Store already resolved pins in a Hash instead of finding them every time.

This change dramatically speedups `solargraph scan` here.


ojab:
Not sure if it's correct one, but hey, specs are passing /shrug.

castwide:
Thanks!

PR #234, opened 2019-10-13
Drop legacy ruby versions

epergo:
Support for Ruby 2.1 ended on 1st of April 2017 and support for Ruby 2.2 ended on 20th June 2018 so in this PR I have removed support for them

* Removed from Travis version matrix so they are not tested
* Incremented the minimum Ruby version in the gemspec
* Removed `travis-bundler.rb` as it had specif logic only for these versions
* Speed up travis build by caching dependencies

Support for Ruby 2.3 ended last 31th March 2019 but I decided to leave it there for now.

WDYT?


castwide:
Dropping legacy support is inevitable, but I don't have a deadline for it yet. My rough estimate is that about 3% of Solargraph users are on Ruby < 2.3. (A year ago it was closer to 5%.) Eventually we'll update the requirement, and legacy users will need to update Ruby or remain on an earlier version of the gem. I expect it to happen by Solargraph 1.0, if not earlier.

Some considerations:

1. The `master` branch currently uses `Redcarpet` for markdown parsing, which requires a native extension. Only MRI and Rubinius are officially supported. If the minimum Ruby is 2.3, we can switch to `kramdown`, which is pure Ruby and therefore works in alternative interpreters like JRuby.
2. As you noticed, legacy support requires some fragile Bundler integration. Requiring Ruby >= 2.3 would eliminate (most of) it.
3. Ruby >= 2.6 has some new features like `RubyVM` that I'd like to explore in upcoming versions of Solargraph. The more we incorporate new features of Ruby, the less tenable legacy support becomes. I'm willing to provide alternative implementations of features for, e.g., Ruby 2.5 vs. Ruby 2.6, but perpetual support back to 2.1 is unrealistic.

In short, the advantages of dropping legacy support outweigh the benefits, but I'm trying to proceed with caution.

I'll keep this PR open so we can hammer out the details. Any comments and feedback are appreciated.

epergo:
Wise words @castwide, if you think that people is still using Ruby versions < 2.3 is worth maintaining them for a bit more

About switching `redcarpet` for `kramdown` while I totally agree I would do it in another PR once this one is merged

Thanks for leaving the PR open, whenever you think is a good time to merge this ping me and I will update it with master and do whatever change is needed

castwide:
Version 0.40.0 will require Ruby >= 2.4.0.

Issue #237, opened 2019-10-22
Rubocop not found


PyvesB:
Hello, hello!

When running the formatter within Eclipse, I'm consistently seeing the following error (both on Windows and macOS):
````
[WARN] Error processing request: [Errno::ENOENT] No such file or directory - rubocop -a -f fi -s XXX
````
The associated stacktrace point to the following line of code:
https://github.com/castwide/solargraph/blob/7e4ba6a43364ddfd5752956ab18f57b34db8d79c/lib/solargraph/language_server/message/text_document/formatting.rb#L16

Basically Rubocop is called by executing a command in a shell.

This is problematic for two different reasons in my use case:
* even though Rubocop is a dependency of Solargraph, my Eclipse plugin installs all executables in a local folder owned by the plugin (see #177). This avoids adding executables to the user's path as a side-effect of installing the plugin and allows things to be nicely cleaned up when the plugin in uninstalled. Consequently, if the user had not previously installed Rubocop, it will not be available globally.
* even if Rubocop was installed by the user independently of Solargraph, it will likely not be available in the IDE's environement. For example, macOS users often manage their path through their _.profile_, _.bash_profile_ or _.bashrc_, and that's where all gems are added to the PATH. When Eclipse is launched as a standard macOS application, it doesn't pick up any of the values in these files and doesn't know about any gems (apart from Solargraph, which my plugin has installed locally and controls).

To sum things up: I don't think Solargraph should assume that the `rubocop` command is available in the environment it's running in.

Could Solargraph integrate with the Rubocop API rather than executing shell commands?

Alternatively, an approach which would cover my second point is to execute the Rubocop command in a login shell ([this post](https://unix.stackexchange.com/a/46856) is quite helpful in explaining some of these concepts) so that it picks up the user's environment properly. On Windows, this would be
`"cmd.exe /c 'rubocop -a -f fi -s #{Shellwords.escape(filename)}'"`. On macOS/Linux, `"/bin/bash -c -l 'rubocop -a -f fi -s #{Shellwords.escape(filename)}'"` does the trick. I'm happy to contribute this workaround if you don't think the Rubocop API is a viable alternative.

What are your thoughts?

Cheers,

Pyves

castwide:
It should be possible to use the RuboCop API for the formatting command. The [RuboCop diagnostics reporter](https://github.com/castwide/solargraph/blob/master/lib/solargraph/diagnostics/rubocop.rb) already works that way.

[...]

PR #239, opened 2019-10-26
RuboCop formatting integration through API

PyvesB:
This pull request aims at fixing #237.

Things don't quite seem to work properly yet: the `result` variable returns a file path rather than the formatted contents. I'm not very familiar with RuboCop, but with a bit of help we can probably turn this into a working solution!

### DIFF
--- a/lib/solargraph/diagnostics/rubocop.rb
+++ b/lib/solargraph/diagnostics/rubocop.rb
@@ -35,17 +35,6 @@ module Solargraph
 
       private
 
-      # @todo This is a smelly way to redirect output, but the RuboCop specs do
-      #   the same thing.
-      # @return [String]
-      def redirect_stdout
-        redir = StringIO.new
-        $stdout = redir
-        yield if block_given?
-        $stdout = STDOUT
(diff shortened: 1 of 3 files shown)


castwide:
I pushed a branch called `PyvesB-rubocop-formatting` that uses a temp file to process formatting: https://github.com/castwide/solargraph/blob/PyvesB-rubocop-formatting/lib/solargraph/language_server/message/text_document/formatting.rb

For some reason, a temp file was the only way I could retrieve the formatting results. The `-s` parameter for sending to stdout stopped working when the `-a` parameter was set.

If it looks okay, you can update your PR whatever way is convenient and I'll merge it.

PyvesB:
Thanks for helping out! That's indeed quite odd. I updated the PR to pretty much the contents of your branch, the only change being the removal of the somewhat superfluous `runner` local variable.

castwide:
Looks good. Thanks again for the help. I'll merge with master and plan to publish this change with version 0.38.0, hopefully in the next couple weeks.

PR #245, opened 2019-11-14
Test if ns is nil to avoid exception

GPif:
the variable ns is sometime nil which cause an exception on solargraph startup.
The & will return nil for ns&.file instead of a nil exception.


castwide:
Thanks. I'll merge this and make a minor change for backwards compatibility (Ruby < 2.3 does not support `&.`).

PR #253, opened 2019-12-26
Ci against ruby27

yahonda:
This pull request runs CI against Ruby 2.7. There are some issues with this pull request. Let me work on later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ rvm:
   - 2.4
   - 2.5
   - 2.6
+  - 2.7
   - jruby-head
 matrix:
   include:
--- a/solargraph.gemspec
+++ b/solargraph.gemspec
@@ -31,6 +31,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'tilt', '~> 2.0'
   s.add_runtime_dependency 'yard', '~> 0.9'
   s.add_runtime_dependency 'e2mmap'
+  s.add_runtime_dependency 'irb', '1.1.0'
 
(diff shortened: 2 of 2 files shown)


yahonda:
RSpec fails with Ruby 2.7.

https://travis-ci.org/castwide/solargraph/jobs/629586594#L915-L927

```ruby
Failures:
916
917  1) Solargraph::TypeChecker ignores undefined return types from external libraries
918     Failure/Error: expect(checker.strict_type_problems).to be_empty
919       expected `[#<Solargraph::TypeChecker::Problem:0x0000000003b2ce30 @location=#<Solargraph::Location test.rb, #<So...tion 5, 27>>>, @message="Unresolved method signature Benchmark.measure", @pin=nil, @suggestion=nil>].empty?` to return true, got false
920     # ./spec/type_checker_spec.rb:187:in `block (2 levels) in <top (required)>'
921
922Finished in 24.61 seconds (files took 1.92 seconds to load)
923676 examples, 1 failure
924
925Failed examples:
926
927rspec ./spec/type_checker_spec.rb:175 # Solargraph::TypeChecker ignores undefined return types from external libraries
```

yahonda:
https://github.com/lsegal/yard/pull/1296 should be a better fix.

castwide:
Thanks a lot for working on this. Do you expect to push any more commits? It looks to me like the only failing spec is an errant TypeChecker warning due to the `benchmark` library being moved to a gem. I can fix that one if you're happy with everything else.

yahonda:
Thanks for the update. As long as you are going to fix the TypeChecker failure with benchmark gem, I have no plan to push any more commits.

Issue #261, opened 2020-01-10
api_map.rb requires optimization


jonnyboyC:
Hi! First off let me say awesome job to getting this project up and running! I personally have a language-server as my own side project and I know how much of an undertaking it can be!

I've know there is some consideration on replacing the parser with a more high performance implementation but do some profiling I've identified that on very large projects the bottleneck isn't parsing speed. I've been running [rbspy](https://rbspy.github.io/) against our companies main rails server project using `solargraph scan`. Solargraph identifies 12k files, so I would definitely say this is on the higher end use cases. 

Long story short rbspy identified [this chunk](https://github.com/castwide/solargraph/blob/master/lib/solargraph/api_map.rb#L105-L118) of `api_map.rb` as contributing something in the ballpark of 90% of the runtime, potentially more. Here is the same chunk of code below:

``` rb
unless bundle.workspace.require_paths.empty?
  reqs.delete_if do |r|
    result = false
    bundle.workspace.require_paths.each do |l|
      pn = Pathname.new(bundle.workspace.directory).join(l, "#{r}.rb")
      if new_map_hash.keys.include?(pn.to_s)
        local_path_hash[r] = pn.to_s
        result = true
        break
      end
    end
    result
  end
end
```

My understanding of what this chunk of code is trying to do is identify `require`s diff. I'm assuming the idea is to only process things that have changed since last time `catalog` was run. as I've tried to come up with a new algorithm to speed this up I believe I've identified a bug in this implementation. Specifically this line
```rb
pn = Pathname.new(bundle.workspace.directory).join(l, "#{r}.rb")
```

running `api_map_spec.rb` I've noticed that `bundle.workspace.directory` is the empty string causing the tests to pass. I've noticed in the general case this won't remove any reqs. Consider this example

```rb
# input
reqs = ['lib1', 'lib2]
new_map_hash = {
  'some/path/lib1.rb': # ...
  'some/other/path/lib3.rb': # ...
}
bundle.workspace.directory = 'some'
bundle.workspace.require_paths = [
  'some/path',
  'some/other/path',
]

# paths tested in the code snippet above
tested_paths = [
  'some/some/path/lib1.rb',
  'some/some/path/lib2.rb',
  'some/some/other/path/lib1.rb',
  'some/some/other/path/lib2.rb',
]

# output
req_final = ['lib1', 'lib2']
expected_req_final = ['lib2']
```

[...]

PR #262, opened 2020-01-10
Add require diff algo using Trie

jonnyboyC:
The PR is in relation to issue #261. Initial profiling looks promising as seen here:

|master|this pr|
|--|--|
|Screen Shot 2020-01-10 at 4 34 49 PM|Screen Shot 2020-01-10 at 4 35 49 PM|

The rbspy reports its flamegraph in svg which github doesn't support unfortunately. Both of these profiles were ~8 minutes of `solargraph scan` on our main repo. The main take away from the current master branch is the wide flat region of the chart corresponding the determine the correct requires is essentially gone in this PR. While this is only one repo, a deep dive into the data shows this snippet went from 31599 profile samples to 174 a ~170X speedup. 

Currently this PR doesn't pass a test in `spec/language_server/message/text_document/definition_spec.rb`. It appears that it doesn't correctly resolve a require. I've been trying to track down the reason but have not been successful so far. The current suite of api_map_spec test pass so there seems to be an edge case that I haven't covered yet. Any help would be appreciated.

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -4,7 +4,6 @@ require 'rubygems'
 require 'set'
 require 'pathname'
 require 'yard'
-require 'trie'
 require 'yard-solargraph'
 
 module Solargraph
@@ -16,6 +15,7 @@ module Solargraph
     autoload :SourceToYard,   'solargraph/api_map/source_to_yard'
     autoload :Store,          'solargraph/api_map/store'
     autoload :BundlerMethods, 'solargraph/api_map/bundler_methods'
+    autoload :ReqDiff,        'solargraph/api_map/req_diff'
 
     include SourceToYard
     include BundlerMethods
@@ -26,6 +26,7 @@ module Solargraph
     # @param pins [Array<Solargraph::Pin::Base>]
     def initialize pins: []
       @source_map_hash = {}
+      @require_diff = ReqDiff.new
       @cache = Cache.new
       @method_alias_stack = []
       index pins
@@ -101,7 +102,7 @@ module Solargraph
       reqs.merge bundle.workspace.config.required
       local_path_hash.clear
       unless bundle.workspace.require_paths.empty?
-        reqs = require_diff(bundle, reqs.clone, new_map_hash)
+        reqs = @require_diff.unresolved_requires(bundle, reqs.clone, new_map_hash)
       end
       reqs.merge implicit.requires
       pins.concat implicit.overrides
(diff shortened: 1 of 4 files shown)


castwide:
One small issue: in the PR, api_map.rb has an errant `end` keyword at line 220.

A slightly larger concern: the fast_trie gem uses a native extension that doesn't work on JRuby. Support for JRuby has been spotty, but I'd like to maintain it if possible.

I'm not sure what's causing the definition_spec failure, but I'll look into it.

jonnyboyC:
I'll take a look to see if I can come up with something that doesn't require `fast_trie` I'm sure there is still a way to speed snippet. I'm still relatively new to ruby, would it be possible to use `fast_trie` and fallback to the initial implementation if we're not on MRI?

castwide:
Maybe. We could definitely handle the runtime issues by rescuing LoadError on `require 'fast_trie'` and using a different method when `Trie` isn't defined.

The tricky part is how to manage the dependency. JRuby will fail to install the Solargraph gem if any of the runtime dependencies fail. Two possible workarounds:

1. Require the user to install `fast_trie` manually.
2. Write a gem extension (e.g., in `extconf.rb`) that tries to install `fast_trie` and fails gracefully if the platform doesn't support it.

Both options have drawbacks. I'm open to other suggestions.

jonnyboyC:
After a poking around,

jonnyboyC:
I poked around for a bit found a few other native trie extensions, and at least one pure ruby trie. I may try to quickly prototype that today to see how much of a performance difference it is. My guess is it'll likely be slower than the current master implementation.

I think either of the two options you specified make sense to me. Would you think move the method it's own class and pick the strategy based on the presence `trie_fast` using the rescue on LoadError?

jonnyboyC:
I've pushed a new commit that takes the approach of your first suggestion going the manual route. I've added a new `ReqDiff` class that checks if `trie-fast` is available. If `trie-fast` is present it uses the new algorithm otherwise we fall back to the hash based approach. I've moved `trie-fast` out of the the `.gemspec` and placed it in the `Gemfile` under the mri platform. I _believe_ that this will install `trie-fast` when using bundler with mri only. I'll try to install jruby and confirm this is the case.

castwide:
Travis CI confirms that the installation works on jruby-head. A bunch of other spec errors happen, but they appear to be unrelated (or only semi-related) to the optional fast-trie dependency. There's more work to be done, but I think we're on the right track.

jonnyboyC:
Good to hear it appears to not blow up on jruby. I'll try get some more time to figure out why some of theses tests are failing soon.

castwide:
I recently pushed some changes to `master` that impact this PR. In short, they reduce the amount of duplicated effort `ApiMap#catalog` performs while resolving requires. This doesn't preclude the potential performance benefits of using a trie, but should at least complement it significantly.

jonnyboyC:
I've pushed a small push to resolve the conflict with master. I'm still trying to find why this test is failing. I can't say with much confidence but the test may currently be passing in master because of this bug I had mentioned earlier. 

In [this](https://github.com/castwide/solargraph/blob/master/lib/solargraph/api_map.rb#L109-L117) `delete_if` block it appears we're checking to see if a require is already in the `bundle.workspace.require_paths`. For the `definition_spec.rb` it is loading in the `/fixtures/workspace` as the test workspace with the following directory structure.

```
├── lib/
│   ├── other.rb
│   └── thing.rb
├── .solargraph.yml
└── app.rb
```

In the first test it looks like we're sending the definition request to the server in `lib/other.rb` for `
Thing.new.do_thing`. Back in `api_map.rb` it appears the current implementations does not find `lib/thing.rb`. In the block I linked above we have the following locals

``` ruby
reqs = <Set: {"thing"}>
workspace_path = <Pathname:spec/fixtures/workspace>
base = "spec/fixtures/workspace/lib"
r = "thing"
pn = "spec/fixtures/workspace/spec/fixtures/workspace/lib/thing.rb"

file_keys = [
 "spec/fixtures/workspace/lib/other.rb",
 "spec/fixtures/workspace/lib/thing.rb",
 "spec/fixtures/workspace/app.rb"
]
```

In this iteration I think we would have intended to remove "thing" from the `reqs` set and updated `local_path_hash` 
```ruby
local_path_hash = {"thing"=>"spec/fixtures/workspace/lib/thing.rb"}` 
```

I believe the test passes because it depends on this block not removing "thing" from reqs. Would there any additional setup calls in the `definition_spec.rb` that may need to occur?

lib/solargraph/api_map.rb
```diff
@@ -103,19 +105,7 @@ def catalog bundle
...
-            pn = workspace_path.join(base, "#{r}.rb").to_s
-            if file_keys.include? pn
-              local_path_hash[r] = pn
-              true
-            else
-              false
-            end
-          end
-        end
+        local_path_hash = @require_diff.unresolved_requires(bundle, reqs.clone, new_map_hash)
```
castwide:
```suggestion
        local_path_hash.merge! @require_diff.unresolved_requires(bundle, reqs.clone, new_map_hash)
```

castwide:
One issue I see is [this local_path_hash assignment](https://github.com/castwide/solargraph/pull/262/files#r368994425).

```ruby
# This creates a local variable instead of replacing `local_path_hash` because
# ApiMap doesn't have a `local_path_hash=` attribute writer.
local_path_hash = @require_diff.unresolved_requires(bundle, reqs.clone, new_map_hash)

# This merges the hash from `@require_diff.unresolved_requires` into `local_path_hash`.
local_path_hash.merge! @require_diff.unresolved_requires(bundle, reqs.clone, new_map_hash)
```

lib/solargraph/api_map.rb
jonnyboyC: Good catch this should be fixed now as well as a few other potentially problems.

jonnyboyC:
Finally had some time to take a look into the into why these test were failing (actually using your vscode debugging extension). I did notice that jruby failed but saw it was in the allowed to fail bucket. I didn't see anything in there related to this Trie change so I'm assuming it was already failing for other reasons.

mattn:
Any update on this?

jonnyboyC:
I haven't done anything on this branch since my last update. There was a failing test but it didn't appear to me to have anything to do with my changes so I'd assume it was fine. Likely I need to rebase but to my knowledge this could get merged if when @castwide wants.

castwide:
Sorry for the delay in getting back to this. I've tested this locally and it seems to work fine, but I don't see any significant change in performance. Can you give me a reproducible example that demonstrates the improvement?

jonnyboyC:
No worries, let me test it again locally and get some timing. Unfortunately the example I was testing on locally was our private repo which has ~10k files. I may try running this against the rails repo to see if this produces a noticeable speedup

Issue #272, opened 2020-02-04
Cannot scan if bundler returns warnings


earksiinni:
**Edit**: I've issued a PR: https://github.com/castwide/solargraph/pull/273

When I run `solargraph scan`, I get an error:

```
Traceback (most recent call last):
	21: from /nix/store/dmphzmp34cs9a238l6yrbxyz7cb1dv06-gamacy/bin/solargraph:18:in `<main>'
	20: from /nix/store/dmphzmp34cs9a238l6yrbxyz7cb1dv06-gamacy/bin/solargraph:18:in `load'
	19: from /nix/store/dmphzmp34cs9a238l6yrbxyz7cb1dv06-gamacy/lib/ruby/gems/2.6.0/gems/solargraph-0.37.2/bin/solargraph:5:in `<top (required)>'
	18: from /nix/store/nfs5kfm0k0bmhixnj61wrzz2vhz7y22j-ruby2.6.5-thor-0.20.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
	17: from /nix/store/nfs5kfm0k0bmhixnj61wrzz2vhz7y22j-ruby2.6.5-thor-0.20.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
	16: from /nix/store/nfs5kfm0k0bmhixnj61wrzz2vhz7y22j-ruby2.6.5-thor-0.20.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
	15: from /nix/store/nfs5kfm0k0bmhixnj61wrzz2vhz7y22j-ruby2.6.5-thor-0.20.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
	14: from /nix/store/v3a7vpwik3gplb58flyqlzakn63k1vrb-ruby2.6.5-solargraph-0.37.2/lib/ruby/gems/2.6.0/gems/solargraph-0.37.2/lib/solargraph/shell.rb:164:in `scan'
	13: from /nix/store/h3kc2i0z93k5pnb6gsl6678yvgaci708-ruby-2.6.5/lib/ruby/2.6.0/benchmark.rb:293:in `measure'
[...]

PR #273, opened 2020-02-04
Ignore Bundler warnings when parsing JSON

earksiinni:
This PR attempts to address https://github.com/castwide/solargraph/issues/272 by splitting the output from Bundler when executing `Documentor.specs_from_bundle` and ignoring all lines but the last one. The last one is the only one that we care about since that contains the Bundler specs in JSON, and everything else can be assumed to be a Bundler warning.


castwide:
Thanks!

PR #289, opened 2020-02-27
Refactored uri queue synchronization.

Prajjwal:
`Solargraph::LanguageServer::Host::Sources` currently utilizes a queue
of uris that is polled every `0.25s` to see if there's anything there.

This commit refactors that to use a `ConditionVariable` that
automatically blocks until a uri is available. Two new methods -
`add_uri` and `next_uri` wrap this logic to avoid manual waiting and
signaling.


castwide:
Looks good to me. Thanks!

PR #299, opened 2020-04-03
Avoid lazy initialization of Mutex

olleolleolle:
Move initialization of the `Mutex` to the constructor.

  - this makes the code more robust on non-MRI platforms, where no GIL/GVL protects the individual line
  - _perhaps_ this is easier to follow for someone trying to understand the flow

(One could argue it'd be cheaper to refer to the instance variable `@mutex` directly, instead of calling a method. But this change is about avoiding the lazy init.)

### DIFF
--- a/lib/solargraph/language_server/host/sources.rb
+++ b/lib/solargraph/language_server/host/sources.rb
@@ -12,6 +12,7 @@ module Solargraph
         include UriHelpers
 
         def initialize
+          @mutex = Mutex.new
           @stopped = true
         end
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
cc @ioquatix

ioquatix:
Nice work @olleolleolle !

castwide:
Thanks! It occurred to me that the same change would benefit Backport, but you [beat me to the punch](https://github.com/castwide/backport/pull/3) 😄

PR #302, opened 2020-04-08
feat: treat prepend module as include module

SolaWing:
I only add basic prepend support like include. I am not sure if the prepend order matters.


castwide:
Thanks. We'll need to make a couple changes to `ApiMap` so it builds the ancestor tree correctly, but this is a good start.

PR #326, opened 2020-05-25
Allow reverse_markdown 2

aried3r:
The only breaking change was dropping Ruby 1.9 supports, which solargraph
doesn't support anyway.
https://github.com/xijo/reverse_markdown/blob/master/CHANGELOG.md#200---march-2020


castwide:
Thanks!

PR #327, opened 2020-05-25
Speed up Travis CI run

aried3r:
Travis runs `bundler install` itself, but in parallel.
`cache: bundler` should speed up future CI runs.

Screenshot 2020-05-25 at 13 39 03

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 rvm:
   - 2.1
   - 2.2
@@ -23,6 +24,5 @@ matrix:
       
 before_install:
 - ruby ./travis-bundler.rb
-- bundle install
 before_script: yard gems
 script: rspec
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,5 @@ matrix:
       
 before_install:
 - ruby ./travis-bundler.rb
-before_script: yard gems
-script: rspec
+before_script: bundle exec yard gems
+script: bundle exec rspec
--- a/lib/solargraph/documentor.rb
+++ b/lib/solargraph/documentor.rb
@@ -76,6 +76,7 @@ module Solargraph
     def self.specs_from_gemfile directory, name
       Solargraph.with_clean_env do
         Dir.chdir directory do
+          Solargraph.logger.warn Dir.entries(Dir.pwd)
           cmd = [
             'bundle', 'exec', 'ruby', '-e',
             "require 'bundler'; require 'json'; puts Bundler::Definition.build('#{name}', '#{name}.lock', false).specs_for([:default]).map { |spec| [spec.name, spec.version] }.to_h.to_json"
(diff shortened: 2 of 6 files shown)


aried3r:
Since the tests are now failing, I assume I overlooked the specific test setup needed for solargraph. 

Am I missing something here, @castwide or is there a chance this can be "fixed" (assuming something is broken to begin with)?

castwide:
This was an issue that arose with Travis CI builds a while ago. Someone from Travis suggested removing `cache: bundler`, and that fixed it. I'm not sure what the root cause is, but part of the problem might be the hacks involved in supporting Ruby < 2.3.

aried3r:
> Someone from Travis suggested removing cache: bundler, and that fixed it.

Hmm, I guess because caching moves the gems to `--path=${BUNDLE_PATH:-vendor/bundle}`. I'll try to have another look and will remove `cache: bundler` if necessary. The removal of the duplicate `bundle install` should still give a nice speedup.

> but part of the problem might be the hacks involved in supporting Ruby < 2.3.

I've seen #234, what is the current status of that PR? Also, which hacks are you referring to? There are [statistics](https://ecosystem.rubytogether.org/) now, sourced from RubyGems. 2.1 still seems active, but if supporting EOL Rubies means holding back solargraph it's something to consider.

aried3r:
I think I found the changes you mentioned. https://github.com/castwide/solargraph/commit/d463297db2b1f6fc3e582febcdfeaa02811760d9

I'll try to play around a bit. Sadly I cannot reproduce this locally, so I'll create some noise here.

castwide:
> I've seen #234, what is the current status of that PR? Also, which hacks are you referring to?

Primarily the `.travis-bundler.rb` script that selects a version of Bundler based on the Ruby version. There's some other Bundler compatibility code in the library itself, but it's less likely to have side effects.

> There are [statistics](https://ecosystem.rubytogether.org/) now, sourced from RubyGems. 2.1 still seems active, but if supporting EOL Rubies means holding back solargraph it's something to consider.

Dropping support for Ruby < 2.3 is probably inevitable. I'm not sure when it'll happen, but it might be as soon as one or two minor releases, i.e., 0.40 or 0.41.

> I'll try to play around a bit. Sadly I cannot reproduce this locally, so I'll create some noise here.

No problem. Thanks for the heads up.

aried3r:
It seems following my changes I cannot even get the specs to work on any Ruby version. I tried being more explicit using `Bundler::Definition.build` (following [this article](https://nts.strzibny.name/working-with-bundler-from-your-ruby-code/)) but that also didn't work on Travis, while it works locally.

I didn't spend too much time on it, though and the code I wrote was pretty hacky, lacking error handling etc, but I just wanted to play around.

I'm not going to pursue this branch anymore and instead will open another PR which just removes the duplicate `bundler install` (which hopefully isn't the reason it works at all).

gwillcox-r7:
Given Travis has now been removed with https://github.com/castwide/solargraph/pull/627 would it make sense to perhaps close this PR?

PR #342, opened 2020-07-16
Use `ensure` block for File.unlink tempfile

Nitrodist:
This ensures that temporary files are removed even if the begin block
fails. This has been happening on some dev machines.

https://www.reddit.com/r/neovim/comments/gmq965/vim_creating_tmp_files/


castwide:
Thanks!

PR #425, opened 2021-04-02
fix(formatter): strip file:// from file path given to RuboCop

jimeh:
Passing in a file path to RuboCop which begins with `file://` seems to cause
strange behavior. Some parts of RuboCop seem to work correctly, while others
don't.

I have come across a couple of cases where formatting via Solargraph while it is
passing in `file:///path/to/file` causes some disabled cops corrections to run,
while other disabled cops do not run.

I have not narrowed it down to a exact list of cops, or affected versions of
Rubucop, but this change fixes all such instances I've run into.

### DIFF
--- a/lib/solargraph/language_server/message/text_document/formatting.rb
+++ b/lib/solargraph/language_server/message/text_document/formatting.rb
@@ -41,6 +41,8 @@ module Solargraph
           end
 
           def cli_args file, config
+            file = file.gsub(/^file:\/\//, '')
+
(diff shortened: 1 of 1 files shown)


jimeh:
I split out the logging stuff to #426, as I felt it was cleaner having these two things in separate PRs.

jimeh:
#435 fixes this same issue in a nicer way by using `UriHelpers.uri_to_file`.

PR #426, opened 2021-04-06
feat(formatter): log list of rubocop corrections at INFO level

jimeh:
INFO seems like a sensible log level, as it's already decently noisy, and not
quite as detailed as what you'd expect at the DEBUG level.


castwide:
Thanks!

PR #430, opened 2021-04-13
Allow configuring the version of rubocop to require

RaMin0:
Add the ability to specify a `rubocop` version when invoking [diagnostics](https://github.com/castwide/solargraph/blob/18854d7577737cf9929063d96dd99b3d0a77f703/lib/solargraph/diagnostics/rubocop.rb#L3) and [formatting](https://github.com/castwide/solargraph/blob/18854d7577737cf9929063d96dd99b3d0a77f703/lib/solargraph/language_server/message/text_document/formatting.rb#L3). This resolves the issue in case, for example, two versions of `rubocop` are installed and the active project requires the older version to be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,21 @@ Run `bundle install` and use `bundle exec yard gems` to generate the documentati
 
 In order to make sure you're using the correct dependencies, you can start the language server with Bundler. In VS Code, there's a `solargraph.useBundler` option. Other clients will vary, but the command you probably want to run is `bundle exec solargraph socket` or `bundle exec solargraph stdio`.
 
+### Rubocop Version
+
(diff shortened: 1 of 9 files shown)


jimeh:
I've typically solved this by adding `solargraph` to the `Gemfile`, and then have my editor launch Solargraph via `bundle exec`.

But I think this could be useful for people who don't want to add Solargraph to Gemfiles.

RaMin0:
@jimeh Exactly! I find it okay to add development dependencies to the `Gemfile` like `rubocop`, but `solargraph` is a preference dependency that I don't want to force on every one working on the project, hence this PR. If you want to use a custom configuration of `solargraph`, throw in a `.solargraph.yml` and no one will ever know (of course as long as you don't commit it to source control 😅)

castwide:
Thanks!

PR #434, opened 2021-04-29
Fix source of diagnostic

aki77:
The diagnostic source is easy to understand if it is a linter name.

Image from Gyazo

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic

https://github.com/microsoft/vscode-eslint/blob/3e2b4a8a72487da052c46dcf4e145767bb81d110/server/src/eslintServer.ts#L424-L444


castwide:
Thanks!

PR #454, opened 2021-07-07
fix: yard stdout may break language client

SolaWing:
I found recent version server broken some times, and find the unexpected stdio yard out.
seems original position not require early enough. 

图片


castwide:
Good catch! This is likely related to https://github.com/castwide/vscode-solargraph/issues/220.

PR #457, opened 2021-07-10
Improve the performance of catalog by 4 times 

SolaWing:
In My workspace, I have ~100,000 pins, and I often wait for a long time without semantic completion。

this PR optimize this problem by:

1. cache pin.namespace, catalog will call it many times. this boost store index performance to **almost twice**.
2. change store return type from Array to set, to avoid unnecessary array conversion. this gain a little performance.
3. remove duplicate attach call in method library_for. since specific library for already call attach. 
4. sink catalog_inlock, ensure attach also call it and mark @synchronized = true, to avoid unnecssary catalog

In All this actions, In My workspace, I reduce single store.index from `1~1.2s` to `0.5~0.7s`, and catalog call's from 16 to 6(Same as textChanged event count). In total boost 4 tims performance.

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -138,7 +138,7 @@ module Solargraph
 
     # An array of pins based on Ruby keywords (`if`, `end`, etc.).
     #
-    # @return [Array<Solargraph::Pin::Keyword>]
+    # @return [Enumerable<Solargraph::Pin::Keyword>]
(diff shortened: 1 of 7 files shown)


castwide:
Awesome work! I expect to put this through more testing before release, but it hasn't given me any problems in my local environment yet. It's a good candidate for inclusion in 0.43.0. Thanks!

PR #458, opened 2021-07-11
feat: response REQUEST_CANCELLED, according to specification

SolaWing:
see https://microsoft.github.io/language-server-protocol/specifications/specification-current/#cancelRequest. response should always be paired


castwide:
Thanks!

PR #459, opened 2021-07-11
performance: handle request on a specific thread, and cancel completion when there has newer completion request

SolaWing:
The PR first commit processes the message on a dedicated thread and releases the input thread. So it can check how many requests are piled up and cancel unnecessary requests.

and second commit cancels completions when there have newer completion request. to avoid unnecessary catalog.

this will make server's more responsive and have better expreience

### DIFF
--- a/lib/solargraph/language_server/host/message_worker.rb
+++ b/lib/solargraph/language_server/host/message_worker.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+module Solargraph
+  module LanguageServer
+    class Host
+      # A serial worker Thread to handle message.
+      #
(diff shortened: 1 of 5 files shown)


lib/solargraph/language_server/host/message_worker.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Solargraph
+  module LanguageServer
+    class Host
+      # A serail worker Thread to handle message.
```
ccarse:
```suggestion
      # A serial worker Thread to handle message.
```

castwide:
Slick solution! Thanks for all your work on this.

PR #494, opened 2021-10-20
feat: define_method rebind self

### LINKED ISSUE #493: bug: block in define_method should infer to instance type
current self is not change

SolaWing:
I made a simple patch, and it seems work. it fixes #493

### DIFF
--- a/lib/solargraph/pin/block.rb
+++ b/lib/solargraph/pin/block.rb
@@ -44,15 +44,24 @@ module Solargraph
         return nil unless api_map.rebindable_method_names.include?(word)
         chain = Parser.chain(receiver, location.filename)
         locals = api_map.source_map(location.filename).locals_at(location)
-        if ['instance_eval', 'instance_exec', 'class_eval', 'class_exec', 'module_eval', 'module_exec'].include?(chain.links.last.word)
+        links_last_word = chain.links.last.word
+        if %w[instance_eval instance_exec class_eval class_exec module_eval module_exec].include?(links_last_word)
           return chain.base.infer(api_map, self, locals)
-        else
-          receiver_pin = chain.define(api_map, self, locals).first
-          if receiver_pin && receiver_pin.docstring
-            ys = receiver_pin.docstring.tag(:yieldself)
-            if ys && ys.types && !ys.types.empty?
-              return ComplexType.try_parse(*ys.types).qualify(api_map, receiver_pin.context.namespace)
-            end
+        end
+        if 'define_method' == links_last_word and chain.define(api_map, self, locals).first&.path == 'Module#define_method' # change class type to instance type
+          if chain.links.size > 1 # Class.define_method
+            ty = chain.base.infer(api_map, self, locals)
+            return Solargraph::ComplexType.parse(ty.namespace)
+          else # define_method without self
+            return Solargraph::ComplexType.parse(closure.binder.namespace)
+          end
+        end
+        # other case without early return, read block yieldself tags
+        receiver_pin = chain.define(api_map, self, locals).first
+        if receiver_pin && receiver_pin.docstring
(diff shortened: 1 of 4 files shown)


lib/solargraph/pin/block.rb
```diff
@@ -44,8 +44,20 @@ def binder_or_nil api_map
...
-        if ['instance_eval', 'instance_exec', 'class_eval', 'class_exec', 'module_eval', 'module_exec'].include?(chain.links.last.word)
+        links_last_word = chain.links.last.word
+        if ['instance_eval', 'instance_exec', 'class_eval', 'class_exec', 'module_eval', 'module_exec'].include?(links_last_word)
           return chain.base.infer(api_map, self, locals)
+        elsif 'define_method' == links_last_word # change class type to instance type
+          if chain.links.size > 1 # Class.define_method
+            ty = chain.base.infer(api_map, self, locals)
+            return Solargraph::ComplexType.parse(ty.namespace)
+          else # define_method without self
+            ty = context
```
emmohac: Wondering that's your thought on `return scope != :instance ? Solargraph::ComplexType.parse(context.namespace) : context`

SolaWing: I copyed from context method. I add more test cases and refactor this code

lib/solargraph/pin/block.rb
```diff
@@ -44,15 +44,24 @@ def binder_or_nil api_map
...
+            return Solargraph::ComplexType.parse(ty.namespace)
+          else # define_method without self
+            return Solargraph::ComplexType.parse(closure.binder.namespace)
+          end
+        end
+        # other case without early return, read block yieldself tags
+        receiver_pin = chain.define(api_map, self, locals).first
+        if receiver_pin && receiver_pin.docstring
+          ys = receiver_pin.docstring.tag(:yieldself)
+          if ys && ys.types && !ys.types.empty?
```
emmohac, line 63: Maybe `ys&.types && !ys.types.empty?` should work too.

emmohac: Same for the line above in the `if` condition: `receiver_bin&.docstring`

lib/solargraph/pin/block.rb
```diff
@@ -44,15 +44,24 @@ def binder_or_nil api_map
...
-        else
-          receiver_pin = chain.define(api_map, self, locals).first
-          if receiver_pin && receiver_pin.docstring
-            ys = receiver_pin.docstring.tag(:yieldself)
-            if ys && ys.types && !ys.types.empty?
-              return ComplexType.try_parse(*ys.types).qualify(api_map, receiver_pin.context.namespace)
-            end
+        end
+        if 'define_method' == links_last_word and chain.define(api_map, self, locals).first&.path == 'Module#define_method' # change class type to instance type
+          if chain.links.size > 1 # Class.define_method
```
emmohac, line 52:
```
return Solargraph::ComplexType.parse(closure.binder.namespace) if chain.links.size <= 1
ty = chain.base.infer(api_map, self, locals)
return Solargraph::ComplexType.parse(ty.namespace)
```
Logic should be the same, but just a bit shorter.

lib/solargraph/pin/block.rb
emmohac: Cool stuff 💪

lib/solargraph/pin/block.rb
SolaWing: this is original code and i just reinvent it

lib/solargraph/pin/block.rb
SolaWing: I don’t like if modifier in this condition as `if else` make condition more obviously.

castwide:
Thanks!

PR #499, opened 2021-11-10
Honor Rubocop "Exclude" config for diagnostics

grncdr:
RuboCop can be configured to ignore certain files with Exclude directives in `.rubocop.yml`

However, passing filename on the command line (which is what Solargraph emulates) overrides this config. Adding the --force-exclusion option to the command tells RuboCop to apply the config anyways.


castwide:
I tested this with a couple different configurations, and it does seem like the least surprising behavior. Thanks.

PR #508, opened 2021-11-27
Remove condition for skipping gems

alisnic:
@gem_paths is updated below, so this condition may skip gems that are actually searchable.

Currently, solargraph does not include `actionpack` gem yardocs in my test projects because of this condition. Here is the impact of the change in one of my projects:

```bash
# before this change
[3] pry(#<Repl>)> api_map.yard_map.yardocs.size
=> 114

# after this change
[3] pry(#<Repl>)> api_map.yard_map.yardocs.size
=> 126
```

### DIFF
--- a/lib/solargraph/yard_map.rb
+++ b/lib/solargraph/yard_map.rb
@@ -214,7 +214,6 @@ module Solargraph
           if @source_gems.include?(spec.name)
             next
           end
-          next if @gem_paths.key?(spec.name)
           yd = yardoc_file_for_spec(spec)
           # YARD detects gems for certain libraries that do not have a yardoc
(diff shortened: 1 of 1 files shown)


alisnic:
so after re-reviewing the code, it seems to me that this is a fix that should not work, but it does for some reason and I don't understand why.

it's strange that:
- gems are processed several times
- yardocs for `actionpack` are not found first time, but found on the second

@castwide I can provide a repro repository for the issue that I'm trying to fix

alisnic:
this was because I was running solargraph outside bundle, first time it resolved actionpack 6.1, which was not part of the bundle, then it resolved correctly to 5.2.6. I will try to make solargraph run better outside bundle, but that's a topic for another MR

PR #516, opened 2021-12-02
Always use reference YARD tags when resolving param types

### LINKED ISSUE #515: Incomplete handling for @param reference tags
### TLDR
@param reference tags are ignored if the there are regular @param tags defined beside them.

### Problem statement
While  there is support for standalone param reference tags, those don't work correctly if there is a mix of both explicit and reference tags define in the YARD docs of the method. See below a failing test case. This is the test at https://github.com/castwide/solargraph/blob/master/spec/pin/parameter_spec.rb#L150 which I modified by adding the `qux` parameter before the reference.
```ruby
  it 'infers return types from method reference tags' do
    source = Solargraph::Source.load_string(%(
      class Foo
        # @param bla [String]
        def bar(bla)
        end
        # @param qux [Integer]
        # @param (see Foo#bar)
        def baz(qux, bla)
          bla._
        end
      end
    ), 'test.rb')
    api_map.map source
    clip = api_map.clip_at('test.rb', [8, 14])
    paths = clip.complete.pins.map(&:path)
    expect(paths).to include('String#upcase')
  end
```

TeeSeal:
Closes #515

### DIFF
--- a/lib/solargraph/pin/parameter.rb
+++ b/lib/solargraph/pin/parameter.rb
@@ -145,10 +145,7 @@ module Solargraph
         # meths.shift # Ignore the first one
         meths.each do |meth|
           found = nil
-          params = meth.docstring.tags(:param)
-          if params.empty?
-            params = see_reference(docstring, api_map)
-          end
+          params = meth.docstring.tags(:param) + see_reference(docstring, api_map)
           params.each do |p|
             next unless p.name == name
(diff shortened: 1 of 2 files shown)


castwide:
This is excellent, thanks! I'll plan to include it in the next release, but might add a couple more tests beforehand, just to be careful.

PR #529, opened 2022-01-08
Support URIs prefixed with single slashed file scheme

### LINKED ISSUE #521: Workaround faulty URIs
After upgrading eclipse from 2021-09 to 2021-12 the solargraph-plugin is currently not workling.
@PyvesB analyzed the problem and found out that lsp4e sends faulty URIs to solargraph.

This is already reported to the lsp4e-project: https://github.com/eclipse/lsp4e/issues/29

But because java seems to be tolerant about the format
```
    URI _uri = new URI("file:///Z:/dev_tools/");
    System.out.println(_uri + ": " + new File(_uri).exists());
    _uri = new URI("file:/Z:/dev_tools/");
    System.out.println(_uri + ": " + new File(_uri).exists());
```
outputs
```
file:///Z:/dev_tools/: true
file:/Z:/dev_tools/: true
```
You maybe can also make uri_helpers.rb more tolerant by changing
https://github.com/castwide/solargraph/blob/949fa3c3be83de437f84c6acf8b9503347083471/lib/solargraph/language_server/uri_helpers.rb#L17
from
```
      def uri_to_file uri
        decode(uri).sub(/^file\:\/\//, '').sub(/^\/([a-z]\:)/i, '\1')
      end
```
to
```
      def uri_to_file uri
        decode(uri).sub(/^file\:\/\//, '').sub(/^file\:/, '').sub(/^\/([a-z]\:)/i, '\1')
      end
```
This ensures only that the leading `file:` is always removed.
At least in my case solargraph+eclipse+lsp4e is now working again with this change.

I've created this additional issue because I'm not sure when lsp4e fixes the issue and added it to a new release.

It's ok, if you don't want to change uri_to_file - in this case at least the workaround is documented here. :-)

PyvesB:
Fixes #521.

### DIFF
--- a/lib/solargraph/language_server/uri_helpers.rb
+++ b/lib/solargraph/language_server/uri_helpers.rb
@@ -14,7 +14,7 @@ module Solargraph
       # @param uri [String]
       # @return [String]
       def uri_to_file uri
-        decode(uri).sub(/^file\:\/\/?/, '').sub(/^\/([a-z]\:)/i, '\1')
+        decode(uri).sub(/^file\:(?:\/\/)?/, '').sub(/^\/([a-z]\:)/i, '\1')
       end
 
       # Convert a file path to a URI.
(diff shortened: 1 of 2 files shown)


castwide:
Thanks!

PR #536, opened 2022-02-15
Doc: update readme.md for add solargraph support

jayli:
add solargraph support for vim-easycomplete.

More Info: <https://github.com/jayli/vim-easycomplete>


castwide:
Thanks!

PR #556, opened 2022-05-13
details on config behavior

mattmassicotte:
I found the global configuration to be incredibly helpful, but didn't find any docs about it anywhere. Thought it might be nice to include.


castwide:
Thanks!

PR #559, opened 2022-05-20
TypeChecker allow Hash as argument

dezza:
Typechecker complains if second argument is a Hash.
```rb
# Too many arguments to Hash#store
{}.store("key", {foo: "bar"})
```

Not sure if this is correct in context, but I hope it highlights the issue.


castwide:
Confirmed. The bug is partly in the `arity_problems_for` method you identified, but the root cause is twofold:

1. Certain versions of the Ruby core documentation incorrectly define `Hash#store` as not accepting any arguments. This can be fixed with an `@overload` tag, but...
2. The `arity_problems_for` method did not take overloads into account when checking for correct arity.

I pushed a [commit to master](https://github.com/castwide/solargraph/commit/dbc70691991fd5ed48f65baa4dff82c70ea81bb7) that should resolve both issues. If the problem persists, please let me know.

PR #560, opened 2022-05-21
Fix query_symbols changed to exact match

dezza:
I could not find anything about the lsp spec suggesting to use fuzzy matching for this function. Opening this PR to discuss original intent.

Currently it breaks jumping to function in vim with `:tag` which is mapped to `workspace/symbol` in solargraph-lsp since all substring matches are also considered before the exact match. This is how it works in `gopls` as well.

Original issue: #44

### DIFF
--- a/lib/solargraph/source_map.rb
+++ b/lib/solargraph/source_map.rb
@@ -77,7 +77,8 @@ module Solargraph
     # @return [Array<Pin::Base>]
     def query_symbols query
       return document_symbols if query && query.empty?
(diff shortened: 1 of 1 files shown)


castwide:
I suspect that vim's `:tag` feature selects the first symbol from the array in the `workspace/symbol` response. Solargraph doesn't sort the results, so the first symbol is not necessarily the closest match. A better solution might be to sort the results; this will allow `:tag` to find the correct result while still allowing for fuzzy matching.

dezza:
Thanks a lot for the suggestion and commit! Excuse my suggestions while I familiarize myself with the project. I thought it was easier to visualize with a pull request.

Issue #566, opened 2022-06-04
@!visibility YARD metadata tag support


nfedyashev:
Just wanted to clarify if @!visibility meta tags are not supported(yet) or is it something with my local setup?

The following class always autosuggests both methods in autocompletion and "Visibility: public" is displayed in popup for both methods.

```
class Foo
  # @!visibility private
  def bar
  end

  # @!visibility protected
  def baz
  end
end
```

I know `private` method modifier works great in Solagraph and I probably just use that.

The original idea was just to change visibility to private(and exclude from autocompletion) a method that is rarely used in the project but still need to be public.

Solargraph version: `master` branch

castwide:
Solargraph doesn't support the `@!visibility` directive yet, but it should be easy to implement. I'll see if I can get it into the next release.

nfedyashev:
@castwide thank you very much!

castwide:
Released in v0.47.0.

PR #567, opened 2022-06-05
Implement yard directive visibility handling

Liberatys:
Implement handling of the yard @!visibility directive.

Should address #566.

Signed-off-by: liberatys <nick.anthony.flueckiger@hey.com>

### DIFF
--- a/lib/solargraph/source_map/mapper.rb
+++ b/lib/solargraph/source_map/mapper.rb
@@ -12,7 +12,7 @@ module Solargraph
 
       private_class_method :new
 
-      MACRO_REGEXP = /(@\!method|@\!attribute|@\!domain|@\!macro|@\!parse|@\!override)/.freeze
+      MACRO_REGEXP = /(@\!method|@\!attribute|@\!visibility|@\!domain|@\!macro|@\!parse|@\!override)/.freeze
 
       # Generate the data.
       #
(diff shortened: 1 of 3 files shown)


lib/solargraph/source_map/mapper.rb
```diff
@@ -147,6 +147,23 @@ def process_directive source_position, comment_position, directive
               pins.last.docstring.add_tag YARD::Tags::Tag.new(:param, '', pins.last.return_type.to_s.split(', '), 'value')
             end
           end
+        when 'visibility'
+          begin
+            kind = directive.tag.text&.to_sym
+            return unless kind == :private || kind == :protected
```
ghost, line 153: The possible values are 'public' | 'protected' | 'private'.

In the case of **public** I don't see a case where the visibility would have to be adjusted. (Or is a yard comment more important than other means of setting visibility?). To me public should be the default.
In the case of **private** it should overwrite the current pins.
In the case of **protected** one could argue that if the visibility is private it should not overwrite the current pins?

lib/solargraph/source_map/mapper.rb
```diff
@@ -147,6 +147,23 @@ def process_directive source_position, comment_position, directive
...
+
+            name = directive.tag.name
+            namespace = closure_at(source_position) || @pins.first
+            if namespace.location.range.start.line < comment_position.line
+              namespace = closure_at(comment_position)
+            end
+
+            matches = pins.select{ |pin| pin.is_a?(Pin::Method) && pin.name == name && pin.namespace == namespace && pin.context.scope == namespace.is_a?(Pin::Singleton) ? :class : :instance }
+            matches.each do |pin|
+              # @todo Smelly instance variable access
```
ghost, line 163: is there a reason for not having setters for the different variables?

castwide:
This is a great start, but there are still a couple of use cases that don't match [YARD's specification](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#visibility).

1. A `@!visibility` directive that is attached to a method should not modify other methods. Example spec:
    ```ruby
    it "processes private visibility directives attached to methods" do
      map = Solargraph::SourceMap.load_string(%(
        class Example
          # @!visibility private
          def method1; end

          def method2; end
        end
      ))
      method1 = map.first_pin('Example#method1')
      expect(method1.visibility).to be(:private)
      method2 = map.first_pin('Example#method2')
      expect(method2.visibility).to be(:public)
    end
    ```

2. A `@!visibility` directive that is not attached to a method should modify other methods and can be overridden. Example spec:
    ```ruby
    it "processes class-wide private visibility directives" do
      map = Solargraph::SourceMap.load_string(%(
        class Example
          # @!visibility private

          def method1; end

          def method2; end

          # @!visibility public
          def method3; end
        end
      ))
      method1 = map.first_pin('Example#method1')
      expect(method1.visibility).to be(:private)
      method2 = map.first_pin('Example#method2')
      expect(method2.visibility).to be(:private)
      method3 = map.first_pin('Example#method3')
      expect(method3.visibility).to be(:public)
    end
    ```

lib/solargraph/source_map/mapper.rb
castwide: This is a wart left over from the rules that control how pins can be updated in place. Pin attributes are typically not writable because there are finite cases where modifying a pin won't adversely affect the ApiMap. Those instance variable accesses will eventually get refactored.

castwide:
I'll go ahead and merge this and make the changes to pass the two additional specs. It'll be released in 0.47.0.

PR #570, opened 2022-07-01
Differentiate between missing yard docs and gem not being found

dgmora:
I was debugging an issue where I was getting a `Required path <gem> could not be resolved`. Turns out I was missing the yard documentation. I have seen similar issues and I imagine it's quite common to receive this error.

While both are errors, missing a gem or not having the documentation are two different issues. I think this would be helpful to narrow down the possible issues on why is a gem not being found.

I don't know the codebase, so I preferred to open this early rather than spending a lot of time with it. Let me know if this would be acceptable or possible things to watch out and I can implement it, add specs, add the message you prefer etc.

Here's how it looks like for me:

Screenshot 2022-07-01 at 08 20 12

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -21,6 +21,9 @@ module Solargraph
     # @return [Array<String>]
     attr_reader :unresolved_requires
 
+    # @return [Array<String>]
+    attr_reader :missing_docs
+
     # @param pins [Array<Solargraph::Pin::Base>]
     def initialize pins: []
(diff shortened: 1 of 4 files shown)


castwide:
Awesome, thanks a lot!

castwide:
Just a heads up, one of the tests was broken. I made some changes to it and modified it with a comment after manually confirming that missing YARD docs were reported correctly.

dgmora:
I didn't update them because I wasn't sure if it'd be accepted, but thank you!

Issue #135, opened 2019-01-11
Renaming fails in vscode


macifell:
I'm really enjoying this extension so far, but I have noticed an issue. My attempts to rename a symbol, method, class/module, etc.. almost always fail in vscode. The only time I have had success is right after opening a new folder.

This error message is displayed: `Overlapping ranges are not allowed!`

The console log shows:
```
  ERR Error: Overlapping ranges are not allowed!
    at e.applyEdits (pieceTreeTextBuffer.ts:216)
    at o._applyEdits (textModel.ts:1314)
    at o.applyEdits (textModel.ts:1271)
    at e.pushEditOperation (editStack.ts:168)
    at o._pushEditOperations (textModel.ts:1264)
    at o.pushEditOperations (textModel.ts:1173)
    at t.executeEdits (codeEditorWidget.ts:989)
    at t.apply (bulkEditService.ts:120)
    at e.apply (bulkEditService.ts:213)
    at e.<anonymous> (bulkEditService.ts:355)
    at errors.ts:237
    at Object.s.__generator.a.label [as next] (errors.ts:237)
    at s (errors.ts:237)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188)
```

vscode:
```
Version: 1.30.2
Commit: 61122f88f0bf01e2ac16bdb9e1bc4571755f5bd8
Date: 2019-01-07T22:49:48.319Z
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
OS: Linux x64 4.15.0-43-generic
```

vscode ruby extensions:
```
Ruby 0.21.0
Ruby Solargraph 0.19.1
```

vscode settings.json:
```
{
...
  "ruby.lint": {
    "rubocop": true,
    "ruby": true
  },
  "solargraph.formatting": true,
...
}
```

solargraph:
```
$ solargraph --version
0.30.2
$ solargraph list-cores
2.5.3
2.2.2
$ solargraph reporters
rubocop
require_not_found
type_not_defined
update_errors
```

ruby:
```
$ ruby --version
ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-linux]
```

castwide:
Confirmed in v0.30.2. There's a fix for it in the `v0.31.0` branch, which I expect to publish this week.

macifell:
nice! :slightly_smiling_face:

castwide:
[...]

Issue #107, opened 2018-11-19
Symbol rename issue


whitebyte:
Hi, I'm experiencing an issue with Solargraph symbol renaming. Consider this minimal example.rb:

```ruby
class TestClass
  SOME_CONST = 123
end
```

When I try to rename  `TestClass` to `TestClass2` it becomes `TTestClass2` instead. When I try to rename `SOME_CONST` to `SOME_CONST2` it becomes `SOMESOME_CONST2`.

I'm on Linux, Neovim, using https://github.com/autozimu/LanguageClient-neovim, Solargraph 0.28.4 started in stdio mode. I use exactly the same setup for various language servers with no issues, so I do believe it's related to Solargraph

castwide:
Confirmed. The same thing happens in the VS Code extension, so it looks like a problem in Solargraph itself. I'll look into it.

castwide:
The fix is in the master branch. I'll release it in version 0.29.0.

whitebyte:
Wow, that was fast! Closing then, thanks much

whitebyte:
Unfortunately, win 0.29.1 the problem persists. While it works OK for more simple cases (like my previous example), it fails when there is a namespace prefix:

```ruby
class Namespace::TestClass
  # whatever
end
```
Attempt to rename either a namespace part of a class part of the signature leads to malformed names

castwide:
Does this problem still exist with gem 0.31.3? I tested with gVim and vim-lsp. Given the following code:

```ruby
module Namespace; end
class Namespace::TestClass
end
obj = Namespace::TestClass.new
```

Using `:LspRename` to change `Namespace` to `OtherName` worked as expected in all three occurrences.

PR #573, opened 2022-07-12
added failing (and passing?) tests

BonitaTerror:
Hello, I've been following along with a couple old issues some have reported, and that I have personally experienced with renaming.

The respective issues I think are related are #135 and #107.

I am not familiar with language server protocol, but was able to add a few tests that correlate with the renaming bug that I and others have had in our respective editors. The first 2 tests are related to comments at the bottom of #135, and the extra element in `rename.result[:changes]['file:///file.rb']` I believe is the 'overlapping ranges error'. Both of those tests are currently failing.

The final test at the bottom is for #107, but seems to be passing, so it's just an added test at this point.

If time permits, I'll come back soon and figure out how to get these tests to pass, but wanted to post the PR in case it sparks an idea for anyone more familiar with LSP.


castwide:
These tests are succeeding now, so I'll just go ahead and merge them.

PR #576, opened 2022-08-10
make rubocop info severity Severity::HINT

maxnotarangelo:
At the moment, solargraph doesn't handle rubocop's "info" severity level. As a result, if you specify that a rubocop rule has the "info" severity level, it will default to the highest severity level. In a code editor like vscode, you'll get a red underline.

For example, in my project, I changed rubocop's line length rule to have the "info" severity level. Long lines then get a red underline:

Screen Shot 2022-08-09 at 9 24 12 PM


dleavitt:
Affected by this as well, seems like a pretty safe merge! See: https://github.com/rubocop/rubocop/pull/9391

PR #581, opened 2022-08-28
Process non-self singleton classes

k0kubun:
To avoid deep indentation while still using `private` and skipping `def self.`, I sometimes declare a class and define singleton methods on it with `class << Example = Class.new`. I'd like Solargraph to process its children.

```rb
class << Example = Class.new
  def foo
    bar
  end

  private

  def bar
    "example"
  end
end
```

The current `SclassNodes` [supports only `class << self`](https://github.com/castwide/solargraph/commit/b0852417a1d9f8d44e4ad8db05b0465325b5b66c), so I extended it to handle the above `CDECL` case as well as `CONST` case, .e.g. `class << Example`.

### DIFF
--- a/lib/solargraph/parser/rubyvm/node_processors/sclass_node.rb
+++ b/lib/solargraph/parser/rubyvm/node_processors/sclass_node.rb
@@ -6,11 +6,22 @@ module Solargraph
       module NodeProcessors
         class SclassNode < Parser::NodeProcessor::Base
           def process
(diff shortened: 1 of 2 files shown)


k0kubun:
@castwide Any thoughts? I'll close this if you're uncomfortable with supporting this feature. I use this syntax in many of my projects and it's been painful whenever solargraph without this is installed.

castwide:
I've been meaning to support this for a while. Thanks for taking care of it.

PR #585, opened 2022-09-01
Strip 'file ' prefix from all filenames in RdocToYard

### LINKED ISSUE #473: RDocToYard overwrites previously defined YARD objects, even if they aren't defined in RDoc
EDIT: this issue is quite far from where it started, look [here](https://github.com/castwide/solargraph/issues/473#issuecomment-921930024) for the latest.

====

[`api_map.get_complex_type_methods` returns empty for `ActiveRecord::Associations::CollectionProxy<MyModel>`:

```
> type = Solargraph::ComplexType.parse('ActiveRecord::Associations::CollectionProxy<Book>')
#<Solargraph::ComplexType:0x0000000107607ba8 @items=[#<Solargraph::ComplexType::UniqueType:0x0000000107607ec8 @name="ActiveRecord::Associations::CollectionProxy", @rooted=false, @substring="<Book>", @tag="ActiveRecord::Associations::CollectionProxy<Book>", @key_types=[], @subtypes=[#<Solargraph::ComplexType::UniqueType:0x0000000107607c48 @name="Book", @rooted=false, @substring="", @tag="Book", @key_types=[], @subtypes=[], @duck_type=false>], @duck_type=false>]>
> type_methods = api_map.get_complex_type_methods(type).reject { |p| p.closure.name =~ /Object|Expectations/ }.map(&:name)
[]
```

I expect at least the methods described here: https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html like `.any`, `.first` etc.

compared to the same thing with an Array of MyType:

```
> type = Solargraph::ComplexType.parse('Array<Book>')
#<Solargraph::ComplexType:0x0000000118a11be8 @items=[#<Solargraph::ComplexType::UniqueType:0x0000000118a11f08 @name="Array", @rooted=false, @substring="<Book>", @tag="Array<Book>", @key_types=[], @subtypes=[#<Solargraph::ComplexType::UniqueType:0x0000000118a11c88 @name="Book", @rooted=false, @substring="", @tag="Book", @key_types=[], @subtypes=[]>], @duck_type=false, @nil_type=false, @scope=:instance, @namespace="Array">], @namespace="Array">
> type_methods = api_map.get_complex_type_methods(type).reject { |p| p.closure.name =~ /Object|Expectations/ }.map(&:name)
["&", "*", "+", "-", "<<", "<=>", "==", "[]", "[]=", "any?", "assoc", "at", "bsearch", "clear", "collect", "collect!", "combination", ...]
```

Steps to reproduce:

1. start Rails console in a Rails app with Book model
2. Create api_map: `api_map = Solargraph::ApiMap.load(Rails.root)
3. type = ... (follow from above)

grncdr:
This fixes #473 (at least for me, see extended notes below).

The core issue is that the `obj.filename` property would sometimes be prefixed with the string "file ", which then causes `YardMap::Mapper#map` to remove the pins: https://github.com/castwide/solargraph/blob/18854d7577737cf9929063d96dd99b3d0a77f703/lib/solargraph/yard_map/mapper.rb#L26

This change adds the prefix stripping regardless of whether we get a filename from `in_files` or `filename`.

<details>
<summary>
lots of notes about debugging this, feel free to skip
</summary>

Reviving this thread, as I'm trying to fix https://github.com/iftheshoefritz/solargraph-rails/issues/34. Here is what I observe:

## Preamble

```
=> bundle exec solargraph --version
0.46.0

=> bundle exec yard --version
yard 0.9.28

=> bundle exec rails --version
Rails 7.0.2.4
```

## Part 1 - generating YARD objects

Starting from scratch:

```
bundle exec yard gems
bundle exec solargraph bundle
```

```rb
require 'yard'
yardoc_file = "#{ENV['HOME']}/.solargraph/cache/gems/activerecord-7.0.2.4/yardoc"
YARD::Registry.load_yardoc(yardoc_file)
YARD::Registry.all.select { |o| o.namespace.name == :QueryMethods }
#=> [#<yardoc class ActiveRecord::QueryMethods::WhereChain>,
 #<yardoc method ActiveRecord::QueryMethods#and>,
 #<yardoc method ActiveRecord::QueryMethods#annotate>,
 #<yardoc method ActiveRecord::QueryMethods#create_with>,
 #<yardoc method ActiveRecord::QueryMethods#distinct>,
 #<yardoc method ActiveRecord::QueryMethods#eager_load>,
 #<yardoc method ActiveRecord::QueryMethods#excluding>,
 #<yardoc method ActiveRecord::QueryMethods#extending>,
 #<yardoc method ActiveRecord::QueryMethods#extract_associated>,
 #<yardoc method ActiveRecord::QueryMethods#from>,
 #<yardoc method ActiveRecord::QueryMethods#group>,
 #<yardoc method ActiveRecord::QueryMethods#having>,
 #<yardoc method ActiveRecord::QueryMethods#in_order_of>,
 #<yardoc method ActiveRecord::QueryMethods#includes>,
 #<yardoc method ActiveRecord::QueryMethods#invert_where>,
 #<yardoc method ActiveRecord::QueryMethods#joins>,
 #<yardoc method ActiveRecord::QueryMethods#left_joins>,
 #<yardoc method ActiveRecord::QueryMethods#left_outer_joins>,
 #<yardoc method ActiveRecord::QueryMethods#limit>,
 #<yardoc method ActiveRecord::QueryMethods#lock>,
 #<yardoc method ActiveRecord::QueryMethods#none>,
 #<yardoc method ActiveRecord::QueryMethods#offset>,
 #<yardoc method ActiveRecord::QueryMethods#optimizer_hints>,
 #<yardoc method ActiveRecord::QueryMethods#or>,
 #<yardoc method ActiveRecord::QueryMethods#order>,
 #<yardoc method ActiveRecord::QueryMethods#preload>,
 #<yardoc method ActiveRecord::QueryMethods#readonly>,
 #<yardoc method ActiveRecord::QueryMethods#references>,
 #<yardoc method ActiveRecord::QueryMethods#reorder>,
 #<yardoc method ActiveRecord::QueryMethods#reselect>,
 #<yardoc method ActiveRecord::QueryMethods#reverse_order>,
 #<yardoc method ActiveRecord::QueryMethods#rewhere>,
 #<yardoc method ActiveRecord::QueryMethods#select>,
 #<yardoc method ActiveRecord::QueryMethods#strict_loading>,
 #<yardoc method ActiveRecord::QueryMethods#structurally_compatible?>,
 #<yardoc method ActiveRecord::QueryMethods#uniq!>,
 #<yardoc method ActiveRecord::QueryMethods#unscope>,
 #<yardoc method ActiveRecord::QueryMethods#where>,
 #<yardoc method ActiveRecord::QueryMethods#without>]
```

Good start, the cached documentation in `~/.solargraph/cache` is contains the expected method definitions.

However, `solargraph scan -v` does not show e.g. `QueryMethods#where`:

```sh
bundle exec solargraph scan -v | grep 'QueryMethods#where'
# no output
```

So it seems as though somewhere the yardoc from `~/.solargraph/cache` is either not loaded, or is unloaded.

## Part 2 - when do we load the cache?

Usages of `cache_dir`

1. https://github.com/castwide/solargraph/blob/87f627551941d50f58aba42d466e426ca606de75/lib/solargraph/yard_map.rb#L309
2. [yardoc_file_for_spec](https://github.com/castwide/solargraph/blob/87f627551941d50f58aba42d466e426ca606de75/lib/solargraph/yard_map.rb#L348-L356)

The correct (cached) yardoc is found, because it gets loaded when I break in YardMap#process_yardoc with the condition `y =~ /activerecord/`

The yardoc contains the method objects at the time the yardoc is loaded (checked by breaking in `YardMap::Mapper#map` when `co.name == :where`)

However the `@pins.keep_if` line is removing 1596 pins, because many of the pins have a `location.filename` that looks like "<correct_gem_path>/activerecord-7.0.2.4/**file lib**/active_record/" (emphasis on the "file " in there).

Eventually we call `process_yardoc`

So, I hacked up YardMap::Helpers#object_location to strip the "file " prefix from `code_object.file` like so:

```diff
       def object_location code_object, spec
         return nil if spec.nil? || code_object.nil? || code_object.file.nil? || code_object.line.nil?
-        file = File.join(spec.full_gem_path, code_object.file)
+        file = File.join(spec.full_gem_path, code_object.file.sub(/^file /, ''))
         Solargraph::Location.new(file, Solargraph::Range.from_to(code_object.line - 1, 0, code_object.line - 1, 0))
       end
```

After this `solargraph scan -v` shows the missing methods, and auto-complete works!

## Part 3 - a real fix?

So it's pretty weird that yard code objects contain an invalid path like that. Especially considering that these code objects _should_ have been produced by Solargraphs own RdocToYard conversion. Taking a peek at `RdocToYard.find_file` reveals the issue:

```rb
      def self.find_file obj
        if obj.respond_to?(:in_files) && !obj.in_files.empty?
          [obj.in_files.first.to_s.sub(/^file /, ''), obj.line]
        else
          [obj.file, obj.line]
        end
      end
```

Turns out there was already code to strip the "file " prefix, but it only exists in one of the two branches. Hence this PR.
</details>


iftheshoefritz:
@grncdr this does seem to fix #473, and I'm excited to see it go in! 

However I'm not seeing what I'd expect if https://github.com/iftheshoefritz/solargraph-rails/issues/34 were fixed. It looks like when I autocomplete `my_associated_models.first`, there is still no type on `.first`.

```
    {
      "label": "first",
      "kind": 2,
      "detail": "(limit = nil)",
      "data": {
        "path": "ActiveRecord::Associations::CollectionProxy#first",
        "return_type": "undefined",
        "location": {
          "filename": "/Users/fritz/.asdf/installs/ruby/3.0.0/lib/ruby/gems/3.0.0/gems/activerecord-7.0.1/lib/active_record/associations/collection_proxy.rb",
          "range": {
            "start": {
              "line": 141,
              "character": 0
            },
            "end": {
              "line": 141,
              "character": 0
            }
          }
        },
        ...
```

What I expect is `"return_type": "MyOtherModel"` in the response from solargraph.

grncdr:
> I'm not sure though if this change could solve that? What is the behaviour that you observe?

I don't expect it to. As far as I know, the original Rdoc doesn't specify a return type and so it needs augmentation on the part of solargraph-rails.

I've fixed that (and loads of other stuff) in my ["hacks" branch of solargraph-rails](https://github.com/iftheshoefritz/solargraph-rails/compare/main...grncdr:solargraph-rails:hacks), but that branch depends on some of my other changes to solargraph (which are in my ["hacks" branch for solargraph](https://github.com/castwide/solargraph/compare/master...grncdr:solargraph:hacks)).

If you switch to using those branches in your Gemfile, you'll get a **very** improved solargraph-rails experience, but it all needs more specs and especially some feedback on whether my Pin::DelegatedMethod PR (#602, #591) is on the right track.

castwide:
Thanks!

PR #588, opened 2022-09-04
Update YARD tags on Pin::Block methods

grncdr:
I was debugging and noticed that these tags are inaccurate. The `@parameters` array in a `Pin::Block` contains `Pin::Parameter` instances.


castwide:
:+1:

PR #590, opened 2022-09-05
Fix parsing of nested subtypes

### LINKED ISSUE #589: ComplexType#self_to is removing the final '>' from the dst param
Simple to reproduce, doesn't seem to depend on anything about the actual project.

```rb
# this does not raise an error
Solargraph::ComplexType.parse('Array').self_to("Foo<String>")

# this raises ComplexTypeError, "Unclosed subtype in Foo<String"
Solargraph::ComplexType.parse('Array<self>').self_to("Foo<String>")
```

grncdr:
Fixes #589

### DIFF
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -154,9 +154,9 @@ module Solargraph
                 subtype_string.clear
                 next
               else
+                raise ComplexTypeError, "Invalid close in type #{type_string}" if point_stack == 0
                 point_stack -= 1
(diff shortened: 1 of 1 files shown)


castwide:
Thanks! This could use a unit test, but I can take care of it. Doesn't have to be more complex than this:

```ruby
  it 'parses a complex subtype' do
    type = Solargraph::ComplexType.parse('Array<self>').self_to('Foo<String>')
    expect(type.tag).to eq('Array<Foo<String>>')
  end
```

PR #602, opened 2022-09-29
Add Pin::DelegatedMethod

grncdr:
This is a more complicated version of a method alias, where the type of the method receiver can be resolved dynamically using a Source::Chain.

The idea didn't get much of a response in #591, but I really wanted to implement proper support for the `Module.delegate` method from ActiveSupport. You can see how it's used in my solargraph-rails work [here](https://github.com/grncdr/solargraph-rails/commit/f53fe4889fb50fb7da4be6811b6a73e021a8ad23).

I'd really like to get this merged, and I'm happy to add specs for it, but I want to get feedback on the implementation, as it's a pretty "unusual" pin type.

### DIFF
--- a/lib/solargraph/pin.rb
+++ b/lib/solargraph/pin.rb
@@ -12,6 +12,7 @@ module Solargraph
     autoload :Method,           'solargraph/pin/method'
     autoload :Signature,        'solargraph/pin/signature'
     autoload :MethodAlias,      'solargraph/pin/method_alias'
+    autoload :DelegatedMethod,  'solargraph/pin/delegated_method'
     autoload :BaseVariable,     'solargraph/pin/base_variable'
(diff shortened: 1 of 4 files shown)


grncdr:
@castwide not sure if you get notified for draft PRs, but I would definitely appreciate your thoughts on this one before moving forward.

castwide:
Sorry for the long delay. This looks like a solid start. It might also be useful for implementing proper support of the `!@macro` directive. I'll pull it down and play with it, but right now I don't see any red flags.

grncdr:
Hi @castwide, I've just rebased my branch on the latest `master`, did you have a chance to try it out?

castwide:
This one's long overdue. Thanks, @grncdr!

PR #604, opened 2022-10-06
Add Sublime Text to README

themilkman:
Hi @castwide,

first of all: Thank you so much for all your work in solargraph!

I use it with Sublime Text since a long time and it works really nice. Thus, I thought ST could be added to the Readme file.

Cheers!


castwide:
Thanks!

Issue #464, opened 2021-07-24
RBS Support


castwide:
Integration of RBS into Solargraph is officially on the roadmap.

## Features

* Map the Ruby core and stdlib with RBS
* Map gems with RBS when sigs are available
* Fall back to YARD for gems that do not provide sigs

## Pros

* **Ostensibly faster.** In my early experiments, mapping from RBS instead of YARD appears to provide a slight performance boost.
* **Better support.** Updates to core and stdlib signatures are likely to be more frequent and more accurate.
* **No need for core docs.** It will no longer be necessary to use `solargraph download-core` to update core and stdlib documentation.
* **YARD will continue to be supported.** RBS sigs will be treated as the canonical authority for external libraries when they're available, but Solargraph will still use YARD annotation in workspace maps.

## Cons

* **Requires Ruby 2.6+.** Solargraph current supports Ruby 2.4+. With integration of RBS, 2.6 will become the minimum requirement.
* **Major rearchitecture.** The concepts of maps and pins will undergo a thorough redesign.
* **Missing features.** Some code intelligence provided by YARD may not be currently available in RBS. One known example: [RBS does not support protected methods](https://github.com/ruby/rbs/issues/579).

Current progress can be previewed on the repo's `rbs` branch.

dleal-mojotech:
Do you have any plans to support generating RBS files from YARD docs?

Thanks for all the work. This gem is a huge help!

castwide:
@dleal-mojotech Thanks!

If I do, it will probably be based on [Sord](https://github.com/AaronC81/sord).

wagenet:
I've been using sord with a decent degree of success.

tmszcncl:
Any update on rbs support? rbs is working really well on rubymine and I miss it in vscode :)

castwide:
It's still in progress. I just updated the `rbs` branch with a few commits I hadn't pushed yet. I can't give an exact ETA yet, but certainly before the end of the year.

steakknife:
[steep](https://github.com/soutaro/steep) seems to be the way things are going with rbs. I use RubyMine at home but work uses something resembling VSCode. I like the idea of rbs but I'm honestly wondering if there should've been a language break at 3 to support gradual typing like Crystal and Python.

castwide:
The `rbs` branch is merged into `master` and will be released shortly. Some notes on the current version:

* The core and stdlib maps are generated from RBS. This means it is no longer necessary to download or generate core documentation.
[...]

PR #606, opened 2022-10-10
Stop using deprecated rubocop args

baelter:


baelter:
Fixes https://github.com/castwide/vscode-solargraph/issues/260

castwide:
This is a little tricky at the moment, because Solargraph currently supports Ruby >= 2.4 and RuboCop >= 0.52. That problem is likely to go away soon as we get closer to releasing [RBS support](https://github.com/castwide/solargraph/issues/464), at which point the minimum Ruby will be 2.6 and we can update the RuboCop dependency accordingly.

PR #618, opened 2022-12-08
fix: hover and jump for union type

### LINKED ISSUE #617: Bug: Union Type no hover and can't jump for latter type
```ruby
class Array
  # custom doc
  def custom
  end
end

class Thing
  # @return [String, Array]
  def foo; end
end
Thing.new.foo.custom # <-- no hover, can't jump
Thing.new.foo.upcase # has hover info

```

try this code, and custom in Array can complete, but no hover, and no jump

commit: 0a520aeec1c855add1b4b0c2d2846b0787866c1a
newest HEAD

SolaWing:
fixes #617

seems fine on mine case. not sure fixes all cases..

### DIFF
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -40,6 +40,16 @@ module Solargraph
       @items.each &block
     end
 
+    # @yieldparam [UniqueType]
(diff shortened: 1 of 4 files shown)


castwide:
Tested locally, works as expected. Thanks!

PR #620, opened 2022-12-09
fix: offset error for unicode

### LINKED ISSUE #539: undefined method `offset' for nil:NilClass
Got this in vscode, some missing error handling? 

```[Error - 10:24:41 AM] Request textDocument/documentHighlight failed.
  Message: [NoMethodError] undefined method `offset' for nil:NilClass

            extract_offset = ->(code, offset) { reg.match(code, offset).offset(0) }
                                                                       ^^^^^^^
  Code: -32603 
```

SolaWing:
Character have different length in different standards, and will affect how offset calculated

for example, for a string "𐐀":

`String#length` returns 1 (utf32 character length, no matter which encoding used) `RubyVM::AbstractSyntaxTree.parse("𐐀").children[2].last_column` return 4 (utf8 bytes size)
[LSP Spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments) default use utf16, which length is 2. ( 4 bytes size)

if the standard does not match, offset will be wrong and get wrong result.

Fortunately, most char utf16 and utf32 char lengths are the same, and rarely cause problems. but In the unicode environment, the byte length and character length are generally different, and will report error

Because Ruby String handle char instead of bytes. so the offset all should be char length based

Since AbstractSyntaxTree::Node is passed to many place and hard to control, the simplest repair is to wrap it when create and return a character based column

this should fixes #539 and maybe other unicode related offset errors

if LSP client support UTF32, notify it use utf-32 position [capabilities.positionEncoding](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities), then all position should be same.
But same client may be only support default utf-16 position...

### DIFF
--- a/lib/solargraph/language_server/message/initialize.rb
+++ b/lib/solargraph/language_server/message/initialize.rb
@@ -25,6 +25,8 @@ module Solargraph
               }
             }
           }
+          # FIXME: lsp default is utf-16, may have different position
+          result[:capabilities][:positionEncoding] = "utf-32" if params.dig("capabilities", "general", "positionEncodings")&.include?("utf-32")
           result[:capabilities].merge! static_completion unless dynamic_registration_for?('textDocument', 'completion')
           result[:capabilities].merge! static_signature_help unless dynamic_registration_for?('textDocument', 'signatureHelp')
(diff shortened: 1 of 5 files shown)


castwide:
Thanks!

PR #627, opened 2023-01-06
Remove .travis.yml as its not longer used

gwillcox-r7:
This PR removes pre 2.6 versions of Ruby as the minimum version of Ruby that is supported will be bumped to 2.6 once the RBS changes land. Additionally I noticed that whilst I don't think we officially support Ruby 3.2 yet, we do support Ruby 3.1 testing already in the GitHub Actions, yet Travis isn't yet supporting this. For this reason I went ahead and added Ruby 3.1 into the testing array.

Let me know if there are any concerns r.e this. Not sure if Travis is still used by this project or if things moved over to GitHub Actions, in which case it may be better to just remove this file to avoid confusion. Alternatively if there are concerns, please let me know and I'll do my best to try and resolve them.


castwide:
Is there any value in keeping the .travis.yml file anymore? I stopped using Travis CI after moving to GitHub Actions, but I don't mind keeping it if anyone else finds it useful.

gwillcox-r7:
> Is there any value in keeping the .travis.yml file anymore? I stopped using Travis CI after moving to GitHub Actions, but I don't mind keeping it if anyone else finds it useful.

I'd be happy to remove it. I was thinking it might still be in use somewhere but if your not using it its probably better to just remove it flat out then. I can update this PR to just remove the file instead if that would be preferred?

castwide:
Sounds good :+1:

PR #629, opened 2023-01-25
Define `e` in `JSON::ParserError` rescue block

keithdoggett:
The `JSON::ParserError` rescue statement in the Rubocop `diagnose` method does not pass the exception to the block, but the block calls `e.message` which causes the server to crash with an `rescue in 'diagnose': undefined method 'message' for nil:NilClass (NoMethodError)` error. This PR simply passes `e` to the block.


castwide:
:+1:

Issue #631, opened 2023-02-04
Error processing private_constant


hallerm:
A typo in source code such as:

```ruby
  class Oops; end
  private_constant(&:Oops)
```

Will cause solargraph 0.48.0 to terminate parsing with the following error:

```
[Extension Host] [WARN] Error processing request: [NoMethodError] undefined method `type' for nil:NilClass

console.ts:137 [Extension Host] [WARN] /opt/homebrew/lib/ruby/gems/2.6.0/gems/solargraph-0.48.0/lib/solargraph/parser/rubyvm/node_processors/send_node.rb:235:in `block in process_private_constant'
/opt/homebrew/lib/ruby/gems/2.6.0/gems/solargraph-0.48.0/lib/solargraph/parser/rubyvm/node_processors/send_node.rb:234:in `each'
/opt/homebrew/lib/ruby/gems/2.6.0/gems/solargraph-0.48.0/lib/solargraph/parser/rubyvm/node_processors/send_node.rb:234:in `process_private_constant'
```

PR #632, opened 2023-02-04
Fix corner case handling private_constant

hallerm:
If someone accidentally gives private_constant a block, such as `private_constant(&:Oops)`, then solargraph would crash while handling the arguments. This change corrects the issue.

Fix castwide/solargraph#631


castwide:
:+1:

PR #638, opened 2023-03-03
Batch processing of changed files for improved performance

naveg:
When multiple files change on disk (eg after a git branch change),
solargraph does a re-index operation for each change, because the
changes are processed individually. Indexing is a slow operation that
can take 10s of seconds on a very large codebase. These codebases are
also likely to have 10s or even 100s of changed files when checking out
a different git branch. This led to solargraph being very slow.

We can instead process all of the changed files in one go, requiring
only a single re-index.

This change also removes the `would_merge?` method, which is
unnecessary and redundant with `merge`, which already checks if a source
should be included.

### DIFF
--- a/lib/solargraph/language_server/host.rb
+++ b/lib/solargraph/language_server/host.rb
@@ -130,35 +130,39 @@ module Solargraph
         end
       end
 
-      # Respond to a notification that a file was created in the workspace.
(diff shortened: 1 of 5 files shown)


castwide:
Looks great, thanks! I noticed that the `Library#create` method might also be redundant now, but that can be handled as a separate issue.

PR #644, opened 2023-04-04
Fix empty string case when processing requires

giorni:
My codebase has a require to an absolute path (`/opt/to/something/unusual`). Finding spec for it uses a nil value (spliting empty string returns emtpy array) that breaks `Gem::Specification.find_by_name`, therefore, scans fail.


castwide (approved):
:+1:

PR #661, opened 2023-05-14
Fix spec version comparison

grncdr:
The return type of spec.version is a Gem::Version, which can be compared to a string, but not the other way around.


castwide:
:+1: :shipit:

Issue #652, opened 2023-04-18
Formatting no longer works with Rubocop < 1.30


noniq:
https://github.com/castwide/solargraph/commit/c195d3f237e0f6ca876bb90d6206181bcdd5d706 changed `--auto-correct` to `--autocorrect` to avoid deprecation warnings with recent versions of Rubocop. However, this broke formatting with Rubocop versions < 1.30, as those do not support `--autocorrect`.

Error message:

```
[Error - 20:06:28] Request textDocument/formatting failed.
  Message: [OptionParser::InvalidOption] invalid option: --autocorrect
Did you mean?  auto-correct
  Code: -32603 
```

A possible fix would be to simply use `-a` instead, this works in all Rubocop versions.

PR #665, opened 2023-06-06
modify rubocop option for rubocop versions < 1.30

hyuraku:
closes https://github.com/castwide/solargraph/issues/652


castwide:
:+1:

PR #677, opened 2023-08-18
Fix idle cpu usage

hmsf:
I've noticed the cpu usage has been pretty high, around 10% when idling.

I believe this issue may have originated from the change in this [commit](https://github.com/castwide/solargraph/pull/136/commits/ac3f0628dcd67d80c7f4dc693fb6e5b1bc8be27b#), which altered the sleep duration from 0.1 to 0.01 seconds.

Changing it back to 0.1 seconds has returned the cpu usage to 1-2%.


castwide:
LGTM :+1:

PR #693, opened 2023-10-23
Add rbs mention to readme

Earlopain:
Updating core documentation is not needed with the current version of the gem anymore.


castwide:
Thanks!

PR #697, opened 2023-11-25
Conversion between dashes in gemspec name and slashes in required paths

kosuke-suzuki:
I was faced with a strange behavior and finally found the possible root cause of it.

# Background

As recommended by Rubygems.org ["Name your gem"](https://guides.rubygems.org/name-your-gem/), there is a conventional correspondence between dashes `-` in a gem name and slashes `/` in a required path.

For example, if you want to use `net-smtp` gem,
first you have to add to your Gemfile
```ruby
gem "net-smtp"
```
and second you require it:
```ruby
require "net/smtp"

# Do anything fancy with Net::SMTP
Net::SMTP.start('127.0.0.1')
```

# Problem

However, I noticed that solargraph did not suggest to me the source code comments on `net-smtp`.
In the code above I expected solargraph to show me the signature of `Net::SMTP.start()` but it did not.

I found that the root cause of this phenomenon was how solargraph treated `require`d paths.
Currently it regards the first part of slash-separated `require`d path as the name of the gem.
https://github.com/castwide/solargraph/blob/master/lib/solargraph/yard_map.rb#L285
```ruby
name = path.split('/').first.to_s
```
Therefore, when it finds
```ruby
require "net/smtp"
```
sentence, the regarded gem name is not `net-smtp` but `net`.

So there is no way for solargraph to detect gems whose names are dash-concatenated and whose actual relative paths are NOT dash-concatenated but slash-separated.

# Proposal

Here I propose another algorithm to determine the name of the gem.

1. Split the `require`d path by slashes `/`
2. Concatenate all by dashes `-` and let it be the candidate of the name of the gem
3. Check if there is such gem
4. If it exists, it is the actual name of such gem. If it does not, omit the last part of the dashes and go back to step 3

For example, with this algorithm, when solargraph encounters the line
```ruby
require "net/smtp/authenticator"
```

1. `["net", "smtp", "authenticator"]`
2. `net-smtp-authenticator`
3. Does the gem with name `net-smtp-authenticator` exist? No it does not.
4. Does the gem with name `net-smtp` exists? Yes it does.

After I applied this algorithm to my environment, solargraph showed me source code comments in `Net::SMTP` as I expected.

Any possible drawbacks?

### DIFF
--- a/lib/solargraph/yard_map.rb
+++ b/lib/solargraph/yard_map.rb
@@ -220,9 +220,11 @@ module Solargraph
       result = []
       begin
         name = r.split('/').first.to_s
-        return [] if name.empty? || @source_gems.include?(name) || @gem_paths.key?(name)
-        spec = spec_for_require(name)
-        @gem_paths[name] = spec.full_gem_path
+        return [] if name.empty?
+
+        spec = spec_for_require(r)
+        return [] if @source_gems.include?(spec.name) || @gem_paths.key?(spec.name)
+        @gem_paths[spec.name] = spec.full_gem_path
 
         yd = yardoc_file_for_spec(spec)
         # YARD detects gems for certain libraries that do not have a yardoc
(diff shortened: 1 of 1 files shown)


castwide:
:+1: :shipit:

PR #698, opened 2023-12-12
Add ruby 3.2 to CI

Earlopain:
Bumps the github actions, and especially setup-ruby since it didn't know about ruby 3.2 beforehand.

Uses a tag for setup-ruby, for convenience. It will not need to be updated for ruby 3.3 again, and will pick up the latest patch version available at runtime.


Earlopain:
Nevermind, duplicate of #659

PR #702, opened 2023-12-27
Fixed Solargraph::Cache.base_dir

earl-gresh:
Switched Ruby's || operator to 'or' to account for precedence.

When the environment variable SOLARGRAPH_CACHE is defined, the cache is always configured to use XDG_CACHE_HOME. When XDG_CACHE_HOME is defined, it silently replaces SOLARGRAPH_CACHE. When it is undefined, it leads to a TypeError.


castwide:
This bug was addressed in #704. If there's some issue I missed, please let me know.

PR #714, opened 2024-05-25
Complete global methods from a file inside modules/classes

lekemula:
Hi, 

I noticed this issue while trying to make the "normal ruby" method completion work for RSpec `context` blocks initially in https://github.com/lekemula/solargraph-rspec. Even though I ultimately ended up with a different solution, I thought to still port this change as it's still a valid ruby code, albeit a rare case in real life (?).

```ruby
# thing.rb
def some_method;end

class Thing
  def foo
    some_
    #   ^ this does not work
  end
end
```


castwide:
:+1: :shipit:

Issue #683, opened 2023-09-08
RSPEC support


reeganviljoen:
I know their is support for Rails with `solargraph` through the `solargraph-rails` plugin , is their a `solargraph-rspec` plugin or built in support I can use for `rspec`

castwide:
There's an RSpec convention that currently ships with Solargraph, but it's far from complete. It's definitely a good candidate to become a separate extension.

reeganviljoen:
@castwide how would that candidacy process work for solargraph

castwide:
There's no official candidacy process. When I have time to work more on RSpec support, one of my first tasks would probably be to move it to its own extension. If someone else is interested in working on it, I'd be willing to assist.

lekemula:
Hey @reeganviljoen @castwide I've got some good news for you here! After playing around for a bit I've been able to put some work and created an extension gem for RSpec here: https://github.com/lekemula/solargraph-rspec.

Luckily after some iterations it ended up not needing any patches here, although I still think some of the initial changes would [there](https://github.com/castwide/solargraph/compare/master...lekemula:solargraph:rspec-support?expand=1) be still beneficial. @castwide WDYT? I would also appreciate any feedback on the gem itself as I'm a complete newbie in this domain. 🙏 

Please give it a try and let me know how it works for you!

Cheers!

PR #716, opened 2024-05-26
Remove Rspec convention in favour of solargraph-rspec plugin

lekemula:
Hi,

ICYMI: https://github.com/castwide/solargraph/issues/683#issuecomment-2106296460

The new plugin includes all existing functionality + much more! (See gem's README: https://github.com/lekemula/solargraph-rspec.)

After you have given it a try and find it worthy, being that `rspec` is [the de-facto preferred testing framework](https://railsdeveloper.com/survey/2022/#what-rails-testing-frameworks-are-you-using-to-write-tests) in Ruby, would it make sense to:

- mention it alongside `solargraph-rails` in the Readme/Documentation?
- ~~remove the `exclude: ['spec/**/*']` folder as a default config? 
https://github.com/castwide/solargraph/blob/d610a6bb77f2f95f9e01b70e79e625958b66e523/lib/solargraph/workspace/config.rb#L151~~

Looking forward to your feedback.

Resolves: https://github.com/castwide/solargraph/issues/683


lekemula:
> remove the exclude: ['spec/**/*'] folder as a default config?

@castwide  It turns out that keeping the `specs` folder in the `exclude` list, is a good idea. Luckily Solargraph still parses them when opened or changed in the editor, something which I initially thought was a bug, but in our case, it's a very important feature allowing us to "lazy parse" specs 🥳 

https://github.com/lekemula/solargraph-rspec/commit/3f0fc39e59e99bf9430e55c52ecb88650e49315e

It's worth mentioning that with this setup, there was no performance issues or degradation whatsoever when testing against [Gitlab's](https://gitlab.com/gitlab-org/gitlab/-/blob/master/.solargraph.yml.example?ref_type=heads) codebase.

castwide:
LGTM. Thanks!

PR #718, opened 2024-06-09
Include @example tags in Pin::Method documentation

lekemula:
When trying to [add yardoc comments for some of RSpec DSL](https://github.com/lekemula/solargraph-rspec/commit/a08427ad6544b6e72abaab7e4d8081f2093293ec) I noticed that `@example` tags are not being included. And for RSpec DSL those seem to be the most useful part. I believe examples are quite helpful in general for devs to see how a particular method is used.

Before:

image_1717952008834_0

After:

image_1717951339627_0


castwide:
Thanks!

PR #724, opened 2024-10-03
Include convention (eg. solargraph-rspec) pins in document_symbols

lekemula:
Hi there,

It looks like currently the additional pins added from conventions are not included in [documentSymbol request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol).

This PR adds them, thus enabling finding and easier navigation between `RSpec` `describe/context` blocks via [solargraph-rspec](https://github.com/lekemula/solargraph-rspec) plugin:

solargraph_rspec_document_symbols_1

solargraph_rspec_document_symbols_2

### DIFF
--- a/lib/solargraph/language_server/message/text_document/document_symbol.rb
+++ b/lib/solargraph/language_server/message/text_document/document_symbol.rb
@@ -6,6 +6,8 @@ class Solargraph::LanguageServer::Message::TextDocument::DocumentSymbol < Solarg
   def process
     pins = host.document_symbols params['textDocument']['uri']
     info = pins.map do |pin|
+      next nil unless pin.location&.filename
+
       result = {
         name: pin.name,
         containerName: pin.namespace,
@@ -17,7 +19,8 @@ class Solargraph::LanguageServer::Message::TextDocument::DocumentSymbol < Solarg
         deprecated: pin.deprecated?
       }
       result
-    end
+    end.compact
+
     set_result info
   end
 end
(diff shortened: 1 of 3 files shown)


lib/solargraph/source_map.rb
```diff
@@ -159,6 +161,17 @@ def map source
 
     private
 
+    # @return [Array<Pin::Base>]
+    def convention_pins
+      @convention_pins || []
+    end
+
+    def convention_pins=(pins)
+      # unmemoizing the document_symbols in case it was called from any of convnetions
```
lekemula, line 170: For example, solargraph-rails:

https://github.com/iftheshoefritz/solargraph-rails/blob/144cec0d91028b8073029df0bbaf58f6a2adbb3b/lib/solargraph-rails.rb#L39

This leads to subsequent convention pins not being included, hence we need to clear the memorization after all convention pins are collected.

lib/solargraph/language_server/message/text_document/document_symbol.rb
```diff
@@ -6,6 +6,8 @@ class Solargraph::LanguageServer::Message::TextDocument::DocumentSymbol < Solarg
   def process
     pins = host.document_symbols params['textDocument']['uri']
     info = pins.map do |pin|
+      next nil unless pin.location&.filename
```
lekemula, line 9: `solargraph-rspec` happens to be adding [some RSpec DSL methods](https://github.com/lekemula/solargraph-rspec/blob/169e30c7adae2e33248eec4fc3c4838af1654802/lib/solargraph/rspec/correctors/dsl_methods_corrector.rb#L54-L62) without a filename, which should be excluded.

castwide:
:100: :shipit:

PR #725, opened 2024-10-14
Fix "Not enough arguments to" false positives on kwarg type checking

apiology:
This valid code currently reports a false positive during typechecking:

```
        class Foo
          def bar(baz:, bing:)
          end
        end
        Foo.new.bar(baz: 123, bing: 456)
```


apiology:
Just let me know if you'd like any changes or additional tests, @castwide

apiology:
Happy to also send some (smaller) fixes for some other common issues I see with  codebases, @castwide?

castwide:
Thanks, @apiology. The failing tests are mostly unrelated issues that already exist in master and are being resolved in #739, so I'll go ahead and merge this one.

PR #731, opened 2024-12-08
Bump jaro_winkler dependency to remove deprecation

tagliala:
Require 'jaro_winkler', '~> 1.6', in order to use
`JaroWinkler.similarity` instead of `JaroWinkler.distance` to fix a deprecation message while running specs.


castwide:
:+1: :shipit:

Issue #738, opened 2025-01-12
Cannot find type definitions for library: stringio ([nil]) (RBS::EnvironmentLoader::UnknownLibraryError) when running `solargraph scan`


pangdaxing23:
branch: master@d9a0a89308a3337e2ae4d343a46b96e859b99245
ruby version: 3.4.1
stacktrace:
```
$ solargraph scan
/home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/rbs-3.8.1/lib/rbs/environment_loader.rb:138:in 'block in RBS::EnvironmentLoader#each_dir': Cannot find type definitions for library: stringio ([nil]) (RBS::EnvironmentLoader::UnknownLibraryError)
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/3.4.0/set.rb:504:in 'Hash#each_key'
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/3.4.0/set.rb:504:in 'Set#each'
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/rbs-3.8.1/lib/rbs/environment_loader.rb:136:in 'RBS::EnvironmentLoader#each_dir'
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/rbs-3.8.1/lib/rbs/environment_loader.rb:157:in 'RBS::EnvironmentLoader#each_signature'
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/rbs-3.8.1/lib/rbs/environment_loader.rb:121:in 'RBS::EnvironmentLoader#load'
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/rbs-3.8.1/lib/rbs/environment.rb:172:in 'block in RBS::Environment.from_loader'
        from <internal:kernel>:91:in 'Kernel#tap'
        from /home/MYUSER/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/rbs-3.8.1/lib/rbs/environment.rb:171:in 'RBS::Environment.from_loader'
[...]

PR #739, opened 2025-01-15
Fix Ruby 3.4

castwide:
This PR addresses several issues that arose in Ruby 3.4, including #738.

Smaller changes:
* `CoreMap` includes the stdlib (the source of the `stringio` error)
* `RbsMap` correctly handles the `UntypedFunction` type
* Miscellaneous changes to accommodate incompatible changes to the `RubyVM` AST (which kind of don't matter anymore; see below)

The big change is that the parser stops using `RubyVM::AbstractSyntaxTree` altogether.

Solargraph originally started using the `RubyVM` parser for performance reasons. The "legacy" whitequark parser remained as a fallback for earlier Ruby versions. Supporting both in perpetuity was likely to be untenable for a variety of reasons, and the instability of `RubyVM` (which is still marked experimental) makes whitequark the more practical choice.

For now, the `RubyVM` version of `Solargraph::Parser` is still in the codebase, but I expect to remove it and its related cruft in the near future.

### DIFF
--- a/lib/solargraph/parser.rb
+++ b/lib/solargraph/parser.rb
@@ -13,8 +13,8 @@ module Solargraph
     # True if the parser can use RubyVM.
     #
     def self.rubyvm?
-      !!defined?(RubyVM::AbstractSyntaxTree)
-      # false
+      # !!defined?(RubyVM::AbstractSyntaxTree)
+      false
     end
 
     selected = rubyvm? ? Rubyvm : Legacy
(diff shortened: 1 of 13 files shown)


lekemula:
> The main change here is that the parser no longer uses `RubyVM::AbstractSyntaxTree`.  

> For now, the RubyVM version of Solargraph::Parser is still in the codebase, but I expect to remove it and its related cruft in the near future.

Since Ruby 3.4 now uses [Prism as the default parser](https://www.ruby-lang.org/en/news/2024/10/07/ruby-3-4-0-preview2-released/), does it make sense to start transitioning toward Prism instead? Given that Prism provides [parser gem translation](https://github.com/ruby/prism/blob/main/docs/parser_translation.md), it might be a good time to align with it. RuboCop recently made a similar move: [rubocop/rubocop#12724](https://github.com/rubocop/rubocop/pull/12724).

PR #740, opened 2025-01-25
Fix exception from parser when anonymous block forwarding is used

apiology:
This code is valid as of Ruby 3.1, but solargraph raises an exception when parsing it:

```ruby
      def foo(&)
        bar(&)
      end
```


castwide:
Thanks!

PR #717, opened 2024-05-27
Implement Go To Type Definition

lekemula:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_typeDefinition

Demo in VScode (since in VIM everything is too fast 🤪)
solargraph_go_to_type_definition

All the necessary pieces were in place, and nothing fancy changed, mostly just exposing the functionality for LanguageServer. Almost feels like an easter egg for a new contributor like me 😄.

The only "functionality" added is `ComplexType#namespaces` which also utilizes existing implementation.

PR #742, opened 2025-01-29
Add #717 to changelog

lekemula:
Hi @castwide,

Thanks for your PR reviews and cutting the new release 🙏.

I noticed that #717 was missed in the release of the 0.51.0 changelog, so I thought maybe would be interesting for others to know about this feature added.


castwide:
:+1:

Issue #681, opened 2023-08-28
Question: Is it possible to define own generic/parameterized classes?


lekemula:
Hi,

YARDoc supports [parameterized types](https://rubydoc.info/gems/yard/file/docs/Tags.md#parametrized-types), although it doesn't really tell how to make use of them. Does solargraph support defining them somehow?

For example:

```ruby
class SuccessResult < Result
  def initialize(value)
    @value = value
  end

  # @return ???
  attr_reader :value
end

# @return [SuccessResult<Integer>]
def my_method
  SuccessResult.new(1)
end

my_method.value
#         ^^^^^ Integer
```

Looking at typical Ruby's array example they seem to be parsed as following: `Enumerator<param<Elem>>`
<img width="763" alt="image" src="https://github.com/castwide/solargraph/assets/9197495/f7fb6c90-d000-4e13-96b3-572ef7ffd5fa">

But I haven't found anything in the docs that shows how can one define custom generic classes to make use of such parameterized types.

I also tried defining the `.rbs` types declaration which supports such a feature, but that didn't help either. I'm assuming the RBS integration is still in progress (?).

Thanks in advance for your answers!

castwide:
Solargraph has limited support for parameterized types, but it mainly works with core classes like `Array` and `Set`. RBS is indeed still in progress. The code maps currently don't use it for the current workspace or third-party gems.

PR #743, opened 2025-02-13
Better support for generics via Class @param tags

lekemula:
Closes: https://github.com/castwide/solargraph/issues/681

This PR adds type inference for generics/parameterized params for `@return` and `@yieldparam` tags.

I went with `@param` instead of adding a custom tag for simplicity reasons. The tag is already parsed by YARDoc. Also, I think it's better to have fewer tags for developers to memorize. 

Does it make sense to you? Let me know if you have another opinion here.
 
<details><summary>Screenshots</summary>
<p>
	Screenshot 2025-02-13 at 21 07 36
	Screenshot 2025-02-13 at 22 40 58
	Screenshot 2025-02-13 at 22 43 22
</p>
</details> 

### Next steps

- [ ] Adjust documentation accordingly based on inferred types - right now raw yardoc `[param<...>]` is returned. 		<details><summary>Example</summary>
		Screenshot 2025-02-13 at 22 52 22
		</details> 

- [ ] Deprecate `@return_single_parameter` and `@yieldparam_single_parameter`(?)

### DIFF
--- a/lib/solargraph/pin/namespace.rb
+++ b/lib/solargraph/pin/namespace.rb
@@ -14,7 +14,7 @@ module Solargraph
       # @param type [::Symbol] :class or :module
       # @param visibility [::Symbol] :public or :private
       # @param gates [Array<String>]
-      def initialize type: :class, visibility: :public, gates: [''], parameters: [], **splat
+      def initialize type: :class, visibility: :public, gates: [''], parameters: nil, **splat
         # super(location, namespace, name, comments)
         super(**splat)
         @type = type
@@ -88,6 +88,11 @@ module Solargraph
         else
           [path] + @open_gates
         end
+      end
+
+      # @return [Array<String>]
+      def parameters
+        @parameters ||= docstring.tags(:param).map(&:name)
       end
     end
   end
--- a/lib/solargraph/pin/namespace.rb
+++ b/lib/solargraph/pin/namespace.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'yard-solargraph'
+
 module Solargraph
   module Pin
     class Namespace < Closure
@@ -92,7 +94,7 @@ module Solargraph
 
       # @return [Array<String>]
       def parameters
-        @parameters ||= docstring.tags(:param).map(&:name)
+        @parameters ||= docstring.tags(:generic).map(&:name)
       end
     end
   end
--- a/lib/solargraph/pin/namespace.rb
+++ b/lib/solargraph/pin/namespace.rb
@@ -11,12 +11,12 @@ module Solargraph
       # @return [::Symbol] :class or :module
       attr_reader :type
 
-      attr_reader :parameters
+      attr_reader :generics
 
       # @param type [::Symbol] :class or :module
       # @param visibility [::Symbol] :public or :private
       # @param gates [Array<String>]
-      def initialize type: :class, visibility: :public, gates: [''], parameters: nil, **splat
+      def initialize type: :class, visibility: :public, gates: [''], generics: nil, **splat
         # super(location, namespace, name, comments)
         super(**splat)
         @type = type
@@ -40,7 +40,7 @@ module Solargraph
           @closure = Pin::Namespace.new(name: closure_name, gates: [parts.join('::')])
           @context = nil
         end
-        @parameters = parameters
+        @generics = generics
       end
 
       def namespace
@@ -93,8 +93,8 @@ module Solargraph
       end
 
       # @return [Array<String>]
-      def parameters
-        @parameters ||= docstring.tags(:generic).map(&:name)
+      def generics
+        @generics ||= docstring.tags(:generic).map(&:name)
       end
     end
   end
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -199,6 +199,13 @@ module Solargraph
       result
     end
 
+    # @param namespace [String]
+    # @param context [String]
+    # @return [Array<Pin::Namespace>]
+    def get_namespace_pins namespace, context
+      store.fqns_pins(qualify(namespace, context))
+    end
+
     # Get a fully qualified namespace name. This method will start the search
     # in the specified context until it finds a match for the name.
     #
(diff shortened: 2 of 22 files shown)


lekemula:
@castwide, in regards of:

>  Deprecate @return_single_parameter and @yieldparam_single_parameter(?)

Here you can find the draft PRs:

- `@yieldparam_single_parameter` - https://github.com/lekemula/solargraph/pull/1
	- It looks like the new implementation seems to play nicely with RBS maps already 🔥 
- `@return_value_parameter` & `@return_single_parameter` https://github.com/lekemula/solargraph/pull/2
	- These tags seem to be not in use already. Does it make sense to remove them, or do you still see any of them being used in the future?

castwide:
Thanks, @lekemula! I've been thinking about better ways to handle parametrized type inference, and your solution is very close to how I pictured it.

Good to see that it also works with multiple parameters, e.g., when a Hash-like object needs key and value parameters:

```ruby
# @param Key
# @param Value
class Map
  # @yieldparam [param<Key>]
  # @yieldparam [param<Value>]
  def each(&block); end
end

# @return [Map<String, Integer>]
def map_method
  Map.new
end

map_method.each do |key, value|
  key._ # Completion for String
  value._ # Completion for Integer
end
```

YARD doesn't complain about `@param` tags attached to class definitions, but it's a use case that's not covered in the [specification](https://github.com/lsegal/yard/blob/main/docs/Tags.md). We might want to get some input from @lsegal to make sure we're not introducing undesirable behavior in things like generated docs.

I can think of a couple ways to extend the syntax around this feature, but we can probably leave those to future enhancements.

As for the `@return_single_parameter` and `@yieldparam_single_parameter`, both are definitely candidates for deprecation. They started as a quick and dirty way to infer yieldparams for Enumerable methods back before RBS existed.

castwide:
I [opened a YARD issue](https://github.com/lsegal/yard/issues/1606) to ask about the param syntax and he made a strong argument for using a custom tag instead of repurposing the `@param` tag. His suggestion of `@generic` seems reasonable. There's no problem with using `param<T>` in types.

My example, modified:
```ruby
# @generic [Key, Value]
class Map
  # @yieldparam [param<Key>]
  # @yieldparam [param<Value>]
  def each(&block); end
end

# @return [Map<String, Integer>]
def map_method
  Map.new
end

map_method.each do |key, value|
  key._ # Completion for String
  value._ # Completion for Integer
end
```

What do you think, @lekemula?

castwide:
It also occurred to me that we can bypass the need for a class tag if we can reference the type parameters by index:

```ruby
class Map
  # @yieldparam [param<0>]
  # @yieldparam [param<1>]
  def each(&block); end
end
```

castwide:
Good points about named vs. indexed parameters. Let's go with the `@generic` tag. I might revisit indexed parameters later, but right now it's just a maybe-nice-to-have.

I've been planning to publish a separate `yard-solargraph` extension. Other gems would be able to include it in their YARD configurations so they'd be included in generated yardocs. Currently I think the only two tags we would need to include are `@generic` and `@yieldreceiver` (see https://github.com/lsegal/yard/pull/1491). Barring native YARD support, an extension should work well enough.

For now, we can add `@generic` to `lib\yard-solargraph.rb`, and I'll split it into a separate extension afterwards. (I already have projects where I'm eager to have `@yieldreceiver` work correctly, so I plan to do it ASAP.)

lsegal:
> My main concern, however, is ensuring that @generic tags defined in third-party gems get indexed properly. If we introduce this new custom tag in solargraph, wouldn’t that mean the yard gems command wouldn’t parse it? If so, is there an easy way to work around this?

I'm not _exactly_ sure how solargraph works, but if you're in charge of running `yard gems`, `YARD.parse` or equivalent, you can define your own YARD tags. In the case of generic it would be simply making sure you had the following somewhere in your runtime:

```ruby
YARD::Tags::Library.define_tag("Generic", :generic, :with_types)
```

In the case of running with `yard gems`, it might require you to run the command in a slightly different way (maybe `yard gems -e /path/to/the/file/with/the/above/ruby/code.rb` or use `--plugin`).

This would allow you to parse these tags regardless of if library has done so (though they probably would want to in order to generate their own docs).

lib/solargraph/pin/namespace.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'yard-solargraph'
```
lekemula, line 3: This was needed for the unit test to pass.

lib/solargraph/pin/namespace.rb
```diff
@@ -89,6 +91,11 @@ def gates
           [path] + @open_gates
         end
       end
+
+      # @return [Array<String>]
+      def parameters
```
lekemula: @castwide would it make sense to rename this to something like:
```suggestion
      def generics
```

or

```suggestion
      def generic_parameters
```

The current `def parameters` interface seems to be incompatible with other pins anyway with the return types:

https://github.com/castwide/solargraph/blob/99801d1aaea62c28a6ac6c558c812c50c1e81fe5/lib/solargraph/pin/block.rb#L29-L32

lekemula:
Thanks, @lsegal, for your helpful information! 🙏 

> In the case of running with yard gems, it might require you to run the command in a slightly different way (maybe yard gems -e /path/to/the/file/with/the/above/ruby/code.rb or use --plugin).

As far as I know, Solargraph recommends running `yard gems` [directly](https://solargraph.org/guides/yard). However, we should be able to create a wrapper or shim around yard gems based on the suggestion above. I'm unsure, though, how we could maintain the convenience of `yard config --gem-install-yri`.

@castwide will you be able to look into this, or should I give it a try on a separate PR?

castwide:
Thanks, @lekemula! This is awesome.

Good call on renaming `Namespace#parameters`. I think `#generics` fits well.

Along the same lines, do you think we should use `generic` instead of `param` to clarify the relationship between the two?

```ruby
# @generic T
class Example
  # @yieldparam [generic<T>]
  def evaluate; end
end
```

Regarding `yard gems`: I think the first step is to release the yard-solargraph extension so gem authors can opt into the additional tags. That should help us determine what else, if anything, we need to do to support them.

castwide:
:100:

I've also started work on the [yard-solargraph](https://github.com/castwide/yard-solargraph) plugin.

lekemula:
> I've also started work on the [yard-solargraph](https://github.com/castwide/yard-solargraph) plugin.

Appreciate that @castwide and thanks for the great feedback. That will be really helpful in allowing third-party gems to provide a better developer experience for their users—hoping it gets adopted by those already using YARD comments!

PR #744, opened 2025-02-17
Handle symbol tags in method tag values

apiology:
Ensure that symbol-typed default values for method arguments in @!method tags don't have their method pin get disregarded because of the extra symbol pin created.

### DIFF
--- a/lib/solargraph/source_map/mapper.rb
+++ b/lib/solargraph/source_map/mapper.rb
@@ -107,7 +107,8 @@ module Solargraph
           begin
             src = Solargraph::Source.load_string("def #{directive.tag.name};end", @source.filename)
             region = Parser::Region.new(source: src, closure: namespace)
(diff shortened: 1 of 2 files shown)


apiology:
I hit this case in my own codebase - let me know if this patch looks right!

lib/solargraph/source_map/mapper.rb
```diff
@@ -107,7 +107,8 @@ def process_directive source_position, comment_position, directive
           begin
             src = Solargraph::Source.load_string("def #{directive.tag.name};end", @source.filename)
             region = Parser::Region.new(source: src, closure: namespace)
-            gen_pin = Parser.process_node(src.node, region).first.last
```
apiology, line 110: In the case where a symbol was the default value of method parameter, this was returning the symbol pin, not the method pin

castwide:
Looks good to me. Thanks!

PR #747, opened 2025-02-19
Handle interpolated symbol literals

apiology:


castwide:
Thanks! This fixes a long-time issue with literal symbols and I especially appreciate the tests.

PR #748, opened 2025-02-20
Don't require redundant attribute @return and @param tags

apiology:
If the underlying instance variable is already typed (either via @type or via creation from a typed initializer param), don't require attributes (e.g., attr_accessor, attr_reader, etc.) to have redundant @return and @param tags, even under strong typechecking.

### DIFF
--- a/lib/solargraph/type_checker.rb
+++ b/lib/solargraph/type_checker.rb
@@ -91,7 +91,12 @@ module Solargraph
       declared = pin.typify(api_map).self_to(pin.full_context.namespace)
       if declared.undefined?
         if pin.return_type.undefined? && rules.require_type_tags?
-          result.push Problem.new(pin.location, "Missing @return tag for #{pin.path}", pin: pin)
+          if pin.attribute?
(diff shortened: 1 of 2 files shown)


apiology:
Yeah, I love the idea.  

I've been playing around with using @AaronC81's https://github.com/AaronC81/sord and https://github.com/AaronC81/parlour to use my yard type annotations to feed into Sorbet's engine and to generate .rbi and .rbs files to package with my published gems: https://github.com/apiology/cookiecutter-gem/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/Makefile#L31-L41

At a higher level - seems like we want to let annotations interoperate between and inside gems and local source files so we're not throwing away information.

Once we have that, I think we should think about decoupling inference and type checking engines from annotation processors so we can focus energy on having really high quality implementations of all three.

...one step at a time :)

apiology:
Along these lines, @Morriar just dropped support for RBS comments in Sorbet - https://github.com/sorbet/sorbet/pull/8470

I haven't played around with it yet, but it's in line with these thoughts (excuse the pun): https://sorbet.org/docs/rbs-support

PR #749, opened 2025-02-20
Use @yieldreturn tags for type inference

apiology:
* Track yield calls similarly to super() calls in source chains
* Add a block's Signature pin in method Signature when block parameter declared or documented
* Include info from yieldparam/yieldreturn tags in that block Signature pin when they exist
* Teach Signature pin to be included in inference chains
* Fix a docstring-related test bug in method_spec.rb
* Include yieldparam/yieldreturn info in method documentation from LSP
* Add misc inference specs to lock down existing related behavior
* Add ability to method return types based on yield call's @yieldreturn


castwide:
:+1: :shipit:

PR #750, opened 2025-02-22
Fix type annotations identified at 'typed' level

apiology:
This doesn't get the codebase to be --level=typed clean yet, but does get rid of the easy things.

### DIFF
--- a/lib/solargraph.rb
+++ b/lib/solargraph.rb
@@ -56,8 +56,6 @@ module Solargraph
 
   # A helper method that runs Bundler.with_unbundled_env or falls back to
   # Bundler.with_clean_env for earlier versions of Bundler.
-  #
-  # @return [void]
   def self.with_clean_env &block
     meth = if Bundler.respond_to?(:with_original_env)
       :with_original_env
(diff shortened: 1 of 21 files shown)


lib/solargraph.rb
```diff
@@ -56,8 +56,6 @@ def self.logger
 
   # A helper method that runs Bundler.with_unbundled_env or falls back to
   # Bundler.with_clean_env for earlier versions of Bundler.
-  #
-  # @return [void]
```
apiology, line 60: We use this in a place where we save the return value.  The return type and the yieldreturn here really should be `param<T>` once we support that.

lib/solargraph/source/chain/call.rb
```diff
@@ -197,7 +197,7 @@ def arguments_match arguments, signature
 
         # @param api_map [ApiMap]
         # @param name_pin [Pin::Base]
-        # @return [Array<Pin::Base>]
+        # @return [::Array<Pin::Base>]
```
apiology, line 200: This clarifies between Solargraph::Chain::Array and the stdlib Array class.  I have another fix in the works to fix a bug where we confuse the two during typechecking and report a false positive, but this at least makes the declaration correct.

castwide:
:+1: :shipit:

PR #751, opened 2025-02-22
Support RBS class aliases

apiology:
This was preventing some core/stdlib constants from resolving correctly. e.g., "Mutex" is an alias of the class "Thread::Mutex"

### DIFF
--- a/lib/solargraph/rbs_map/conversions.rb
+++ b/lib/solargraph/rbs_map/conversions.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'rbs'
+
 module Solargraph
   class RbsMap
     # Functions for converting RBS declarations to Solargraph pins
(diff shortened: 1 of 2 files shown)


apiology:
This was the next one in line to get a clean typecheck of the sourcegraph codebase!  Looks like there's some low hanging fruit in adding more RBS AST types - I may take a crack at that.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'rbs'
```
apiology, line 3: This was so I could get RBS class information into the Solargraph LSP as I made these changes - I can throw this in a `@!parse` block or something if leaving it out was intentional.

castwide:
LGTM, thanks!

PR #752, opened 2025-02-22
Avoid discarding 'rooted?' information in Pin::Method#return_type

apiology:
Solargraph gets confused when a return type's name exists at multiple levels - e.g., "@return [::Array]" in a Pin class method is understood as Solargraph::Pin::Array instead of Ruby's provided Array class.

Avoid that issue by avoiding unnecessary round-trip serialization of UniqueType objects via #to_s, which is ambiguous as it doesn't include the '::' prefix.


castwide:
Thanks!

PR #753, opened 2025-02-22
Fix a typecheck-found issue in the RubyVM parser

apiology:


castwide:
The RubyVM version of the parser is deprecated. I expect to yank it altogether, but in the meantime there's no harm in merging this if it suppresses a typecheck warning.

PR #754, opened 2025-02-22
Fix inference of begin expression types

apiology:
Currently the first line of a begin...end expression is used as the type of the expression; instead, the last line should be used.


apiology:
This one was causing a number of false alarm errors while typechecking the Solargraph codebase (and mine, as well!)

I'm still a little shaky at how the chaining code works, so if the code I changed was intentional, just give me an example of code that needs that behavior to typecheck and I'll take anotehr crack at it :)

castwide:
Looks right to me. :shipit:

PR #756, opened 2025-02-23
Fix RBS ingestion implicit initializer issues, missing param types

apiology:
Also fix up a couple of issues with tags in Solargraph code itself

### DIFF
--- a/lib/solargraph/parser/rubyvm/class_methods.rb
+++ b/lib/solargraph/parser/rubyvm/class_methods.rb
@@ -6,9 +6,8 @@ module Solargraph
     module Rubyvm
       module ClassMethods
         # @param code [String]
-        # @param filename [String]
+        # @param filename [String, nil]
         # @return [Array(Parser::AST::Node, Array<Parser::Source::Comment>)]
-        # @sg-ignore
         def parse_with_comments code, filename = nil
           node = RubyVM::AbstractSyntaxTree.parse(code).children[2]
           node &&= RubyVM::AbstractSyntaxTree::NodeWrapper.from(node, code.lines)
@@ -22,7 +21,6 @@ module Solargraph
         # @param filename [String, nil]
         # @param line [Integer]
         # @return [Parser::AST::Node]
-        # @sg-ignore
         def parse code, filename = nil, line = 0
           node = RubyVM::AbstractSyntaxTree.parse(code).children[2]
           node and RubyVM::AbstractSyntaxTree::NodeWrapper.from(node, code.lines)
--- a/lib/solargraph/rbs_map/conversions.rb
+++ b/lib/solargraph/rbs_map/conversions.rb
@@ -29,11 +29,37 @@ module Solargraph
         @type_aliases ||= {}
       end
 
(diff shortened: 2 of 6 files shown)


lib/solargraph/rbs_map/conversions.rb
```diff
@@ -29,11 +29,37 @@ def type_aliases
...
 
+      # @param loader [RBS::EnvironmentLoader]
+      # @return [void]
+      def load_environment_to_pins(loader)
+        environment = RBS::Environment.from_loader(loader).resolve_type_names
+        cursor = pins.length
+        environment.declarations.each { |decl| convert_decl_to_pin(decl, Solargraph::Pin::ROOT_PIN) }
+        added_pins = pins[cursor..-1]
+        add_back_implicit_pins(added_pins)
+      end
```
apiology, line 40: The definition of a class may be split among different declaration files.  As a result, we were prematurely creating implicit initializers in cases when we didn't need them, resulting in multiple initializers defined as well as one of them likely having the wrong arity.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -29,11 +29,37 @@ def type_aliases
...
+      # @param added_pins [Range<Pin>]
+      # @return [void]
+      def add_back_implicit_pins(added_pins)
+        added_pins.each do |pin|
+          pin.source = :rbs
+          next unless pin.is_a?(Pin::Namespace) && pin.type == :class
+          next if pins.any? { |p| p.path == "#{pin.path}.new"}
+          pins.push Solargraph::Pin::Method.new(
+                      location: nil,
+                      closure: pin,
```
apiology, line 51: This line changed from closure.pin to pin.  We are being passed the pin for the class; we don't want to create the initializer in the class' surrounding closure, but in the class we were passed directly, as a class method.

apiology: I would guess at some point we'd want to teach Solargraph to chase the parent classes to look for the first one with an initializer and use its arity and types, replacing its return type, rather than assuming the only one defined was Object.new.  

Let me know if I'm misunderstanding anything.

lib/solargraph/parser/rubyvm/class_methods.rb
```diff
@@ -22,7 +21,6 @@ def parse_with_comments code, filename = nil
         # @param filename [String, nil]
         # @param line [Integer]
         # @return [Parser::AST::Node]
-        # @sg-ignore
```
apiology, line 25: One of these days I'd like to look into solargraph warning when an @sg-ignore is not needed, so the user makes sure to remove them once the issue is addressed.

lib/solargraph/rbs_map/conversions.rb
castwide: That would definitely be more helpful. The implementation might be tricky. I'm guessing we would either have to update the initializers after all the pins have been generated or figure out a way to resolve them lazily.

castwide:
Thanks!

PR #758, opened 2025-02-26
Annotation fixes towards a clean Solargraph typecheck of itself

apiology:


apiology:
With the other PRs, this gets us to only two problems left at the 'typed' level!

I'll PR in a CI check once we're at zero on the main branch to keep it there.

apiology:
I'm not able to reproduce this faillure - any idea what's going on, @castwide?

castwide:
I just reran the tests for now. :shrug:

PR #761, opened 2025-02-27
Fix an internal warning while typechecking

apiology:
The current typechecking code logs:

```
WARNING: Pin #infer methods are deprecated. Use #typify or #probe
instead.
```

Move the infer() call to typify(), as we know are interested in validating the declared type exists, not the inferred one.


castwide:
:+1: :shipit:

PR #764, opened 2025-02-28
Address remaining typecheck issues at 'typed' level and add CI task

apiology:


castwide:
Woooow. This is super interesting. I feel like I need to look closer before merging and releasing, but I'm super happy with the direction it's going.

castwide:
:100: :shipit:

PR #765, opened 2025-03-01
Add additional type annotations to Solargraph code

apiology:


apiology:
Another chunk of work towards typechecking at strict level!

lib/solargraph/api_map.rb
```diff
@@ -659,8 +666,8 @@ def get_namespace_type fqns
 
     # Sort an array of pins to put nil or undefined variables last.
     #
-    # @param pins [Array<Solargraph::Pin::Base>]
-    # @return [Array<Solargraph::Pin::Base>]
+    # @param pins [Enumerable<Solargraph::Pin::Base>]
+    # @return [Enumerable<Solargraph::Pin::Base>]
```
castwide, line 670: Some of the changes from `Array` to `Enumerable` look erroneous. One concern is that Solargraph does not recognize `Array` methods, e.g., `Array#each`, as valid calls from an `Enumerable`. In the particular case of `ApiMap#prefer_non_nil_variables`, keeping the `@return` type as `[Array<Solargraph::Pin::Base>]` does not produce an error in strict typechecking. If there are cases where that happens, we might need to explore whether it's a problem in static code analysis rather than the documented type.

castwide: I just saw #769, which addresses this exact problem, and it looks like the right solution. Lemme verify a couple other things, but right now I'm inclined to merge :+1:

apiology: haha, exactly right.  I thought it was an elegant solution to the mixin problem by the RBS (or perhaps Sorbet) folks.

Generally, I'm sure I got some of these wrong.  I figure of the ones which are wrong, 80% will be resolved as we get rid of the current strict/strong typechecking violations, then the rest are probably going to be a good candidates for typechecking bug fixes / improvements as we run across them.

If you do happen to spot something blatantly off, let me know of course - this exercise was really helpful in getting a better understanding of the codebase, so any non-trivial corrections will be even more educational!

castwide: Yeah, I agree with your approach. As we continue to improve RBS support and static inference, I suspect we'll see cases where it makes sense to change the type back to `Array`, and other cases where it makes sense to improve Solargraph's understanding of `Enumerable`. In the meantime, `Enumerable` is still technically correct.

castwide:
Thanks! :shipit:

PR #767, opened 2025-03-01
DeepInference: Fix some bugs, add docs, refactor

apiology:
Address some unhandled cases around rescue blocks after documenting and refactoring DeepInference based on my best understanding of it.

This fixes some false positives given by Solargraph while typechecking its own codebase at level 'strict'.


apiology:
@castwide: I've got a couple of more cases in the backlog to handle, and I think there's a larger refactor to be done, but I'd like to get your take on the terminology I'm using here first.

Happy to standardize on different terms or draw more/less fine-grained distinction between them, but having a common vocabulary is a good idea.  In the meantime, having some definitions in front of me helped build a mental model and take the refactor far enough to get the begin/rescue case working.

castwide:
Thanks! Your definitions in the `DeepInference` comments look solid. :shipit:

PR #768, opened 2025-03-06
Populate location information from RBS files

apiology:
The 'rbs' gem maps the location of different definitions to the relevant point in the RGS files themselves - this change provides the ability to jump into the right place in those files to see the type definition via the LSP.


castwide:
I've been on the fence about assigning a pin's `location` from RBS. When I use go-to-definition, I've never wanted to see the RBS definition instead of the source code. Maybe we should create a new pin attribute (e.g., `rbs_location`), and either 1) use the `rbs_location` when `location` doesn't exist, or 2) figure out a reasonable way to differentiate between requests for RBS definitions and source definitions.

apiology:
All right, this now prefers code location to RBS location when both are available

apiology:
@castwide: Merge conflicts resolved - should be good to go when you're ready for it

castwide:
Looks good! I definitely appreciate tracking both code and type locations. We might consider some changes to how they're accessed through the LSP down the road, but this is a great place to start.

PR #770, opened 2025-03-07
Process RBS globals, module aliases, class vars and class instance vars

apiology:
Translate the following from RBS:

* globals: for stdlib/core, this gets us things like '$stdout'
* module alias: this gets us YAML as an alias for Module<Psych> without a hard-coded workaround
* class variable: mainly for completeness - only a few defined within the RBS code types themselves
* class instance variable: ditto


castwide:
Thanks! `ClassVariable` and `ClassInstanceVariable` are the last of the missing type warnings I get while building the core caches :100:

PR #772, opened 2025-03-09
Add some regression test specs

apiology:


apiology:
I pulled these out of a development branch I'm working on - most of them are things I managed to break that weren't caught by the existing suite 😅

PR #775, opened 2025-03-11
Handle parsing tuples of tuples in tags

apiology:


lib/solargraph/complex_type.rb
```diff
@@ -241,7 +241,7 @@ def parse *strings, partial: false
               paren_stack += 1
             elsif char == ')'
               paren_stack -= 1
-              subtype_string += char if paren_stack == 0
```
apiology, line 244: There wasn't much in the commit log - if you remember why this was there, happy to add more specs and ensure they pass as well.

spec/complex_type_spec.rb
```diff
@@ -310,4 +310,16 @@ def make_bar
...
+  it 'parses tuples of tuples' do
+    api_map = Solargraph::ApiMap.new
+    type = Solargraph::ComplexType.parse('Array(Array(String), String)')
+    expect(type.tag).to eq('Array(Array(String), String)')
+  end
+
+  it 'parses tuples of tuples with same type twice in a row' do
+    api_map = Solargraph::ApiMap.new
+    type = Solargraph::ComplexType.parse('Array(Symbol, String, Array(Integer, Integer))')
+    expect(type.to_s).to eq('Array(Symbol, String, Array(Integer, Integer))')
```
apiology, line 323: The (Integer, Integer) here does collapse into (Integer) at some point between here and appearing in my LSP results.  I'll chase this down next and either add this to the same PR or push a new one if you get to this first.

apiology: Ah, the qualify() method in UniqueType was applying .uniq, assuming that every list of types represented  a union type.  Fixed.

apiology:
These bugfixes knock out a batch of strict typechecking issues.

Looks like we'll still need:

- something to resolve the #[] and #[]= calls with to the right type when we have a literal integer argument
- multiple assignment and destructuring support
- support for treating literals like tuples when they have heterogeneous types (?)

I also suspect translating `Array()` to `Tuple() `is when we parse it is probably a good idea.  It'll help us stay aligned with RBS standard type names and also pull in anything useful we can from those definitions.  

Let me know if you have thoughts on any of the above!  I think this change can be independent of any of the above, though; it doesn't make anything worse to my knowledge and at least helps humans understand the types without garbling them for the moment.

lib/solargraph/complex_type.rb
castwide: If I remember correctly, that condition was a quick fix for some wonky issue with the first iteration of type parameter support. It was seven years ago, so the specifics are lost to me. 🤷 Given the many examples of parametrized types in the specs, I'm plenty confident this won't introduce a regression.

castwide:
> These bugfixes knock out a batch of strict typechecking issues.
> 
> Looks like we'll still need:
> 
>     * something to resolve the #[] and #[]= calls with to the right type when we have a literal integer argument
> 
>     * multiple assignment and destructuring support
> 
>     * support for treating literals like tuples when they have heterogeneous types (?)
> 
> 
> I also suspect translating `Array()` to `Tuple() `is when we parse it is probably a good idea. It'll help us stay aligned with RBS standard type names and also pull in anything useful we can from those definitions.
> 
> Let me know if you have thoughts on any of the above! I think this change can be independent of any of the above, though; it doesn't make anything worse to my knowledge and at least helps humans understand the types without garbling them for the moment.

Sounds good to me. Generics and enhanced parameter support have opened a lot of possibilities. Thanks! :100: :shipit:

PR #776, opened 2025-03-11
Retire the RubyVM parser

apiology:


apiology:
@castwide: Let me know if this is what you had in mind!  I tried to leave room for a potential future transition to the parser flavor of the month.

castwide:
Looks good! I might look into using `prism` if it proves to be significantly faster, but one of its benefits is that it can output `parser`-compatible ASTs.

PR #779, opened 2025-03-12
Fix internal mix-in type-checking issues

apiology:


lib/solargraph/complex_type.rb
```diff
@@ -219,6 +219,9 @@ def parse *strings, partial: false
                 raise ComplexTypeError, "Invalid hash thing" unless key_types.nil?
                 # types.push ComplexType.new([UniqueType.new(base[0..-2].strip)])
                 types.push UniqueType.new(base[0..-2].strip)
+                # @todo this should either expand key_type's type
+                #   automatically or complain about not being
+                #   compatible with key_type's type in type checking
```
apiology, line 224: (not related)

lib/solargraph/language_server/transport/adapter.rb
```diff
@@ -8,6 +8,20 @@ module Transport
...
+        #
+        # @!parse
+        #   # @abstract
+        #   # Write the change to the specified text.
+        #   #
+        #   # @param text [String] The text to be changed.
+        #   # @return [String] The updated text.
+        #   def write text
+        #     raise NotImplementedError
+        #   end
```
castwide: You could also do this with a `@!method` directive:

```suggestion
        # @!method write(text)
        #   @abstract
        #   Write the change to the specified text.
        #   @param text [String] The text to be changed.
        #   @return [String] The updated text.
```

In this case they appear to be functionally equivalent, but sometimes `@!parse` can do unexpected things.

apiology: Thanks!  Yeah, for sure I need to check my tag crib notes more instead of leaning on @!parse so much :)

Issue #777, opened 2025-03-12
Incorrect understanding of `max_by` return type


amake:
The following code:

```ruby
# @return [Integer]
def foo
  [1, 2, 3].max_by(&:abs)
end
```

Fails to type check with the following error:

```
% solargraph typecheck --level typed tmp.rb
tmp.rb:2 - Declared return type Integer does not match inferred type Enumerator<Integer, Integer> for #foo
1 problem found.
```

Looking at the RBS it seems like it does indicate that a single object can be returned:

https://github.com/ruby/rbs/blob/97ed5d9492b9d2050b87dd4b247be4097e970b9d/core/enumerable.rbs#L973-L976

apiology:
Yeah, I've seen similar with these methods where the return type varies with the arity of arguments / block provided.  

This is likely the place to debug: https://github.com/castwide/solargraph/blob/master/lib/solargraph/source/chain/call.rb#L60-L83

PR #780, opened 2025-03-12
Improve block handling in signature selection

apiology:
Fixes https://github.com/castwide/solargraph/issues/777


apiology:
I have some work that builds on this to resolve method generics here: https://github.com/apiology/solargraph/pull/1

lib/solargraph/source/chain/call.rb
```diff
@@ -57,17 +57,30 @@ def inferred_pins pins, api_map, context, locals
...
+                atype = nil
                 if param.nil?
-                  match = false unless ol.parameters.any?(&:restarg?)
+                  last_arg = idx == arguments.length - 1
+                  match = if ol.parameters.any?(&:restarg?)
+                            true
+                          elsif last_arg && ol.block?
+                            # block argument that isn't declared as an arg as well
+                            atype ||= arg.infer(api_map, Pin::ProxyType.anonymous(context), locals)
+                            atype.namespace == 'Proc'
```
castwide, line 77: Is `atype.namespace == 'Proc'` a noop?

apiology: No - confusing logic, but it creates a boolean that flows up and gets assigned to the `match` variable

castwide: \*smacks head\* Now I see it. Thanks.

castwide:
:100: :shipit:

PR #781, opened 2025-03-13
Cache YARD and RBS

castwide:
* Generate cached yardocs with the yard-solargraph plugin
* Enhance cached gem pins with RBS

This update will involve some major changes to `YardMap`, `RbsMap`, and `ApiMap`.

Objectives:
- [x] `Yardoc` generates cached yardocs with the yard-solargraph plugin
- [x] `YardMap` loads yardocs from the Solargraph cache
- [x] Map gem pins from RBS
- [x] YARD mapping no longer requires yardocs to exist in gem directories
- [x] `ApiMap` is capable of tracking unresolved gems
- [x] `LanguageServer::Host` builds uncached gem documentation in the background
- [x] `LanguageServer::Host` updates libraries after new gems are cached
- [x] Command-line tasks are capable of updating cached gems in place
- [x] Enhance cached pins with RBS
- [x] Pins enhanced with RBS use locations from YARD where available
- [x] Add a `solargraph gems` command to cache pins

Next steps:
- Support gem_rbs_collection
- Generate workspace sig folders


castwide:
* The language server can now build and cache gem documentation in the background. It is no longer necessary to run `yard gems` manually or use `yard config` to generate yardocs on installation. The language server is still usable while the gems are being processed. Each gem will be added to the map on the fly after it gets cached.
* YARD mapping includes the yard-solargraph plugin to enable custom tags like `@generic` and `@yieldreceiver`.
* `RbsMap` loads from gems in addition to core and stdlib. When a gem provides RBS definitions, it gets combined with YARD in the cached documentation. RBS takes precedence for designation of parameters, generics, and return types. Source code locations are still used for LSP features like go-to-definition.

apiology:
I'll start running with this today and let you know if I see any issues - first with existing gems directory and then with a fresh one

.github/workflows/typecheck.yml
```diff
@@ -30,7 +30,5 @@ jobs:
         bundler-cache: false
     - name: Install gems
       run: bundle install
-    - name: Set up yardocs
-      run: bundle exec yard gems
```
apiology, line 34: References to types used in annotations but without a 'require' statement now result in typechecking errors at level 'typed':

```
/Users/broz/src/solargraph/lib/solargraph/converters/dd.rb:7 - Unresolved type Nokogiri::XML::Element for node param on ReverseMarkdown::Converters::Dd#convert
/Users/broz/src/solargraph/lib/solargraph/converters/dl.rb:7 - Unresolved type Nokogiri::XML::Element for node param on ReverseMarkdown::Converters::Dl#convert
/Users/broz/src/solargraph/lib/solargraph/converters/dt.rb:7 - Unresolved type Nokogiri::XML::Element for node param on ReverseMarkdown::Converters::Dt#convert
```

You can reproduce this by merging in https://github.com/castwide/solargraph/pull/787 and running `bundle exec solargraph typecheck --level typed`

Adding a require 'nokogiri' in one of the 3 files resolved it for all of them.

I think this is probably OK behavior and I think users will understand they need to add require statements if they reference types - just wanted to note the change.

lib/solargraph/shell.rb
```diff
@@ -232,13 +232,48 @@ def bundle
...
+
+            puts "Processing gem: #{spec.name} #{spec.version}"
+            pins = GemPins.build(spec)
+            Cache.save("#{spec.name}-#{spec.version}.ser", pins)
+          else
+            warn "Gem '#{name}' not found"
+          end
+        end
+      end
+    end
```
apiology, line 267: Appreciate you adding this one - will help simplify some CI caching logic I have in projects that use solargraph

lib/solargraph/api_map.rb
```diff
@@ -141,6 +134,23 @@ def self.load directory
...
+    # any missing gems.
+    #
+    # @param directory [String]
+    # @return [ApiMap]
+    def self.load_with_cache directory
+      api_map = load(directory)
+      return api_map if api_map.uncached_gemspecs.empty?
+
+      api_map.uncached_gemspecs.each do |gemspec|
+        puts "Caching #{gemspec.name} #{gemspec.version}..."
```
apiology:
```suggestion
        $stderr.puts "Caching #{gemspec.name} #{gemspec.version}..."
```

Appreciate this status message!  Maybe we can add one to the RBS ingestion path as well so the user knows where their type info is coming from?

I'm inclined to keep the stdout info as-is for better tool ingestability and leave this sort of status message to stderr - but let me know what you think.  

For instance, I use solargraph with overcommit now (thankfully it looks like it wouldn't break it): https://github.com/apiology/cookiecutter-ruby/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/.git-hooks/pre_commit/solargraph_typecheck.rb

lib/solargraph/shell.rb
```diff
@@ -232,13 +232,48 @@ def bundle
...
+        end
+      else
+        names.each do |name|
+          spec = Gem::Specification.find_by_name(name)
+          if spec
+            next unless options.rebuild || !Yardoc.cached?(spec)
+
+            puts "Processing gem: #{spec.name} #{spec.version}"
+            pins = GemPins.build(spec)
+            Cache.save("#{spec.name}-#{spec.version}.ser", pins)
```
apiology: Is this intended to be a different path than the save() line above?

lib/solargraph/cache.rb
```diff
@@ -23,10 +24,21 @@ def work_dir
...
+      # @return [String]
+      def join *path
+        File.join(work_dir, *path)
+      end
+
       # @param path [Array<String>]
       # @return [Array<Solargraph::Pin::Base>, nil]
       def load *path
-        file = File.join(work_dir, *path)
+        file = join(*path)
```
apiology, line 41: Does save() need the same change?

.github/workflows/typecheck.yml
castwide: Yeah, that should be okay.

The removal of `yard gems` from the workflow is a little tangential to this. The `Yardoc` module builds yardocs with the yard-solargraph plugin and saves them in the cache. This seemed like the best way to assure that the yardocs we use always include `@generic` and `@yieldreceiver` tags. I thought `yard gems` might pick up the plugin if it was in the gem's .yardopts, but it doesn't (which makes sense in retrospect).

If we want to cache all the gem yardocs at once, we can use `solargraph gems` instead.

lib/solargraph/shell.rb
castwide: Good catch. I fixed it and pushed a commit.

I think this indicates an issue I should resolve with caching in general. There are now four different types of cached files:

* Three closely related but subtly different caches for core, stdlib, and gem pins
* Yardocs built with yard-solargraph

I'll try to codify the standard for each cache to avoid confusion.

lib/solargraph/api_map.rb
castwide: Yeah, that was an oversight on my part. I changed it to `Solargraph.logger.info`.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -229,14 +228,6 @@ def create_constant(name, tag, comments, type)
...
           comments: comments
         )
-        tag = if type == :class
-                "Class<#{tag}>"
-              elsif type == :module
-                "Module<#{tag}>"
-              else
-                Solargraph.logger.warn "Unrecognized constant type: #{type.inspect}"
-                "Class<#{tag}>"
-              end
```
castwide, line 239: @apiology This tag adjustment appears to be redundant. The `generic_type` method already handles parameterized `Class` and `Module` types. I verified with a test in `rbs_map_spec.rb`, but figured I should bring it to your attention in case there's something I missed.

castwide: I might have misconstrued the problem here because I tested off a stale cache. There's still a discrepancy somewhere, but I'll need to investigate further.

castwide: I think I got it. Aliases need the conversion to `Class<T>` or `Module<T>`, but constants do not.

apiology: Ah, got it!

lib/solargraph/yard_map.rb
```diff
@@ -17,296 +17,5 @@ class NoYardocError < StandardError; end
...
-    end
-
-    # @return [void]
-    def process_requires
-      @gemset = process_gemsets
-      required.merge @gemset.keys if required.include?('bundler/require')
-      pins.clear
-      unresolved_requires.clear
-      missing_docs.clear
-      environ = Convention.for_global(self)
```
lekemula, line 179: @castwide I noticed that we don't call the `Convention.for_global` anymore, is that intentional?

Both `solargraph-rails` and `solargraph-rspec` plugins depend on it:

- https://github.com/lekemula/solargraph-rspec/blob/main/lib/solargraph/rspec/convention.rb#L85
- https://github.com/iftheshoefritz/solargraph-rails/blob/main/lib/solargraph-rails.rb#L25

Also, I was looking into adding support for `Struct.new` and `Data.define` as a conventions.

My idea was to introduce a new node processor for `class` to add appropriate methods on defined attributes there, something like:

```ruby
# frozen_string_literal: true

module Solargraph
  module Convention
    class StructDefinition < Base
      module NodeProcessors
        class ClassNode < Parser::NodeProcessor::Base
          def process
              # code to add struct/data conventional attributes goes here
          end
        end
      end

      # The Environ for a YARD map.
      # Subclasses can override this method.
      #
      # @param yard_map [YardMap]
      # @return [Environ]
      def global yard_map
        Solargraph::Parser::NodeProcessor.register(:class, NodeProcessors::ClassNode)
      end
    end
  end
end
```

Does this make sense? Can we ensure that `Convention.for_global` is called before parsing starts?

castwide: Good catch. We can reimplement global conventions, but they'll need to behave a little differently. The `global` method should most likely receive a `DocMap` instead of a `YardMap`. In `solargraph-rspec`, `Convention#global` ignores the argument, so it shouldn't make a difference. The `solargraph-rails` convention will need to change a little. I'll see if I can put together a PR.

I'll have to give some more thought to support for `Struct.new` and `Data.define`. You might not need a convention at all if you just need to register a node processor. Keep in mind that registering `ClassNode` will replace the existing `NamespaceNode` processor, so you might need to subclass it. There might also be other options besides modifying the node processors. Since `Struct` and `Data` both exist in Ruby core, I wouldn't be opposed to supporting them in Solargraph itself.

PR #782, opened 2025-03-13
Avoid trailing whitespace dependency in spec

apiology:


apiology:
Right now if you have an editor that trims trailing whitespace automatically, this spec breaks every time you save the file.

Ask me how I know!  😅

PR #783, opened 2025-03-13
Update Parser::NodeMethods to match parser gem implementation

apiology:


apiology:
This (combined with retiring RubyVM) fixes a number of strict-level type violations.

PR #784, opened 2025-03-13
Fix bug handling Array(A, B) syntax while resolving generics

apiology:
We generate Array(A, B) tuple syntax for some methods from rbs_map/conversions.rb that deal with tuples, but then turn around and turn them back into union arrays - Array<A, B> - after resolving generics.  This preserves the tuple type throughout resolution of generics.


lib/solargraph/complex_type/unique_type.rb
```diff
@@ -68,6 +68,7 @@ def items
       # @return [String]
       def to_rbs
         "#{namespace}#{parameters? ? "[#{subtypes.map { |s| s.to_rbs }.join(', ')}]" : ''}"
+      # "
```
apiology, line 71: Forgive me this weirdness - Emacs is not smart enough to parse the expression above it and thinks a runaway double-quoted string starts here.  The commented-out double quote seems enough to let it handle the rest of the file!

castwide: Heheh. VS Code has similar problems with certain string syntaxes. I have a feeling it's an unavoidable limitation of trying to identify Ruby string ranges with a regular expression.

PR #786, opened 2025-03-14
Reinstate Class core fills

castwide:
Version 0.52.0 removed most of the Class core fills, but there are some edge cases where they are still necessary. The specific case where I noticed it is `Gem::Specification.new`. Without the core fills, its inferred return type is `BasicObject`.

### DIFF
--- a/lib/solargraph/rbs_map/core_fills.rb
+++ b/lib/solargraph/rbs_map/core_fills.rb
@@ -48,7 +48,10 @@ module Solargraph
       end
 
       CLASS_RETURN_TYPES = [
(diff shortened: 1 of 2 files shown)


apiology:
Oh, nice!  This cleared up some mysterious behavior I was seeing on a method-generics PR that I'm working on.

lekemula:
Good that you caught this @castwide 👍 

I came across this [here](https://github.com/lekemula/solargraph-rspec/pull/8/commits/75ffb07e3562fdfb0b6086d71fc2d889d62bdf43) as well.

PR #787, opened 2025-03-15
More internal type annotations

apiology:


apiology:
These reduce our 'strong' typechecking backlog a bit.  

As a side note, I'm wondering what it would take to automatically insert annotations based on probed types where we can...would be a nice start for existing projects which aren't fully annotated.  Something like RuboCop's 'autocorrect' mode.  Hmm.

castwide:
> These reduce our 'strong' typechecking backlog a bit.
> 
> As a side note, I'm wondering what it would take to automatically insert annotations based on probed types where we can...would be a nice start for existing projects which aren't fully annotated. Something like RuboCop's 'autocorrect' mode. Hmm.

That's definitely possible. On a similar note, I'm working on a command that generates rbs files with probed types in place of undefined types. I can push a draft PR for it now that #781 is merged.

PR #789, opened 2025-03-15
Add #to_rbs methods to pins, use for better .inspect() output

apiology:


lib/solargraph/pin/signature.rb
```diff
@@ -21,11 +21,32 @@ def initialize generics, parameters, return_type, block = nil
...
-      # @return [Array<String>]
       def generics
-        @generics ||= []
+        @generics ||= [].freeze
+      end
+
+      # @return [String]
+      def to_rbs
+        @rbs ||= rbs_generics + '(' + parameters.map { |param| param.to_rbs }.join(', ') + ') ' + (block.nil? ? '' : '{ ' + block.to_rbs + ' } ') + '-> ' + return_type.to_rbs
       end
```
apiology, line 31: I'm not expecting this to be terribly correct at the moment against the spec, but it's been extremely helpful in the `.inspect` output of pins.

castwide:
Looks good, thanks! I appreciate the inspect enhancements.

PR #785, opened 2025-03-13
Remove redundant YIELDPARAMS backfills

lekemula:
These should be generated by rbs signatures and should work properly after https://github.com/castwide/solargraph/pull/743 which makes them redundant.

PR #791, opened 2025-03-16
Require Ruby >= 3

castwide:
#781 encountered a bug in RBS versions < 3.3. In order to continue development of RBS support, we'll have to require Ruby >=3 and drop support for 2.6/7.

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['2.6', '2.7', '3.0', '3.1', '3.2', '3.3', '3.4', 'head']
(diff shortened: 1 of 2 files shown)


lekemula:
I'm wondering whether this would also help with https://github.com/castwide/solargraph/pull/785 🤔

castwide:
> I'm wondering whether this would also help with #785 🤔 Maybe it wasn't the issue with ruby version but rather the rbs.

That's very possible. RBS >= 3.3 keeps us from needing to to backfill `YAML`, and the same might be true of the `Class` core fills.

lekemula:
@castwide indeed it worked! 🎉 

https://github.com/castwide/solargraph/actions/runs/13908424703/job/38916779646?pr=796

I pointed [the rebased PR](https://github.com/castwide/solargraph/pull/796) here if you want to include it here, or we can merge it afterward.

PR #792, opened 2025-03-16
Fix typeDefinitions for neovim

lekemula:
Hi,

I happened to be migrating from Vim to Neovim recently when I noticed that go-to-type-definitions were not working with [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md#solargraph) (default configs).

image

Looks like I had overlooked the capabilities registration in https://github.com/castwide/solargraph/pull/717, and apparently the VScode and coc.nvim plugins were more lenient hence didn't notice it.


castwide:
Looks good, thanks!

PR #794, opened 2025-03-17
Method generics resolution via argument types

apiology:
This resolves method-scoped generics by iterating through the known/inferred types (right now, method arguments), pattern matching them against the declared generic parameters, then iterating until progress is no longer being made.


lib/solargraph/complex_type.rb
```diff
@@ -30,6 +30,16 @@ def qualify api_map, context = ''
       ComplexType.new(types).reduce_object
     end
 
+    # @param generics_to_resolve [Enumerable<String>]]
+    # @param context_type [UniqueType, nil]
+    # @param resolved_generic_values [Hash{String => ComplexType}] Added to as types are encountered or resolved
+    # @return [self]
+    def resolve_generics_from_context generics_to_resolve, context_type, resolved_generic_values: {}
```
apiology, line 37: I'm open to ideas on the naming here - I don't think "context" is the right word.  "from assignment"?  "by pattern matching"?  Not sure.

spec/source/chain/call_spec.rb
```diff
@@ -305,6 +305,6 @@ def baz
     chain = Solargraph::Source::SourceChainer.chain(source, Solargraph::Position.new(4, 11))
     type = chain.infer(api_map, Solargraph::Pin::ROOT_PIN, api_map.source_map('test.rb').locals)
     # @todo It would be more accurate to return `Enumerator<Array<Integer>>` here
-    expect(type.tag).to eq('Enumerator<String, Array<Integer>>')
+    expect(type.tag).to eq('Enumerator<undefined, String, Array<Integer>>')
```
apiology: I'm not sure why this behavior change happened.  If you think it might be a concern, let me know - I can dig in on it with a fresh head later on.

apiology: Ah, I'm getting a better picture.  A small fix gets us a real type here.

apiology:
Note: This will need a `solargraph clear` for the improvement to take effect.  Not a problem after the release, when the cache key clears.

castwide:
Thanks! I'm going to merge this next because I suspect it'll be significant to the failing test in #793.

Issue #721, opened 2024-06-18
Ruby 2.6 isn't really supported


alagos:
[According to the gemspec](https://github.com/castwide/solargraph/blob/524c94e955938292fb63eee1741a4d95ce86f643/solargraph.gemspec#L24) it is, but in practice I'm getting:
```
ERROR:  Error installing solargraph:
	There are no versions of rubocop-ast (>= 1.31.1, < 2.0) compatible with your Ruby & RubyGems. Maybe try installing an older version of the gem you're looking for?
	rubocop-ast requires Ruby version >= 2.7.0. The current ruby version is 2.6.10.210.
```
I tried installing rubocop 1.50.2 [as 1.51.0 dropped the 2.6 support](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#changes-17) and that worked. 
So it should be dropped ruby 2.6 support or added to the gemspec:
```ruby
s.add_runtime_dependency 'rubocop', '~> 1.50.0'
```

kevgathuku:
Same issue here with Ruby 2.6. 
Thanks for the workaround @alagos  🎉🎉🎉🎉🎉🎉

castwide:
I'd say that Ruby 2.6 is supported, but requires some manual dependency management. Workarounds like this are a common requirement when working with older versions of Ruby. It's also worth noting that Bundler is smart enough to find a compatible Rubocop dependency on its own if you add Solargraph to your project's Gemfile.

Regardless, it won't be long before 2.6 gets dropped in a future release. Probably less than a year.

castwide:
Due to an issue with RBS < 3.3, the next release is likely to drop support for Ruby < 3.

castwide:
Version 0.53.0 released requiring Ruby >= 3.

PR #797, opened 2025-03-18
Retire more RubyVM-specific code

apiology:


lib/solargraph/parser.rb
```diff
@@ -9,10 +9,6 @@ module Parser
     class SyntaxError < StandardError
     end
 
-    def self.rubyvm?
-      false
-    end
```
apiology, line 14: @lekemula - whoops!  It looks like solargraph-rspec relies on this method.  I'll keep it around with this PR, but mark it as deprecated - maybe you could remove uses of it so it can be removed eventually?

lekemula: Good catch! Yeah, that would be nice, once https://github.com/castwide/solargraph/issues/721 gets released in the next version, I will deprecate any support for the "old-new" RubyVM parser and will open a PR to remove this method here. 👍

lekemula: Removed it here: [lekemula/solargraph-rspec@`ec94345` (#8)](https://github.com/lekemula/solargraph-rspec/pull/8/commits/ec94345c5f0124284797d7af7cbd5e207ee0171c)

apiology: Great!  Will your new code work with the current solargraph release?  If so, I think it'd be best for solargraph-rspec to release a new version before we merge this so that users can upgrade solargraph-rspec to fix the exception they see.  If not, let me know and I'll keep this method around and just mark it as deprecated.

lekemula: I’ll be releasing it the weekend after this one as I’m currently AFK - no laptop within reach at the moment. Hope that works for you!

apiology: Thanks for the heads-up - no rush here!

lekemula: Done: https://github.com/lekemula/solargraph-rspec/releases/tag/v0.5.1

PR #800, opened 2025-03-20
Various regression and future-feature specs

apiology:


spec/source_map/clip_spec.rb
```diff
@@ -1691,28 +1741,13 @@ def foo; end
     expect(type.to_s).to eq('Gem::Specification')
   end
 
-  it 'picks correct overload in Hash#transform_values!' do
```
apiology, line 1721: There's another identical copy of this spec above

apiology:
This is a dump of the specs I was hoarding in my stashes; some work today, some are marked as pending for future features to make work.

PR #803, opened 2025-03-20
At strong level, don't allow method calls off of nilable types

apiology:
Require that all unique types of a complex type offer the same method before it can match when "loose unions" mode is turned off.

### DIFF
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -36,9 +36,16 @@ module Solargraph
             []
           end
           return inferred_pins(found, api_map, name_pin.context, locals) unless found.empty?
-          # @param [ComplexType::UniqueType]
-          pins = name_pin.binder.each_unique_type.flat_map do |context|
-            api_map.get_method_stack(context.namespace == '' ? '' : context.tag, word, scope: context.scope)
+          if api_map.loose_unions
+            # fetch methods which ANY of the potential context types provide
+            pins = name_pin.binder.each_unique_type.flat_map do |context|
+              api_map.get_method_stack(context.namespace == '' ? '' : context.to_s, word, scope: context.scope)
+            end
+          else
+            # grab pins which are provided by every potential context type
+            pins = name_pin.binder.each_unique_type.map do |context|
+              api_map.get_method_stack(context.namespace == '' ? '' : context.to_s, word, scope: context.scope)
+            end.reduce(:&)
           end
           return [] if pins.empty?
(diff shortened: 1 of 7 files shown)


apiology:
Realistically, there's enough `return if foo.nil?` code out there that this will probably have to wait until we have flow-sensitive typing.  If that end up being difficult, I may propose hiding this under a `--level={alpha,beta}` level or somesuch.  Maybe a flag to turn on new checks introduced between major releases?  Not sure.

lib/solargraph/complex_type.rb
```diff
@@ -76,9 +76,13 @@ def self_to_type dst
...
     # @yieldparam [UniqueType]
+    # @yieldreturn [UniqueType]
     # @return [Array<UniqueType>]
-    def map &block
-      @items.map &block
+    # @sg-ignore Declared return type
+    #   ::Array<::Solargraph::ComplexType::UniqueType> does not match
+    #   inferred type ::Array<::Proc> for Solargraph::ComplexType#map
+    def map(&block)
+      @items.map(&block)
```
apiology, line 85: I suspect we're not understanding the &block construct

lib/solargraph/source/chain/call.rb
```diff
@@ -56,11 +56,15 @@ def resolve api_map, name_pin, locals
...
-          pins = name_pin.binder.each_unique_type.flat_map do |context|
+          pin_groups = name_pin.binder.each_unique_type.map do |context|
             ns_tag = context.namespace == '' ? '' : context.namespace_type.tag
             stack = api_map.get_method_stack(ns_tag, word, scope: context.scope)
             [stack.first].compact
           end
+          if !api_map.loose_unions && pin_groups.any? { |pins| pins.empty? }
+            pin_groups = []
+          end
+          pins = pin_groups.flatten.uniq(&:path)
```
apiology, line 67: This is a key bit - sharpening our existing behavior when asked to require all union types to match the behavior.

spec/type_checker/levels/alpha_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+        # @param b [String, nil]
+        # @return [void]
+        def bar(b)
+         foo(b)
+        end
+      ))
+      expect(checker.problems.map(&:message))
+        .to eq(['Wrong argument type for #foo: a expected String, received String, nil'])
+    end
+  end
```
apiology, line 24: This is the real win here - the ability to check for potential nil issues

apiology:

PR #804, opened 2025-03-21
Resolve return type of block symbols and use for method resolution

apiology:
Builds on #793 as a diff against the infer-block-symbol branch

### DIFF
--- a/lib/solargraph/source/chain/block_symbol.rb
+++ b/lib/solargraph/source/chain/block_symbol.rb
@@ -5,7 +5,14 @@ module Solargraph
     class Chain
       class BlockSymbol < Link
         def resolve api_map, name_pin, locals
-          [Pin::ProxyType.anonymous(ComplexType.try_parse('Proc'))]
+          return [Pin::ProxyType.anonymous(ComplexType::UNDEFINED)] if locals.empty?
+          first_arg_return_type = locals.first.return_type
+          first_arg_return_type.items.flat_map do |unique_type|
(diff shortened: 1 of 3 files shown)


apiology:
@castwide: This is probably the wrong way to accomplish things, but this is what I had in mind to resolve `[].map(&:block_symbol)`

castwide:
I think the return type of the `&:block_symbol` pin should be a `Proc` and the inferred type of `[].map(&:block_symbol)` should be `undefined`.

The concept is simpler if you look at it with a block variable:

```ruby
def foo(&block)
  block # The return type is `Proc`
end

block = proc { 100 } # The return type of the `block` variable is `Proc`
block # Chain inference on `block` returns `Proc`
block.call # Chain inference on `call` returns `Integer`
instance_exec(&block) # Chain inference on `instance_exec` returns `Integer`
```

With a block-pass symbol, `proc(&:block_symbol)` is equivalent to `proc { |arg| arg.block_symbol }`.

```ruby
block = proc(&:to_s)
block # Chain inference on `block` returns `Proc`
[0].map(&block) # Chain inference on `map` returns `Array<String>
```

apiology:
Incidentally, rbs represents procs like this:

```
^(Integer) -> String                  # A procedure with an `Integer` parameter and returns `String`
^(?String, size: Integer) -> bool     # A procedure with `String` optional parameter, `size` keyword of `Integer`, and returns `bool`
```

We could probably use something them like this, which may work more easily with the existing machinery:

```
Proc{(a, b, c) => String}
```

PR #805, opened 2025-03-22
Add :if support to NodeChainer for if statements as lvalues

apiology:
Example:

```ruby
        # the output of the "if" here is the receiver of the String#+ method
        a = (b ? 'foo' : 'bar') + 'baz' 
```

Previously the :if node in the chain would become an undefined link, breaking the chain and resulting in an `Unresolved call to +` error in strict typechecking


apiology:
This knocks about 25 items off of our strict list - with the other fixes, we're down to 173!

castwide:
:100: :shipit:

PR #813, opened 2025-03-23
[breaking] Fix more complex_type_spec.rb cases

apiology:
* Consolidate the code that generates unique type substrings to a single place
* Clean up more paths to pass through 'rooted' flag unchanged
* [breaking] Drop special-case interpretation of, e.g., `Array<(String)>` as 'one element tuple' for future consistency with https://yardoc.org/types - users should instead use Array(String)

  * (String) is a one element tuple in https://yardoc.org/types
  * <String> is an array of zero or more Strings in https://yardoc.org/types
  * Array<(String)> could be an Array of one-element tuples or a one element tuple.  https://yardoc.org/types treats it as the former.
  * Array<(String), Integer> is not ambiguous if we accept (String) as a tuple type, but not currently understood by Solargraph.
* Close out complex_type_spec \@todos for better RBS output


apiology:
See also https://github.com/apiology/solargraph/pull/3, which builds on this PR:
* [Reduce use of ComplexType.parse() to preserve rooted? information](https://github.com/apiology/solargraph/pull/3)

castwide:
Looks good, thanks!

PR #815, opened 2025-03-23
Fix issue with wrong signature selection by call with block node

apiology:
We were picking the wrong signature for Enumerator#select and Enumerator#find because Chain::Call didn't have access to the block that was passed in and didn't think the arity matched as a result.

### DIFF
--- a/lib/solargraph/pin/block.rb
+++ b/lib/solargraph/pin/block.rb
@@ -3,19 +3,22 @@
 module Solargraph
   module Pin
     class Block < Closure
-      # The signature of the method that receives this block.
-      #
       # @return [Parser::AST::Node]
       attr_reader :receiver
 
+      # @return [Parser::AST::Node]
+      attr_reader :node
+
       # @param receiver [Parser::AST::Node, nil]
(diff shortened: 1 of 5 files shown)


lib/solargraph/pin/block.rb
```diff
@@ -3,19 +3,22 @@
 module Solargraph
   module Pin
     class Block < Closure
-      # The signature of the method that receives this block.
-      #
       # @return [Parser::AST::Node]
       attr_reader :receiver
```
apiology, line 7: Not sure if this was the intention, but `receiver` is the :send node, not the :def node as the comment implied

lib/solargraph/source/chain/call.rb
```diff
@@ -72,10 +72,8 @@ def inferred_pins pins, api_map, context, locals
               match = true
 
               atypes = []
-              block_parameter = nil
               arguments.each_with_index do |arg, idx|
                 param = ol.parameters[idx]
-                atype = nil
```
apiology, line 78: Unneeded / unused

PR #818, opened 2025-03-24
Use return type of literal blocks in inference

apiology:
Also:
* Document block symbol handling issue


lib/solargraph/pin/signature.rb
```diff
@@ -27,7 +27,7 @@ def generics
 
       # @return [String]
       def to_rbs
-        @rbs ||= rbs_generics + '(' + parameters.map { |param| param.to_rbs }.join(', ') + ') ' + (block.nil? ? '' : '{ ' + block.to_rbs + ' } ') + '-> ' + return_type.to_rbs
+        rbs_generics + '(' + parameters.map { |param| param.to_rbs }.join(', ') + ') ' + (block.nil? ? '' : '{ ' + block.to_rbs + ' } ') + '-> ' + return_type.to_rbs
```
apiology, line 30: Whoops - caching this led to the proxied signatures with resolved generics still showing old RBS info

castwide:
:+1: :shipit:

PR #836, opened 2025-03-26
Tuple enabler: infer literal types and use them for signature selection

apiology:


spec/source/chain_spec.rb
```diff
@@ -197,7 +197,7 @@ class NotCorrect; end
     api_map = Solargraph::ApiMap.new
     chain = Solargraph::Parser.chain(source.node)
     type = chain.infer(api_map, Solargraph::Pin::ROOT_PIN, [])
-    expect(type.to_s).to eq('String')
```
apiology, line 200: I thought about having `.to_s` or `.tag` apply the simplifications by default, but oh boy did they make debugging a pain.  Probably best we make the simplification explicit.

spec/source_map/clip_spec.rb
```diff
@@ -1933,7 +1948,7 @@ def foo
     expect(type.to_s).to eq('Integer')
   end
 
-  xit 'identifies tuple types' do
+  xit 'dereferences tuple types with [](idx) via literals' do
```
apiology, line 2064: This is ultimately what I'd like to get working!  

One way to get there would be to create static signatures for, say, 0-10.  It would be a little hokey, but would allow us to resolve an enormous number of tuple dereferences in the real world.

apiology: https://github.com/apiology/solargraph/pull/4 implements this idea!

spec/source_map/clip_spec.rb
```diff
@@ -2066,6 +2087,197 @@ def foo
...
+
+    clip = api_map.clip_at('test.rb', [21, 6])
+    expect(clip.infer.to_s).to eq('Integer')
+
+    clip = api_map.clip_at('test.rb', [23, 6])
+    expect(clip.infer.to_s).to eq('Float')
+  end
+
+  it 'u
+  tch' do
```
apiology:
```suggestion
  it 'can use strings and symbols to choose a signature' do
```

castwide:
Tested locally, works great! This and a few of the other big changes in the pipeline should be good to go in 0.55.0.

PR #844, opened 2025-03-28
Consolidate parameter handling into Pin::Closure

apiology:


apiology:
I've been dealing with more nuanced cases matching signatures against call arguments recently.  Before I go much further into it, I'd like to have a single interface we can call into where that sort of logic can live, so that TypeChecker can also benefit from it.  

Pin::Closure seems pretty close to me, even if classes and namespaces don't deal with parameters.  That said, I can move this code to its own subclass or mix-in ("Pin::Callable"?) if you have a different take.

castwide:
Consolidating `parameters` is a solid idea, but I would definitely prefer it to be in a `Pin::Callable` subclass or mixin.

PR #824, opened 2025-03-25
Remove redundant core fills

castwide:
Clean up some of the core fills and related features in favor of better RBS integration.

PR #849, opened 2025-03-29
[regression] Restore 'Unresolved call' typecheck for stdlib objects

apiology:
The 'Unresolved call to' typecheck suppresses warnings for calls against non-RBS-origin pins.  However, the labeling assigning :rbs as the source of pins from conversions.rb was recently removed as part of an unrelated cleanup.


apiology:
See https://github.com/castwide/solargraph/pull/824/files#r2019815203 for details on the removal

castwide:
:+1:

PR #850, opened 2025-03-29
[regression] Ensure YardMap gems have return type for Class<T>.new

apiology:


apiology:
I'm also seeing this in StringIO.new, but the fix above does not help, even though it is inserting the relevant Pin.  I am digging in now

castwide:
Thanks for iterating on this! I'm working on a separate PR to fix some weirdness around new/initialize in workspace mapping, and then we should be good for another patch release 🙌

PR #857, opened 2025-03-30
Memoize result of Chain#infer

apiology:
This brings down typechecking time on my machine from >5 minutes to ~20 seconds

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -91,7 +91,7 @@ module Solargraph
     end
 
     protected def equality_fields
-      [self.class, @source_map_hash, @cache, implicit, @doc_map, @unresolved_requires, @missing_docs]
+      [self.class, @source_map_hash, implicit, @doc_map, @unresolved_requires, @missing_docs]
     end
 
     # @return [::Array<Gem::Specification>]
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -29,6 +29,25 @@ module Solargraph
       index pins
     end
 
+    #
+    # This is a mutable object, which is cached in the Chain class -
+    # if you add any fields which change the results of calls (not
+    # just caches), please also change `equality_fields` below.
+    #
+
+    def eql?(other)
+      self.class == other.class &&
+        equality_fields == other.equality_fields
+    end
+
+    def ==(other)
+      self.eql?(other)
+    end
+
+    def hash
+      equality_fields.hash
+    end
+
     # @param pins [Array<Pin::Base>]
     # @return [self]
     def index pins
@@ -69,6 +88,10 @@ module Solargraph
       @unresolved_requires = @doc_map.unresolved_requires
       @missing_docs = [] # @todo Implement missing docs
       self
+    end
+
+    protected def equality_fields
+      [self.class, @source_map_hash, @cache, implicit, @doc_map, @unresolved_requires, @missing_docs]
     end
 
     # @return [::Array<Gem::Specification>]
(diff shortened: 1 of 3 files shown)


lib/solargraph/source/chain.rb
```diff
@@ -81,7 +83,25 @@ def define api_map, name_pin, locals
...
       def infer api_map, name_pin, locals
+        out = nil
+        cached = @@inference_cache[[node, name_pin&.return_type, locals]]
+        return cached if cached && @@inference_invalidation_key == api_map.hash
+        out = infer_uncached api_map, name_pin, locals
+        if @@inference_invalidation_key != api_map.hash
+          @@inference_cache = {}
+          @@inference_invalidation_key = api_map.hash
+        end
+        @@inference_cache[[node, name_pin&.return_type, locals]] = out unless node.nil?
```
apiology: @castwide: Does this seem like a reasonable way to ensure cache coherence?

apiology:
Yeah, I'm going to let it bake here too - I hit one issue last night and figured it out this morning, relating to a 'self' node.  I think this may be a bug in Parser::AST, which adds 'location' in Parser::AST::Node to its superclass AST::Node, but doesn't overwrite the #hash and #eql?

https://github.com/whitequark/ast/blob/master/lib/ast/node.rb

https://github.com/whitequark/parser/blob/05b88aa0468ce29dfa21e1aff2f0506b1d5d82e1/lib/parser/ast/node.rb#L17

I think adding node.location explicitly is probably a good idea.

I also wasn't including 'links' in the key, which even without the above fix was enough to resolve the bug I was seeing, so maybe I'll add both for good measure.

apiology:
Ah, I see this has been hashed out (so to speak) in https://github.com/whitequark/parser/issues/184

Consider this a datapoint to that discussion, but no objections here if you want to close this issue.

apiology:
Using 'links' directly killed performance - there's probably something in there using object identity instead of value equality.  I'll lean on `links.map(&:word)` for now and see how that shakes out.

castwide:
I encountered issues with the inference cache when workspaces change in the language server. The library modifies its `ApiMap` in place via `#catalog`, so its hash never changes. This has two side effects:

1. Changes can invalidate cached entries so they become unreachable, but they stay in memory.
2. Sometimes a change in a source map can impact chain inference in other source maps that haven't changed, so the other maps' cached inferences become incorrect.

I originally tried to solve this by caching clips in `ApiMap::Cache`, but it didn't perform significantly better (and in some cases performed worse). Caching at the chain level seems like the better choice, but we might need to invalidate the cache based on different criteria. One possible solution is to create a timestamp attribute that gets updated in `ApiMap#catalog` and check both the timestamp and `ApiMap#hash` for validation. Another is to cache inferences in `ApiMap` itself and clear them in `#catalog`.

apiology:
What field of the api map was modified?  If you could push up a failing spec I could take a look.  In principle `.hash()` and `.eql()` should be taught to pick up on something like that if they don't do so today for whatever reason.

apiology:
Thanks!!  Will take a look.

apiology:
All right - turns out that by default eql?() and hash() operate just on object identity, rather than recursing down all of the instance variables like I was expecting.  Array and Hash are exceptions - they do recurse into the contained objects.

So we'll want to implement hash() and eql?() for anything that's mutable.  That's definitely ApiMap, but let me know if you think of anything else that changes state over time as well.

lib/solargraph/api_map.rb
```diff
@@ -71,6 +90,10 @@ def catalog bench
       self
     end
 
+    protected def equality_fields
+      [self.class, @source_map_hash, @cache, implicit, @doc_map, @unresolved_requires, @missing_docs]
```
castwide: You can probably leave `@cache` out of the `equality_fields`. Its `hash` doesn't change, and even if it did, cached queries don't indicate a change in the map.

castwide:
:100: :shipit:

PR #856, opened 2025-03-30
Flow-sensitive typing - automatically downcast from is_a? calls

apiology:

PR #863, opened 2025-04-01
Specify more inference behavior for variable reassignment

apiology:


apiology:
@castwide: As I work towards flow-sensitive typing, I'd like to get consensus on what exactly we ideally want types to be inferred as when variables are reassigned.  Would you mind looking through these to see if any of them seem controversial?

spec/source_map/clip_spec.rb
```diff
@@ -2048,4 +2048,164 @@ def foo
...
+      bar = bar.bar
+      bar
+  ), 'test.rb')
+    api_map = Solargraph::ApiMap.new.map(source)
+    clip = api_map.clip_at('test.rb', [6, 6])
+    expect(clip.infer.to_s).to eq('Foo')
+
+    clip = api_map.clip_at('test.rb', [8, 6])
+    expect(clip.infer.to_s).to eq('Foo')
+  end
```
apiology, line 2107: The above are behavior that I managed to break in some of my early attempts, so I figured I'd get them merged in advance to tie my hands in #856 :)

castwide:
Tests look reasonable to me. :+1: :shipit:

PR #864, opened 2025-04-01
Adjust local variable presence to start after assignment, not before

apiology:
Currently local variables are considered present within their assignment with the types of their assignments:

```
# a is considered as an integer in the next line
a = 123
```

However, in reality, a name is created, but assigned nil in its initial assignment code until the assignment statement is complete:

```sh
a = a + 1 # NoMethodError: undefined method `+' for nil:NilClass
```

(Note that we don't get a `NameError: undefined local variable or method `a' for main:Object`, so the variable has been created but not assigned while evaluating the rhs)

Currently, if you ask Solargraph to infer the position of 'a' in `a = 123', it gives Integer, which seems premature.  

The current behavior may also seem innocuous, but it's getting in the way of my approach to #856, so I'd like to back off and instead have a local variable assignment's presence start on the next statement.  Up until that point, it will have the same type it had before the statement.  (`nil` would probably be better if it had no previous value, but I'm not sure that's worth the effort to emulate)


castwide:
> The big question in my mind - can we live with the impact this will have on the LSP world? If not, I might come back with a proposal that adds a 'human visibility' concept which is maybe different than 'machine visibility' :)

This is probably fine on the LSP side. My main concern is that messages like `textDocument/hover` should still provide the inferred type of a variable assignment. So far so good:

Screenshot 2025-04-05 092909

I'm going to test a little more locally, but I haven't seen any red flags yet.

castwide:
:+1: :shipit:

PR #867, opened 2025-04-02
Refactor TypeChecker#argument_problems_for for type safety

apiology:
This should put it in a state that it can pass type checking once we have another feature or two


lib/solargraph/shell.rb
```diff
@@ -173,6 +173,7 @@ def typecheck *files
       }
       puts "Typecheck finished in #{time.real} seconds."
       puts "#{probcount} problem#{probcount != 1 ? 's' : ''} found#{files.length != 1 ? " in #{filecount} of #{files.length} files" : ''}."
+      # "
```
castwide, line 176: I'm working on support for `textDocument/semanticTokens`. Hopefully it'll remove the need for this workaround.

Issue #861, opened 2025-03-31
Invalid close in type DidYouMean::SpellChecker


body-clock:
In my attempts to debug missing hover documentation, I came across the `solargraph scan -v` command. Running this exposes an error in my workspace. Seems to have something to do with `Thor` and its dependency on `DidYouMean`. I'm getting a `[Solargraph::ComplexTypeError]: Invalid close in type DidYouMean::SpellChecker)` while running the following:
```shell
❯ solargraph -v
0.53.4
❯ rails -v
Rails 7.1.5.1
❯ ruby -v
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [arm64-darwin23]
```

Seems related to #634 but is not resolved with the safe navigation patch: 
```
Thor::Error (/Users/pperkins/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/thor-1.3.2/lib/thor/error.rb 19)
Thor::Error | ::StandardError
Thor::UndefinedCommandError (/Users/pperkins/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/thor-1.3.2/lib/thor/error.rb 23)
Thor::UndefinedCommandError | Thor::Error
Thor::UndefinedCommandError | Thor::UndefinedCommandError.if defined?(DidYouMean::SpellChecker) && defined?(DidYouMean::Correctable) # rubocop:disable Naming/ConstantName
  Moduleif defined?(DidYouMean::SpellChecker) && defined?(DidYouMean::Correctable) # rubocop:disable Naming/ConstantName
  Module.new do
    def to_s
      super + DidYouMean.formatter.message_for(corrections)
    end

    def corrections
      @corrections ||= self.class.const_get(:SpellChecker).new(self).corrections
    end
  end
end
Thor::UndefinedCommandError::SpellChecker (/Users/pperkins/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/thor-1.3.2/lib/thor/error.rb 24)
Thor::UndefinedCommandError::SpellChecker | ::Object
Thor::UndefinedCommandError::SpellChecker#error (/Users/pperkins/.local/share/mise/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/thor-1.3.2/lib/thor/error.rb 25)
[...]

PR #869, opened 2025-04-02
Fix flaky rename specs

castwide:
Rename specs occasionally fail because the host is still updating when the test creates the rename message.


castwide:
Merging now to unblock other PRs. I'm also working on a more robust solution that should help with synchronization-related bugs like #861

PR #877, opened 2025-04-07
[breaking] Reimplement global conventions

castwide:
As mentioned in a [comment in #781](https://github.com/castwide/solargraph/pull/781#discussion_r2030116341), the switch from `YardMap` to `DocMap` blocked processing of global conventions. We need a way forward for plugins that depend on it. The proposal here is that `Convention::Base#global` accept a `DocMap` argument. Plugins that expect a `YardMap` will need to be updated accordingly.

### DIFF
--- a/lib/solargraph/doc_map.rb
+++ b/lib/solargraph/doc_map.rb
@@ -225,7 +225,6 @@ module Solargraph
 
     def gemspecs_required_from_external_bundle
       logger.info 'Fetching gemspecs required from external bundle'
-      puts workspace.inspect
       return [] unless workspace&.directory
 
       Solargraph.with_clean_env do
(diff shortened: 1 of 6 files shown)


apiology:
@castwide - how far away do you think this one?  I'm thinking through upcoming releases of solargraph-rails.  I'm guessing solargraph-rails should pin to a solargraph <0.53.0 until then, yeah?

lib/solargraph/doc_map.rb
```diff
@@ -218,6 +225,7 @@ def gemspecs_required_from_bundler
 
     def gemspecs_required_from_external_bundle
       logger.info 'Fetching gemspecs required from external bundle'
+      puts workspace.inspect
```
apiology:
```suggestion
```

PR #883, opened 2025-04-08
Don't false-alarm over splatted non-final params in typechecking

apiology:


apiology:
With the open PRs here and some improvements pending PRs to https://github.com/ruby/rbs, we're down to 90 problems at strict level!

Biggest remaining needs:
* Record support (14 false alarms)
* The block argument to Mutex#synchronize doesn't seem to play well with inference (12 false alarms)
* Dealing with `@foo ||=` in getters (10 false alarms)
* tuples of tuples (8 false alarms)

The rest of the 46 issues form a long tail of one-off things.  At some point I'll propose marking the remaining bits with `@todo` and `@sg-ignore` and either move onto strong-level or take a crack at some other codebases next.

PR #884, opened 2025-04-08
Remove accidental inclusion of Module's methods in objects

apiology:
The instance methods on 'Module' aren't available on arbitrary objects - don't include them in the api_map as such.

The inaccurate method list was causing problems at type-checking time - when an external gem (RBS, in this case) had a method ("name") that we didn't have a type for, it could look for the types on a method in Method--in this case Method#name, which returns String.  Since RBS has its own class ("TypeName"), this would fail to typecheck at :strict.

Also update a clip_spec test, which was expecting incorrect behavior in completions.

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -371,7 +371,7 @@ module Solargraph
           end
         end
         result.concat inner_get_methods('Kernel', :instance, [:public], deep, skip) if visibility.include?(:private)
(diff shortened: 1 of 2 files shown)


castwide:
Good catch, thanks!

Just a quick heads up, I expect to release v0.54.0 this week. I'll probably limit merges to small patches and bug fixes in the meantime and resume merging the bigger changes shortly after the release.

PR #885, opened 2025-04-09
Block method resolution improvements

apiology:


apiology:
These changes tamed the issues we have being able to infer the return type of `Mutex.synchronize { }` by being able to infer the return value of the block - there were a good number of strict typechecking issues in `library.rb` that are resolved with this.

There are probably interesting `@yieldself` use cases not yet captured in our specs - hopefully having access to the Block pin from the Call link will help with that.

Along the way, I took a crack at describing the semantics of `name_pin` - if the definitions don't feel right, or it doesn't look like my code changes are consistent with them, just let me know and I'll work through the updates.

lib/solargraph/source/chain.rb
```diff
@@ -73,25 +73,41 @@ def base
...
       # @param api_map [ApiMap]
-      # @param name_pin [Pin::Base]
-      # @param locals [::Enumerable<Pin::LocalVariable>]
+      # @param name_pin [Pin::Closure] the surrounding closure pin for
+      #   the statement represented by this chain for type resolution
+      #   and method pin lookup.
+      #
+      #   For method calls (Chain::Call objects) as the first element
+      #   in the chain, 'name_pin.binder' should return the
+      #   ComplexType representing the LHS / "self type" of the call.
```
castwide, line 82: LGTM :+1:

PR #887, opened 2025-04-10
Add regression test around assignment in return position

apiology:


apiology:
I broke this behavior in one of my pending branches (working now to figure out which one) - so I figure it'd be good to get the spec on master sooner rather than later.

apiology:
Ah, it was https://github.com/castwide/solargraph/pull/856 - I merged in this branch there to ensure the PR will fail CI until I get a fix in.

apiology:
Fixed on those branches!

PR #891, opened 2025-04-11
Handle super and yield scenarios from blocks

apiology:


spec/source_map/clip_spec.rb
```diff
@@ -2359,6 +2359,24 @@ def blah
...
+        [1,2,3].each do |i|
+          baz = yield
+          baz
+        end
+      end
+    ), 'test.rb')
+    api_map = Solargraph::ApiMap.new.map(source)
+
+    clip = api_map.clip_at('test.rb', [3, 10])
+    expect(clip.infer.to_s).to eq('Integer')
```
castwide, line 2377: One minor suggestion for this test. Other than that, looks good :+1:

```suggestion
    clip = api_map.clip_at('test.rb', [3, 10])
    expect(clip.infer.to_s).to eq('Integer')

    clip = api_map.clip_at('test.rb', [7, 10])
    expect(clip.infer.to_s).to eq('Integer')
```

apiology: Thanks!

PR #900, opened 2025-04-16
Remove redundant extend in Parser

castwide:
The redundant `extend` is reinitializing `ClassMethods` and emitting warnings.

### DIFF
--- a/lib/solargraph/parser.rb
+++ b/lib/solargraph/parser.rb
@@ -9,14 +9,11 @@ module Solargraph
     class SyntaxError < StandardError
     end
 
-    ClassMethods = ParserGem::ClassMethods
-
     # @deprecated
     Legacy = ParserGem
 
     ClassMethods = ParserGem::ClassMethods
-
-    extend ParserGem::ClassMethods
+    extend ClassMethods
 
     NodeMethods = ParserGem::NodeMethods
   end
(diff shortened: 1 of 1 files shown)


apiology:
What would you say to turning RuboCop on in CI to catch things like this?  We could use the todo feature to adopt it gradually and be pragmatic about the settings.  I can throw together a PR as a proposal.

castwide:
> What would you say to turning RuboCop on in CI to catch things like this? We could use the todo feature to adopt it gradually and be pragmatic about the settings. I can throw together a PR as a proposal.

Sounds good to me. There's currently a `lib/.rubocop.yml` file in the repo. We can probably delete it and start fresh. I originally thought we might need different configs for `lib` and `spec`, but it's fine if a single config at the project root is sufficient.

lib/solargraph/parser.rb
```diff
@@ -9,14 +9,11 @@ module Parser
...
 
-    ClassMethods = ParserGem::ClassMethods
-
     # @deprecated
     Legacy = ParserGem
 
     ClassMethods = ParserGem::ClassMethods
-
-    extend ParserGem::ClassMethods
+    extend ClassMethods
```
apiology:
```suggestion
    # @todo should be able to just 'extend ClassMethods' here and
    #   typecheck things off it in strict mode
    extend ParserGem::ClassMethods
```

apiology: Solargraph doesn't like extending off of a module alias :(

PR #901, opened 2025-04-18
Allow log level to be overridden per file

apiology:
Useful for debug-level logging being turned on selectively at dev time

The idea here is that we could have code like this:

```ruby
class SomeRandomClass
  def important_method
    data = logic
    logger.debug { "something happened: #{data}" }
    logic
  end
  
  include Logging
end
```

We can selectively turn on those debugging lines at dev time by temporarily adding a `log_level` method:

```ruby
class SomeRandomClass

  def log_level
    :debug
  end
  
  def important_method
    data = logic
    logger.debug { "something happened: #{data}" }
    logic
  end
  
  include Logging
end
```

As a result, only the debug lines we care about are brought in while debugging something that interacts with `SomeRandomClass`


lib/solargraph/logging.rb
```diff
@@ -29,9 +29,26 @@ module Logging
...
+    #
+    # @return [Symbol]
+    def log_level
+      :warn
+    end
+
     # @return [Logger]
     def logger
-      @@logger
+      if LOG_LEVELS[log_level.to_s] == DEFAULT_LOG_LEVEL
```
apiology: Can we cache this as `@logger` here?

apiology: **Claude:**

Yes - cached as `@logger` in 054171d, since `logger` is a `module_function`
mixed into each including class, `@logger` is scoped per instance the
same way `log_level` is. Added a spec confirming the same instance comes
back across calls.

PR #903, opened 2025-04-19
Library sync and cache invalidation

castwide:
* Library methods that use cursors need to sync_catalog first.
* Ensure that changing a method's return type in YARD tags resets the inference cache.
* Ensure that changes to local pins in source maps reset the inference cache.
* Host attaches opened and unmerged sources to libraries

### DIFF
--- a/lib/solargraph/language_server/host/message_worker.rb
+++ b/lib/solargraph/language_server/host/message_worker.rb
@@ -6,7 +6,10 @@ module Solargraph
       # A serial worker Thread to handle message.
       #
       # this make check pending message possible, and maybe cancelled to speedup process
+      #
       class MessageWorker
+        UPDATE_METHODS = ['textDocument/didOpen', 'textDocument/didChange', 'workspace/didChangeWatchedFiles'].freeze
+
         # @param host [Host]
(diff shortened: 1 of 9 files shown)


castwide:
The three major bottlenecks in library updates are source parsing, source mapping, and api_map cataloging. The top-level approach for improving them is to apply changes at the moment they're needed instead of when they're received.

* Updating (synchronizing) sources creates a new source that tracks the changes without applying them. The library will parse them when a query requires the updates to be executed. This allows sources to queue changes from the language server (e.g., via `textDocument/didChange`) instead of triggering a parse for each change.
* Source maps will not generate pins until they're required by a library catalog.
* Source maps used to perform a `#try_merge!` to determine if a source map's pins could be merged into an existing map to reduce cataloging. Any benefit from this process was lost to the fact that the source map needed to be generated before it could be merged. JIT mapping in the library is ultimately more efficient.
* We can explore whether `try_merge!` would be more effective as part of the `Library#sync_catalog` process, where it still might reduce `ApiMap#catalog` overhead. The question is how much the reduction in cataloging compares to the addition of merge attempts.

PR #907, opened 2025-04-21
Signature selection improvements

apiology:
Removes three workarounds in the code that were holding us back from providing complete types during method resolution:

- Use only the latest element of the method stack, instead of considering all of them
- Discard other types when combined with undefined
- Treat all defined pins as equally valid instead of onlyone

In order to remove these workarounds while not regressing any unit tests or adding any strict-level false alarms, there were a variety of fixes/improvements needed:

- Use different convention for naming anonymous args to avoid conflicts with core definitions
- Centralize erasing remaining generics at end of api_map#get_methods to better handle include/superclass cases
- Drop unneeded Object#tap fill, now fully handled via RBS conversions
- Add _ToAry include to Array pending automatic RBS interface support
- Update type checker to use typify() method to pull tags from down the method stack
- Method alias fixes 
- Fix CoreFills type
- Deduplicate UniqueType key/subtypes when they are implicit unions
- Teach node processor to understand += and friends
- Fix uses of Pin#context and #binder to support qualify()

### DIFF
--- a/spec/source_map/clip_spec.rb
+++ b/spec/source_map/clip_spec.rb
@@ -1907,6 +1907,7 @@ describe Solargraph::SourceMap::Clip do
     type = clip.infer
     expect(type.tag).to eq('Array<Integer>')
 
+    # api_map = Solargraph::ApiMap.new.map(source)
     clip = api_map.clip_at('test.rb', [5, 10])
     type = clip.infer
     expect(type.tag).to eq('Integer')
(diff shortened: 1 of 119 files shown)


spec/source_map/clip_spec.rb
```diff
@@ -1052,6 +1052,25 @@ def foo a
     expect(clip.infer.to_s).to eq('Foo, Bar')
   end
 
+  # pending https://github.com/castwide/solargraph/pull/836
+  xit 'handles parallel type possibilities with #nil?' do
```
apiology, line 1066: I suspect there will also some untested related regressions with this PR solo - to be safe, I think we should release this PR in the same release as #836 .  I'll keep them as separate PRs for ease of review in the meantime.

castwide:
Looks great, thanks! Lots of nice enhancements to type inference here.

I encountered a flaky test, but it's not a blocker. It looks like another discrepancy in array ordering on Windows. I can patch it in a separate PR.

PR #917, opened 2025-04-28
[regression] Avoid non-deterministic behavior with better cache key

apiology:
I've seen some dodgy non-repeatable behavior from Solargraph in the last few days.  I was able to pin it down (so to speak) to a particular use case I could repeat, and it seemed to be related to links being cached by their words alone in Chain.

Unfortunately, I did not record enough about the scenario to reproduce it :(

However, this change ought to strengthen the coherency of the links part of our cache key without affecting performance and hopefully squash this.

### DIFF
--- a/lib/solargraph.rb
+++ b/lib/solargraph.rb
@@ -42,6 +42,7 @@ module Solargraph
   autoload :Logging,          'solargraph/logging'
   autoload :TypeChecker,      'solargraph/type_checker'
   autoload :Environ,          'solargraph/environ'
+  autoload :Equality,         'solargraph/equality'
   autoload :Convention,       'solargraph/convention'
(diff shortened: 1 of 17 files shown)


castwide:
Were you getting stale and/or incorrect type inference from the language server? I saw that happen a few times. My first impression was that it was related to some lingering issues with library synchronization that I resolved in 0.54.1. That was almost certainly part of the problem, but I can see how cache inference problems might have contributed. Unfortunately, I don't have a reproducible example either :(

apiology:
I think that's fair to say--I was seeing results from the typechecker that didn't match what I was seeing in the LSP.

When I first made this fix I had a spec that passed if I re-used the same api_map on the same source with two separate clips.  Each worked individually, and they both worked if I created separate api_map objects for each clip.

PR #922, opened 2025-05-05
Fix pin inference stack

castwide:
Chain inference can encounter pin identity collisions when a variable and a method have the same name. Change the inference stack to use normal object equality instead.


castwide:
The second commit resolves some discrepancies in tag sorting order that only appear to occur in Windows.

PR #923, opened 2025-05-05
Help typechecker understand type of Namespace's @name variable

apiology:
Small rearrangement of code to make this method easier for the typechecker to understand.


apiology:
With this and the other pending changes, we're down to ~25 strict typechecking issues!

PR #925, opened 2025-05-06
Refactor pin equality for performance

castwide:
#922 introduced changes to the chain inference stack that caused some performance issues in conjunction with changes to pin equality evaluation in #917.

* Pin equality evaluation uses the faster `nearly?` version instead of `Equality`.
* `Chain#inference` is refactored for clarity.

### DIFF
--- a/lib/solargraph/pin/base.rb
+++ b/lib/solargraph/pin/base.rb
@@ -134,11 +134,10 @@ module Solargraph
       # Pin equality is determined using the #nearly? method and also
       # requiring both pins to have the same location.
       #
-      def eql? other
+      def == other
         return false unless nearly? other
         comments == other.comments && location == other.location
       end
-      alias == eql?
 
       # The pin's return type.
       #
--- a/lib/solargraph/pin/base.rb
+++ b/lib/solargraph/pin/base.rb
@@ -8,7 +8,6 @@ module Solargraph
       include Common
       include Conversions
       include Documenting
-      include Equality
 
       # @return [YARD::CodeObjects::Base]
       attr_reader :code_object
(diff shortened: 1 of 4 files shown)


lib/solargraph/pin/base.rb
```diff
@@ -148,6 +131,15 @@ def nearly? other
           )
       end
 
+      # Pin equality is determined using the #nearly? method and also
+      # requiring both pins to have the same location.
+      #
+      def eql? other
+        return false unless nearly? other
+        comments == other.comments && location == other.location
+      end
```
apiology, line 140: My assumption is that since we're now not overriding hash(), Object#hash() will return something fairly unique per object in memory, making eql()? above redundant.  

Assuming this PR isn't going to affect cache performance, perhaps we want == to reflect the above logic and leave eql?() unchanged from Object#eql?(), and have it just compare to see if it's the same actual object?

apiology: If this PR does indeed affect cache performance, another approach would be to have a less aggressive equality_fields (maybe just #class and #identity?), and let your eql?() method above sort out any false hash() collisions.

apiology:
You'll want to remove this as well: https://github.com/castwide/solargraph/blob/master/lib/solargraph/pin/block.rb#L24-L27

lib/solargraph/pin/base.rb
castwide: > My assumption is that since we're now not overriding hash(), Object#hash() will return something fairly unique per object in memory, making eql()? above redundant.
> 
> Assuming this PR isn't going to affect cache performance, perhaps we want == to reflect the above logic and leave eql?() unchanged from Object#eql?(), and have it just compare to see if it's the same actual object?

Good call, thanks! Overriding `#==` and leaving `#eql?` alone is sufficient.

> If this PR does indeed affect cache performance, another approach would be to have a less aggressive equality_fields (maybe just #class and #identity?), and let your eql?() method above sort out any false hash() collisions.

Unfortunately, using `#identity` for the inference stack was what led to collisions in #922. Stacking the actual pins instead of their `#identity` resolved it, but that's what led to the performance issue. The root cause might be somewhere in the `Chain#infer` process instead of the cache itself. (Edit: On further reflection, the root cause might also be some other aspect of pin mapping instead of something directly related to chain inference).

PR #926, opened 2025-05-06
Enable users to provide RBS shims for their bundled gems

apiology:
Putting .rbs files in the sig/shims directory will allow users to add types for gems which do not provide them.

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -17,6 +17,7 @@ module Solargraph
     attr_reader :unresolved_requires
 
     @@core_map = RbsMap::CoreMap.new
+    @@shim_pins = {}
 
     # @return [Array<String>]
     attr_reader :missing_docs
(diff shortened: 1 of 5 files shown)


lib/solargraph/api_map.rb
```diff
@@ -17,6 +17,7 @@ class ApiMap
     attr_reader :unresolved_requires
 
     @@core_map = RbsMap::CoreMap.new
+    @@shim_pins = {}
```
apiology, line 21: If there's an invalidation technique better than "wait until user restarts solargraph" that I should use, or a better place to stash these, just let me know--I'm not as savvy on that side of Solargraph.

castwide:
Should this feature be implemented using the existing RBS collection support instead? The default path for external sigs is `.gem_rbs_collection`, but it can be configured in `rbs_collection.yaml`.

apiology:
(from `rbs_collection.yaml`):

```
# You can specify local directories as sources also.
# - type: local
#   path: path/to/your/local/repository
```

Neat!  I didn't know that existed.

PR #939, opened 2025-05-18
Add support for Ruby Structs

lekemula:
Hi,

This PR adds support for Ruby [Struct](https://ruby-doc.org/core-2.5.8/Struct.html) class definitions both via inheritance and const assignment:

```ruby
# @param bar [String]
# @param baz [Integer]
class Foo < Struct.new(:bar, :baz, keyword_init: true)
  def foo
  end
end

# @param bar [String]
# @param baz [Integer]
Foo = Struct.new(:bar, :baz) do
  def foo
  end
end
```

It completes:

- Struct attributes
- Struct initialise method with appropriate args or kwargs based on `keyword_init` option
- Struct instance variables
- Utilises `@param` yard tags on class definition to infer attribute types.
	- While this is not a YARD specification, I think it avoids quite some boilerplate that comes from `@!attribute` usage, which kind of defeats the purpose of `Struct` usage to avoid defining them by hand in the first place.
		```ruby
		class Foo < Struct.new(:bar, :baz)
		  # @!attribute [rw] bar
		  #   @return [String]
		  # @!attribute [rw] baz
		  #   @return [Integer]
		  def foo
		  end
		end
		```

### Approach

So I thought for quite a while, where this code could live, and I went back and forth many times between:

- Adding it to the `lib/solargraph/parser/parser_gem/node_processors`
	- However, `Struct.new` is not a language construct, ie, node per se.
- Adding it to the `lib/solargraph/convention` (direction I went with)
	-	However, the code here mainly needs to _hook into_ the parsing stage, rather than implement the `Solargraph::Convention` interface.

So, as we see, each approach has its upside/downside. I ultimately went with the second approach because while working on the solargraph-rspec plugin, I always felt a need for Solargraph to provide a way to hook into the parsing phase. Something that would prevent plugins from parsing the same code twice!

While I attempted to allow supporting [multiple node processers](https://github.com/castwide/solargraph/commit/6707d22c42951f7e081411356ce9ea4688391e03), I stumbled over some issues with having "duplicate pins", which I couldn't find a good solution to, so I ended up with this "hybrid solution", hoping that we get to solve that problem eventually (which is why I left the `TODO` comments are for)

I plan to add support similarly for [Data](https://docs.ruby-lang.org/en/3.2/Data.html) classes too, before jumping to that I wanted to got some feedback from you on the proposed approach and whether this any sense at all to you?

Thanks in advance for your thoughts and guidance!

### DIFF
--- a/lib/solargraph/parser/parser_gem/node_processors/casgn_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/casgn_node.rb
@@ -8,6 +8,17 @@ module Solargraph
           include ParserGem::NodeMethods
 
           def process
+            if Convention::StructDefinition::StructAssignmentNode.valid?(node)
+              process_struct_assignment
+            else
+              process_constant_assignment
+            end
+          end
+
+          private
+
+          # @return [void]
+          def process_constant_assignment
             pins.push Solargraph::Pin::Constant.new(
               location: get_node_location(node),
               closure: region.closure,
@@ -18,7 +29,14 @@ module Solargraph
             process_children
           end
 
-          private
+          def process_struct_assignment
+            processor_klass = Convention::StructDefinition::NodeProcessors::StructNode
+            processor = processor_klass.new(node, region, pins, locals)
+            processor.process
+
+            @pins = processor.pins
+            @locals = processor.locals
+          end
 
           # @return [String]
           def const_name
--- a/lib/solargraph/parser/parser_gem/node_processors/casgn_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/casgn_node.rb
@@ -29,6 +29,8 @@ module Solargraph
             process_children
           end
 
+          # TODO: Move this out of [CasgnNode] once [Solargraph::Parser::NodeProcessor] supports
+          # multiple processors.
           def process_struct_assignment
             processor_klass = Convention::StructDefinition::NodeProcessors::StructNode
             processor = processor_klass.new(node, region, pins, locals)
--- a/spec/source_map/clip_spec.rb
+++ b/spec/source_map/clip_spec.rb
@@ -2661,4 +2661,23 @@ describe Solargraph::SourceMap::Clip do
     names = clip.complete.pins.map(&:name)
     expect(names).to include('bar', 'bar=', 'baz', 'baz=')
   end
+
+  it 'completes Struct initialize keyword args' do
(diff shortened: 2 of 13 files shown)


spec/source_map/clip_spec.rb
```diff
@@ -2630,4 +2630,117 @@ def foo
     clip = api_map.clip_at('test.rb', [13, 12])
     expect(clip.infer.to_s).to eq('Integer')
   end
+
+  it 'completes Struct methods' do
```
lekemula, line 2634: Are clip_specs the right place for these specs?

lib/solargraph/parser/parser_gem/node_processors/casgn_node.rb
```diff
@@ -8,6 +8,17 @@ class CasgnNode < Parser::NodeProcessor::Base
           include ParserGem::NodeMethods
 
           def process
+            if Convention::StructDefinition::StructAssignmentNode.valid?(node)
```
lekemula, line 11: I don't like the fact that I had to change the node processor classes - do you see a better way to avoid that?

My initial attempt was to allow [multiple node processors](https://github.com/castwide/solargraph/commit/6707d22c42951f7e081411356ce9ea4688391e03#diff-b02ea47997df46c682060e9202b17771cc769e0a3fe5c2171f4769c9e4bc2b3fR33-R49), as I explained in the PR description.

castwide: I suspect that any solution would have required some sort of change to node processors. At the very least, they would need a mechanism to map anonymous class pins for declarations like `class Foo < Struct.new`. I'm good with handling it like this for now. We can try exploring more general solutions when we look into adding support for `Data`.

spec/source_map/clip_spec.rb
castwide: The `Clip` specs are pretty heavy, but I can understand why. They're fine here for now. I'm planning updates to `Chain` that make it less reliant on `Clip` features. Part of that change will probably include moving some tests from `clip_spec` to `chain_spec` or the various link specs.

castwide:
:+1: :shipit:

lib/solargraph/parser/parser_gem/node_processors/casgn_node.rb
lekemula: Sounds good 🙌!

PR #942, opened 2025-05-19
Re-enable support for .gem_rbs_collection directories

apiology:

PR #948, opened 2025-05-22
Support ActiveSupport::Concern pattern for class methods

apiology:


lib/solargraph/api_map.rb
```diff
@@ -643,6 +643,8 @@ def inner_get_methods rooted_tag, scope, visibility, deep, skip, no_core = false
       # namespaces; resolving the generics in the method pins is this
       # class' responsibility
       methods = store.get_methods(fqns, scope: scope, visibility: visibility).sort{ |a, b| a.name <=> b.name }
+      methods = methods.map(&:as_virtual_class_method) if store.get_includes(fqns).include?('ActiveSupport::Concern') && scope == :class
```
lekemula: Support for Concerns arguably falls under the category of a plugin extension, since it's not a native Ruby language feature. Can we somehow generalise the solutions for "conventions" (ie. plugins)? 

Having such extensibility capabilities for `Convention`s would unlock great potential of additional plugins, I believe.

That said, having _Concerns_ support in one way or another would be very beneficial! 🚀

apiology: Yeah, having it separated by a plugin interface of some type would be helpful for sure. 

I do suspect that plugin should be included in the solargraph install by default, given ActiveSupport::Concern is typically an opaque part of gems users bring in, and the plugin would be needed to provide users a full typed surface to work again for those gems.

This particular PR is not even at the proof of concept phase yet, just some half-finished code that I'll probably need soon to get more capability in solargraph-rails working.

apiology: I took a good run at introducing a new 'object()' method in the Convention::Base interface to return the methods off of an object, as well as a new yard_plugins field in Environ that can be populated by the 'global()' method.

I've got the diff in my stashes, but I'm afraid it's dependent on / overlaps with changes in these:
* #877 
* #946 
* #942 

This PR as-is is a huge win for Solargraph Rails, so rather than reimplement this thing three times, I'd suggest we merge this and then retrieve my stash once those three are merged.

lib/solargraph/convention.rb
```diff
@@ -6,12 +6,14 @@ module Solargraph
   # one of its sources.
   #
   module Convention
+
```
apiology:
```suggestion
```

lib/solargraph/api_map.rb
lekemula: @apiology I just saw the `Convention.for_object` hook, thank you so much for this! This will bring tremendous power to the plugins 🔥.

PR #962, opened 2025-06-05
Fix inline Struct definition

lekemula:
Fixes inline struct definition without `do..end` block:

```ruby
# @param bar [String]
# @param baz [Integer]
Foo = Struct.new(:bar, :baz)
```


castwide:
Thanks!

PR #967, opened 2025-06-13
Use undercover tool to report on PR test coverage

apiology:
This sets up a GitHub workflow which should fail if code added / modified in a PR isn't tested in a spec.

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -39,6 +39,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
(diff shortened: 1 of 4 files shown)


apiology:
This is mainly something I find helpful in prompting me to go back and add more comprehensive tests for changes.  We can [formally make it optional](https://stackoverflow.com/questions/78072432/how-to-enable-a-single-github-action-to-not-block-a-merge) or just rely on judgment on whether to merge PRs that aren't fully covered - I'm not too tied up on what that policy is, but having the info is helpful to me, at least.

apiology:
Superceded by #996

PR #970, opened 2025-06-15
Add support for Ruby Data.define

lekemula:
Follow up of https://github.com/castwide/solargraph/pull/939

Adds the same support for Data.define

```ruby
# @param bar [String]
# @param baz [Integer]
class Foo < Data.define(:bar, :baz)
  def foo
  end
end

# @param bar [String]
# @param baz [Integer]
Foo = Struct.define(:bar, :baz)
```

The PR consists of two parts:

1. Merely porting https://github.com/castwide/solargraph/pull/939 for Data.define (first commit)
2. Refactoring of node processors in order to decouple Data/Struct node processors (ie, "conventional processors") from normal Ruby language processors. (other commits)
	- Allow registering multiple node processors for the same node
	- Apply the chain-of-responsibility pattern to skip processing the next processors in the chain.

I'd recommend reviewing them commit-by-commit, in order to see better their effect. Let me know how this looks to you!


lekemula:
@apiology so nice to have the typecheck step in CI 🚀 

[`9dc925a` (#970)](https://github.com/castwide/solargraph/pull/970/commits/9dc925abbee383df5b570c6ca3364c5549610f54)

lib/solargraph/convention/data_definition/data_definition_node.rb
```diff
@@ -0,0 +1,100 @@
...
+        def keyword_init?
+          keyword_init_param = data_attribute_nodes.find do |data_def_param|
+            data_def_param.type == :hash && data_def_param.children[0].type == :pair &&
+              data_def_param.children[0].children[0].children[0] == :keyword_init
+          end
+
+          return false if keyword_init_param.nil?
+
+          keyword_init_param.children[0].children[1].type == :true
+        end
```
lekemula: I should have deleted this method.

lekemula:
Probably we should also port the changes from https://github.com/castwide/solargraph/pull/992 to the struct definition too, but I'd prefer to do that as a follow-up once that gets approved and merged first.

@castwide I rebased and resolved the conflicts here, let me know if you have anything that you'd like me to address.

castwide:
Looks good, thanks! Support for multiple processors per node should be useful elsewhere, too.

apiology:
@lekemula @castwide - it seems as though this creates a circular dependency between ::Convention and ::Parser that for whatever reason didn't fail the build until it was combined with the current code in master.  Perhaps we could revert this, merge master back in and work to resolve that?

lekemula:
> @apiology @lekemula I have a workaround that I'll push to a separate PR shortly.

TY, and sorry for the trouble!

PR #974, opened 2025-06-23
Use Prism

castwide:
Make `ParserGem` use Prism to parse whitequark-compatible ASTs.

Testing on the Rails repo, using Prism appears to reduce time to parse by about 60%.

### DIFF
--- a/lib/solargraph/parser/parser_gem/class_methods.rb
+++ b/lib/solargraph/parser/parser_gem/class_methods.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'parser/current'
-require 'parser/source/buffer'
 require 'prism'
 
 # Awaiting ability to use a version containing https://github.com/whitequark/parser/pull/1076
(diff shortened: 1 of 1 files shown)


lib/solargraph/parser/parser_gem/class_methods.rb
```diff
@@ -39,12 +38,10 @@ def parse code, filename = nil, line = 0
...
         # @return [::Parser::Base]
         def parser
-          # @todo Consider setting an instance variable. We might not need to
-          #   recreate the parser every time we use it.
-          parser = ::Parser::CurrentRuby.new(FlawedBuilder.new)
-          parser.diagnostics.all_errors_are_fatal = true
-          parser.diagnostics.ignore_warnings      = true
-          parser
+          @parser ||= Prism::Translation::Parser.new(FlawedBuilder.new).tap do |parser|
+            parser.diagnostics.all_errors_are_fatal = true
```
lekemula, line 42: What does this flag do?

Prism is known for its good error tolerance. Is this something we could leverage?

castwide: > What does this flag do?

It defers responsibility of handling errors to `Source`, which is capable of repairing certain types of syntax errors.

> Prism is known for its good error tolerance. Is this something we could leverage?

That's very possible. This iteration does the minimum work of staying compatible with the existing whitequark implementation while leveraging prism's speed. In the future, we might be able to rely more on prism instead of the `Source` repair mechanism.

ShadiestGoat:
This is great! Perhaps out of scope for this PR, but would it be possible to add an option to NOT use current ruby?

This can allow the use of solargraph on ruby < 3 projects, while running the server on ruby 3.x.

ShadiestGoat:
I'll open an issue for this discussion, but my 2¢ is that it's not up to solargraph to interpret any version file, or figuring out where gems are or anything like that. It's kind of an edge case anyways, so support can be limited.

My proposal is that the gems need to be installed on the version that is running solargraph, and that there's a key in the solargraph config to use a specific parser version

Issue #964, opened 2025-06-10
Workspace scan failing or stuck


oswaldoleal:
Context: I primarily  use the Ruby Solargraph extension inside VS Code, however over the last few days I notice the function definition links were not working and a pop up to update solarpgrahp came up, I updated but now, every time I try to load the project, the notification displaying the progress of mapping the workspace always gets stuck at 465/4915. 

![Image](https://github.com/user-attachments/assets/3b5d75d7-4c51-4c4d-b203-9345c5a66445)

So naturally I try to run in on console, with this command `solargraph scan -v`, but it fails right after running it.

This is the version output:
```
$ solargraph --version
0.55.1
```

And this is the full output for the scan command:
```
$ solargraph scan -v
/home/oswaldo/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.55.1/lib/solargraph/parser/flow_sensitive_typing.rb:167:in `parse_isa': undefined method `type' for nil (NoMethodError)

        if isa_receiver.type == :send && isa_receiver.children[0].nil? && isa_receiver.children[1].is_a?(Symbol)
                       ^^^^^
	from /home/oswaldo/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.55.1/lib/solargraph/parser/flow_sensitive_typing.rb:185:in `process_isa'
	from /home/oswaldo/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.55.1/lib/solargraph/parser/flow_sensitive_typing.rb:146:in `process_conditional'
	from /home/oswaldo/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.55.1/lib/solargraph/parser/flow_sensitive_typing.rb:61:in `process_if'
	from /home/oswaldo/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.55.1/lib/solargraph/parser/parser_gem/node_processors/if_node.rb:15:in `process'
	from /home/oswaldo/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.55.1/lib/solargraph/parser/node_processor.rb:39:in `process'
[...]

PR #975, opened 2025-06-24
Nil guards in flow-sensitive typing

castwide:
ref #964

`FlowSensitiveTyping#parse_isa` throws errors when the node being processed or its receiver are nil.

In addition to adding the nil guards, this PR moves some related tests from `clip_spec.rb` to `flow_sensitive_typing_spec.rb`. The ultimate goal is to make `Clip` less of a god class (at least as far as tests are concerned) and move some of its tests closer to the specific feature being tested. `clip_spec.rb` is currently the most common cause of merge conflicts.


spec/parser/flow_sensitive_typing_spec.rb
```diff
@@ -0,0 +1,244 @@
+# frozen_string_literal: true
```
apiology, line 1: Thanks for breaking this out!

PR #987, opened 2025-07-01
Ignore directory paths in Workspace#would_require?

castwide:
`Bench` uses `Workspace#would_require?` to determine if a require path is local. Ensure that the `#would_require?` method ignores directory paths.


castwide:
Closing to fix deployment workflow.

PR #992, opened 2025-07-03
Improve struct support

ShadiestGoat:
Please let me know if there are issues!

- Preserve multi-line descriptions of params
- If the param description has the attribute name in it, it used to get removed
- Better support for all the different variations in documenting the struct params - `@return` on the param line, `@param` on the struct line `@param` above the param line, etc


castwide:
:+1: :shipit:

PR #993, opened 2025-07-04
[Bug fix] Fixed an error in rbs/fills/tuple.rbs

TD12734:
When using Solargraph version 0.55.4 or later with vscode on Windows, I get the following error:

error

If this error occurs, various functions of solargraph will not work.

The cause of the error is that when the rbs gem is a relatively old version (for example, 3.4.0), an error occurs in the syntax of `rbs/fills/tuple.rbs`.

To avoid this, we have made it mandatory to also install the rbs gem of version 3.6.1 or higher when installing the solargraph gem.

The value 3.6.1 is the version in which the error no longer occurred in my environment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,6 @@ source 'https://rubygems.org'
 
 gemspec name: 'solargraph'
 
-gem 'rbs', '>= 3.9.4'
-
 # Local gemfile for development tools, etc.
 local_gemfile = File.expand_path(".Gemfile", __dir__)
 instance_eval File.read local_gemfile if File.exist? local_gemfile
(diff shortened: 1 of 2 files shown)


apiology:
Hmmm, unfortunately requiring a newer rbs gem will eliminate support for Ruby 3.0.

We could probably work around it in some way so that LSP support in Ruby 3.0 can work again.

TD12734:
I have also confirmed that solargraph works with rbs3.6.1.

apiology:
Excellent!  Maybe our Ruby 3.0 CI tests worked because it was using a new version - was confused on that point.

apiology:
Yay--I was able to reproduce the issue here: https://github.com/castwide/solargraph/actions/runs/16091976403/job/45409964842?pr=995

castwide:
Thanks! This is likely to be superceded by #995, but I'm happy to iterate through changes before the next release.

PR #994, opened 2025-07-04
Add JSON Schema

ShadiestGoat:
Adds a JSON Schema for the config file, including docs for each field

Note: yaml lang server had a regression and can't display descriptions for consts/enums in arrays on latest stable version. I have a PR with a potential fix, but we shall see how that goes.

Preview:

image

Docs were mostly pulled from the solargraph site

Only thing unclear to me, what is the "update_errors" reporter?

### DIFF
--- a/solargraph.schema.json
+++ b/solargraph.schema.json
@@ -1,0 +1,86 @@
+{
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "properties": {
+        "include": {
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "A list of directory globs to include in Solargraph's code maps. Exclude takes priority over include.",
+            "default": ["**/*.rb"]
+        },
+        "exclude": {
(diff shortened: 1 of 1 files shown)


ShadiestGoat:
@castwide I finished the schema with the additional info

castwide:
I definitely like the idea of validating the schema. One concern I have, specifically with reporters, is that extensions can add their own reporters. If I'm using solargraph-reek, is there a way to make 'reek' appear as an available reporter?

ShadiestGoat:
Youre right - afaik, its not possible to actually like "register" extra schema modules. I'll add a fallback value.

In addition to this, what do you think about pre-registering values for well known modules, like reek/rails/rspec?

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.2.0
```
apiology, line 1: Mind opening a separate PR with an explanation if you feel this should be part of the repo?

PR #995, opened 2025-07-05
Allow newer RBS gem versions

apiology:
This allow users to upgrade to recent Tapioca versions.

Tapioca now requires newish versions of the spoom gem, which depends on 4.x pre-releases of the rbs gem.


jeffwidman:
Looking forward to when a release of `solargraph` includes this PR. 😄

PR #997, opened 2025-07-09
Add some RBS shims to help handle strict-level typechecking issues

apiology:


solargraph.gemspec
```diff
@@ -33,6 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'logger', '~> 1.6'
   s.add_runtime_dependency 'observer', '~> 0.1'
   s.add_runtime_dependency 'ostruct', '~> 0.6'
+  s.add_runtime_dependency 'open3', '~> 0.2.1'
```
apiology, line 37: These are part of stdlib, but rbs collection wants to see an explicit dependency in order to recognize that it should look for shims in my configured source for them.

PR #998, opened 2025-07-11
Combine global pins from plugins

apiology:
Plugin pins aren't always respected - they don't get run through GemPins for combination, leading to the pin pulled not necessarily being from the plugin if there is a pin from another source.


apiology:
Pretty sure this is a false alarm - will reopen if I'm wrong!

Issue #991, opened 2025-07-02
`solargraph cache` subprocess repeats forever


gongfarmer:
Seen on version 0.56 today.  This is not issue #976 -- I had that one too but it is fixed for me in 0.56.

My editor (nvim) starts a solargraph language server with `solargraph stdio`.
The server process continually forks a sub-process with the command-line `solargraph cache ostruct 0.6.2`.
This process seems to last for only a few seconds.  Then the parent process starts it over again.

```
$ ps aux | grep solargrap[h]
fhanson   389347 12.8  3.1 1294016 1044040 ?     Ssl  12:25   1:17 /usr/bin/ruby3.1 /data/home/fhanson/.local/share/nvim/mason/packages/solargraph/bin/solargraph stdio
fhanson   391103  100  0.3 166072 106084 ?       R    12:35   0:01 /usr/bin/ruby3.1 /data/home/fhanson/.local/share/nvim/mason/packages/solargraph/bin/solargraph cache ostruct 0.6.2

```
In the `ps` output above, the first process is the language server, it stays running as long as my editor is open.
The second process is the one that continually repeats.  The PID changes every few seconds but the command-line remains the same.

I tried manually running `~/.local/share/nvim/mason/bin/solargraph cache ostruct 0.6.2` with the same working directory as my editor.
The command completes in a few seconds with 0 exit code.

Within the editor, this shows up as constantly repeated notifications  that the background scan is running.

Strangely, the command is `cache ostruct 0.6.2` but that version of the ostruct gem is not installed on my system.  According to `gem list ostruct` I have 0.6.0 and 0.5.2.

castwide:
I found the cause of the cache loop. `Library` wasn't handling errors from spawned processes. #1000 should resolve it.

gongfarmer:
Thanks, I will try it out as soon as there's a new release

castwide:
Fix released in 0.56.2.

PR #1000, opened 2025-07-13
Use Open3 to cache gemspecs

castwide:
ref #991 

`Library` used `Process.spawn` to run gem cache processes. Use `Open3` so we can handle errors properly instead of retrying in an endless loop.


lekemula:
Congrats on the 1000th PR @castwide! 🎉

PR #1005, opened 2025-07-14
Add RBS fill for Bundler::Dsl

apiology:
Pending types for this class being added to the RBS repo and then to the RBS version the user has loaded, this will allow Gemfiles to be typechecked.


lib/solargraph/pin_cache.rb
```diff
@@ -165,6 +167,8 @@ def uncache *path_segments, out: nil
         if File.exist?(path)
           FileUtils.rm_rf path, secure: true
           out.puts "Clearing pin cache in #{path}" unless out.nil?
+        else
+          out.puts "Pin cache file #{path} does not exist" unless out.nil?
```
apiology: Helpful when debugging the core RBS fills!

lib/solargraph/rbs_map/core_map.rb
```diff
@@ -5,12 +5,13 @@ class RbsMap
...
     #
     class CoreMap
+      include Logging
 
       def resolved?
         true
       end
 
-      FILLS_DIRECTORY = File.join(File.dirname(__FILE__), '..', '..', '..', 'rbs', 'fills')
+      FILLS_DIRECTORY = File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..', 'rbs', 'fills'))
```
apiology, line 14: Nicer to see an absolute path in log messages

lib/solargraph/rbs_map/core_map.rb
```diff
@@ -33,10 +41,6 @@ def pins
         @pins
       end
 
-      def loader
-        @loader ||= RBS::EnvironmentLoader.new(repository: RBS::Repository.new(no_stdlib: false))
-      end
-
```
apiology, line 40: Whoops, we ended up with two identical methods here

lib/solargraph/rbs_map/core_map.rb
```diff
@@ -22,8 +23,15 @@ def pins
...
-          @pins = conversions.pins
+          Dir.glob(File.join(FILLS_DIRECTORY, '*')).each do |path|
+            next unless File.directory?(path)
+            fill_loader = RBS::EnvironmentLoader.new(repository: RBS::Repository.new(no_stdlib: false))
+            fill_loader.add(path: Pathname(path))
+            fill_conversions = Conversions.new(loader: fill_loader)
+            @pins.concat fill_conversions.pins
+          rescue RBS::DuplicatedDeclarationError => e
+            logger.debug "RBS already contains declarations in #{path}, skipping: #{e.message}"
+          end
```
apiology: Unfortunate foot-gun: RBS fails the entire load if a single declaration is duplicated with an existing declaration.  Let's use water-tight compartments here, anticipating that outside fills like this will be integrated upstream with time, so that we don't lose the other fills when they are.

apiology: Ah, never mind - this only happens if you load in the same loader as core pins, which I hadn't turned off here.

lib/solargraph/type_checker.rb
```diff
@@ -147,33 +147,33 @@ def virtual_pin? pin
     # @return [Array<Problem>]
     def method_param_type_problems_for pin
       stack = api_map.get_method_stack(pin.namespace, pin.name, scope: pin.scope)
```
apiology, line 154: The type checker changes were needed to make the test in gemfile_spec work.  It turns out that the type checker was checking parameter types using the result of Pin::Method#parameters, which for RBS-loaded types is always an empty array.

I took the opportunity to refactor the 'first_param_hash' method into 'param_details_from_stack', which takes both the method pins from the stack as well as the particular signature we're testing against, which I think is more compatible with how it is used.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -476,7 +476,15 @@ def parts_of_function type, pin
...
-          parameters.push Solargraph::Pin::Parameter.new(decl: :restarg, name: name, closure: pin, source: :rbs, type_location: type_location)
+          inner_rest_positional_type =
+            ComplexType.try_parse(other_type_to_tag(type.type.rest_positionals.type))
+          rest_positional_type = ComplexType::UniqueType.new('Array',
+                                                             [],
+                                                             [inner_rest_positional_type],
+                                                             rooted: true, parameters_type: :list)
+          parameters.push Solargraph::Pin::Parameter.new(decl: :restarg, name: name, closure: pin,
+                                                         source: :rbs, type_location: type_location,
+                                                         return_type: rest_positional_type,)
```
apiology, line 489: I got into a bit of a yakshave here trying to get a test to pass, and found out that we don't preserve types for restargs from RBS.

lib/solargraph/rbs_map/core_map.rb
```diff
@@ -22,9 +23,16 @@ def pins
...
-          loader.add(path: Pathname(FILLS_DIRECTORY))
-          @pins = conversions.pins
+          @pins.concat conversions.pins
+
+          # Avoid RBS::DuplicatedDeclarationError by loading in a different EnvironmentLoader
+          fill_loader = RBS::EnvironmentLoader.new(core_root: nil, repository: RBS::Repository.new(no_stdlib: false))
+          fill_loader.add(path: Pathname(FILLS_DIRECTORY))
+          fill_conversions = Conversions.new(loader: fill_loader)
+          @pins.concat fill_conversions.pins
+
```
apiology, line 34: A single EnvironmentLoader complains if it's given the same symbol twice - which could happen if one of our temporary fills like Bundler::Dsl is eventually adopted upstream (see https://github.com/ruby/rbs/issues/2599).  We'll get around that by using a separate loader for the fills, and deal with combining the duplicate pins downstream.

spec/convention/gemfile_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        local_gemfile = File.expand_path(".Gemfile", __dir__)
+        instance_eval File.read local_gemfile if File.exist? local_gemfile
+      ))
+
+      expect(checker.problems.map(&:message).sort).
+        to eq(["Unrecognized keyword argument bad_name to Bundler::Dsl#gemspec",
+               "Wrong argument type for Bundler::Dsl#source: source expected String, received Class<File>"].sort)
+    end
+
+    # @todo add rest arg support to type checker
```
apiology: I cut the yakshave off here so the changes to type checker weren't more aggressive than they already are.

PR #1010, opened 2025-07-18
Plugin Util: Add Combination Priority

ShadiestGoat:
While working on adding factory bot support into the rspec plugin, I noticed that I couldn't override existing documentation within factory bot itself. There is the override ref pin, but that doesn't quite let you override an entire method. This would allow it to

This is backwards compatible & only affects plugins, as I cannot see why this would be useful from a writing code/docs perspective

### DIFF
--- a/lib/solargraph/pin/base.rb
+++ b/lib/solargraph/pin/base.rb
@@ -28,6 +28,11 @@ module Solargraph
       # @return [::Symbol]
       attr_accessor :source
 
+      # @type [::Numeric, nil] A priority for determining if pins should be combined or not
+      #   A nil priority is considered the be the lowest. All code, yard & rbs pins have nil priority
+      #   Between 2 pins, the one with the higher priority gets chosen. If the priorities are equal, they are combined.
+      attr_reader :combine_priority
+
       def presence_certain?
         true
       end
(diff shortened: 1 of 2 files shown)


ShadiestGoat:
Hi! Sorry for the long wait, let me provide an example. For the rspec plugin, I'm trying to override the `#create` method that is defined by a `@!method` directive in `FactoryBot::Syntax:Methods`. When converted to RBS, the method has this description:

```
def create: (name untyped, *traits_and_overrides ::Aray, block ::Proc) { () -> } -> Object
```

I'm trying to override it with something like this:

```
def create: (name :person, *traits :some_trait, ?first_name: untyped, ?last_name: String, ?email: untyped) -> Person
  | create(name: :post, *traits (:with_tags, :with_admin_author, :secure), ?title: String, ?body: String) -> Post
  | create(name :tag, *traits (:some_config, :some_other), ?name: String, ?created_at: untyped) -> Tag
```

(note these are not exact RBS definitions I was using, but a close enough approximation to it. Also these are hand written so mistakes blah blah blah)

Anyways, so combined result of this I was getting had the yard defined sig at the top, which I don't want as it *might* interfere with type inference. The real problem though, is that I was getting untyped on all the `name` args in every signature.

apiology:
Thanks for the write-up!  Definitely agree we ought to prioritize pins from plugins, but it's helpful to see the example for the future - there's probably improvement to be done in prioritizing specific signatures over splat arguments, for instance

PR #1024, opened 2025-08-04
Improve a pin combination case around selfy types

apiology:
An attempt at fixing return types that should have been 'self' during pin combination wasn't always working, due to Pin::Base#context sometimes returning Class<Foo>.


spec/pin/combine_with_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(combined.return_type.to_s).to eq('self')
+    end
+
+    it 'combines a valid return type with a dodgy one' do
+      combined = normal_pin.combine_with(dodgy_location_pin)
+      expect(combined.return_type.to_s).to eq('self')
+    end
+  end
+
+  context 'with return types that should probably be self' do
```
apiology, line 43: This exercises the new code

PR #1028, opened 2025-08-10
Fix Convention/Plugin pins not being updated on file change

lekemula:
This PR fixes an issue where [RSpec plugin](https://github.com/lekemula/solargraph-rspec) (or any plugin for that matter) pins weren't being updated after file saves/changes due to a subtle reference sharing bug in index store.

### Problem:
  ~~The implicit.pins array was being shared between `ApiMap` and `Index` objects. Since `implicit` environ (now named `conventions_environ`) is memoized, when `Store#update(*pinsets)` compared pin sets to detect changes, were using the same `implicit.pins` array, thus changes were not detected, since those arrays are always identical:~~
  https://github.com/castwide/solargraph/blob/2bf80f04f53f40a5e818640c0e95eb6e7fe7fc50/lib/solargraph/api_map/store.rb#L29

UPDATE: The above is true; however, solving it by updating all convention pins every time we change turned out to be a bad idea in terms of performance time (obviously).

### Solution:
  
~~Added .dup to ensure each object gets its own copy of the pins array, allowing the change detection to work correctly.~~

~~I have to give credit to @claude, coz this was taking me forever to figure out why the two arrays were always the same! 🤯~~

Include the convention pins in `live_pins` when updating the store.

PS. I hope you don't mind the `implicit` to `conventions` rename. That was adding cognitive load to reason about those pins, and I hope this makes it more explicit that these are pins related to plugins/conventions.

### Testing:
  - [RSpec plugin](https://github.com/lekemula/solargraph-rspec) pins (ie. spec file changes) now properly update after file modifications
  - No regression in existing functionality

### DIFF
--- a/lib/solargraph/source_map.rb
+++ b/lib/solargraph/source_map.rb
@@ -30,9 +30,12 @@ module Solargraph
     def initialize source
       @source = source
 
-      environ.merge Convention.for_local(self) unless filename.nil?
-      self.convention_pins = environ.pins
-      # @type [Hash{Class<generic<T>> => Array<generic<T>>}]
+      conventions_environ.merge Convention.for_local(self) unless filename.nil?
+      # FIXME: unmemoizing the document_symbols in case it was called and memoized from any of conventions above
+      # this is to ensure that the convention_pins from all conventions are used in the document_symbols.
+      # solargraph-rails is known to use this method to get the document symbols. It should probably be removed.
+      @document_symbols = nil
+      self.convention_pins = conventions_environ.pins
       @pin_select_cache = {}
     end
 
@@ -69,8 +72,8 @@ module Solargraph
     end
 
     # @return [Environ]
-    def environ
-      @environ ||= Environ.new
+    def conventions_environ
+      @conventions_environ ||= Environ.new
     end
 
     # all pins except Solargraph::Pin::Reference::Reference
@@ -178,8 +181,6 @@ module Solargraph
     # @param pins [Array<Pin::Base>]
     # @return [Array<Pin::Base>]
     def convention_pins=(pins)
-      # unmemoizing the document_symbols in case it was called from any of conventions
-      @document_symbols = nil
       @convention_pins = pins
     end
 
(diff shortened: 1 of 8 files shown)


lib/solargraph/source_map.rb
```diff
@@ -30,8 +30,12 @@ def locals
     def initialize source
       @source = source
 
-      environ.merge Convention.for_local(self) unless filename.nil?
-      self.convention_pins = environ.pins
+      conventions_environ.merge Convention.for_local(self) unless filename.nil?
+      # FIXME: unmemoizing the document_symbols in case it was called and memoized from any of conventions above
+      # this is to ensure that the convention_pins from all conventions are used in the document_symbols.
+      # solargraph-rails is known to use this method to get the document symbols. It should probably be removed.
+      @document_symbols = nil
```
lekemula, line 42: I just moved it closer to the problem - specifically, the `Convention.for_local(self)` call where the `solargraph-rails` method call causes a conflict by memoizing `document_symbols`, which results in pins from `solargraph-rspec` not being included there.

https://github.com/search?q=repo%3Aiftheshoefritz%2Fsolargraph-rails%20document_symbols&type=code

@apiology heads-up: I will probably open a PR soon, replacing it with `#pins` call and inlining the logic from `document_symbols`. Let me know if this does not make sense for some reason. 

Or is it a good idea to keep this as is? I guess this would avoid accidental problems from any future plugin.. 💭

lekemula:
@castwide, after testing this in a real-life Rails project, this turns out to be highly inefficient. Converting it to a draft and looking for a better solution. 👀 

Let me know if you have any suggestions!

lekemula:
> @castwide, after testing this in a real-life Rails project, this turns out to be highly inefficient. Converting it to a draft and looking for a better solution. 👀
> 
> Let me know if you have any suggestions!

Found a better solution 🚀  [`14f7e90` (#1028)](https://github.com/castwide/solargraph/pull/1028/commits/14f7e90dc110e263c757152f78aa18d621919d02)

(Damn @claude, you disappointed me this time, but I won't give up on you yet!).

Kidding aside, the earlier finding still perplexes me and leaves me with one open question:

When are the `convention_pins` meant to be updated on the index store?

castwide:
:+1: :shipit:

PR #1029, opened 2025-08-10
Resolve class aliases via Constant pins

apiology:
This also eliminates the need for Parser::NodeMethods as a separately defined class.


castwide:
:100: :shipit:

lekemula:
@castwide @apiology I've been testing the `main` branch today, and surprisingly even after https://github.com/castwide/solargraph/pull/1035 I noticed a huge performance degradation.  :(

I was able to bisect the behaviour to this change. I'm afraid we need to revisit the performance implications of these changes:

### With this PR

image_1756125511579_0

### Without this PR

image

apiology:
Oh, nice!!  Really appreciate the walk-through!

castwide:
Could this be a caching issue? These are the benchmarks I get from running rspec on master and 728a798 (the commit before this PR):

```
master:
- 6 minutes 21 seconds (no cache)
- 1 minute 46.44 seconds (cached)
728a798:
- 6 minutes 30 seconds (no cache)
- 2 minutes 13.3 seconds (cached)
master:
- 1 minute 52.73 seconds (stale cache from 728a798)
- 6 minutes 14 seconds (no cache)
- 1 minute 48.97 seconds (cached)
```

I haven't noticed a performance hit running the language server on master yet, but I haven't really hammered it, either. I'll try profiling and see what it reports.

lekemula:
@castwide I retested it again on neovim (on a freshly restarted PC) and unfortunately I can confirm the issue is still there:

First go-to-definition (5e5938aad7598820867077c4fd72c3bd1deb45cc):

<details><summary>Details</summary>
image
</details> 


Second go-to-definition (on a different file - 5e5938aad7598820867077c4fd72c3bd1deb45cc):

<details><summary>Details</summary>
image
</details> 

I did a quick check on the VS Code, too. There, the performance appears to be "less impacted", but I suspect that comes as a result of the IDE making a `textDocument/documentSymbol` when a file is opened (in contrast with my Neovim setup), which leads to the cache being warmed up by the time you click on a method/class. Disabling it via `"solargraph.symbols": false` config, leads to the same latencies:

See recordings: [before revert](https://www.dropbox.com/scl/fi/kf2g3h5wn58otvg16gjr5/solargraph_perfomance_test.mp4?rlkey=fzq4v2vtbpdp3r5b5p21lpcju&st=njlp2iym&dl=0) 5e5938aad7598820867077c4fd72c3bd1deb45cc / [after revert](https://www.dropbox.com/scl/fi/v4pmxabytuebolnzqpba4/solargraph_after_revert_test.mp4?rlkey=gmh3kj6n9e5v0c7fxn2jtiaiw&st=56j3xjxa&dl=0) 449621fa35ce3dd51b88a4a991167c22c83a3dcf. 

They seem to match the timings of initial request from vernier profiles [linked earlier](https://github.com/castwide/solargraph/pull/1029#issuecomment-3220124119).

So yeah, caching definitely has a play here, but I think we should focus on the non-cached responses. Actually, it would be nice if we had the ability to disable it altogether via config for such performance testing.

Additionally, from what I have observed, the issue is more apparent in nested method calls (with a long chain) than in class definitions, which I found somewhat strange, given that this PR is about _resolving class aliases._

Please let me know if you need any additional information.

castwide:
Thanks, @lekemula! Those recordings were helpful. I'll proceed with the revert and open a draft PR to explore causes and solutions.

PR #1031, opened 2025-08-18
[regression] Fix crash while typechecking files with Struct use

apiology:
It looks like a mis-set closure field on a Parameter in is causing exceptions while typechecking code with Struct use inside:

https://github.com/iftheshoefritz/solargraph-rails/actions/runs/17040847020/job/48303963447?pr=158


castwide:
:+1:

PR #1033, opened 2025-08-19
Remove yard reference from gemfile

castwide:
@apiology If you need to use your fork of the yard gem for development, you can create a `.Gemfile` (note the dot) in the project root. Bundler will load it and Git will ignore it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,9 +2,6 @@ source 'https://rubygems.org'
 
 gemspec name: 'solargraph'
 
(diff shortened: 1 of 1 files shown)


apiology:
Got it - this one was to help local linting of Solargraph code until [that PR](https://github.com/lsegal/yard/pull/1622) gets merged.  We can definitely add false violations to the RuboCop todo file instead until then instead to keep the Gemfile less confusing, though.

PR #1035, opened 2025-08-21
Speed-up LSP completion response times

lekemula:
> [!WARNING]
>  DISCLAIMER: @claude was used to generate the code of the new methods, but mostly because I knew what I was looking for, and I was too lazy to figure out the methods needed to be called to make it work.

### Problem

In my current work project (with 7,140 Ruby files with  109,664  total lines), I've noticed that initial or cold LSP completions were taking an extremely long time. By **initial or cold completions**, I mean:

1. First requests after LSP boot (> ~40s)
2. First requests after opening a new file (~10-20s)
3. First requests after editing an open file (~10-20s)

The subsequent completions were fine up until the next edit, I'm assuming due to the caching ( less than typically ~0.5-1s)

```yaml
# .solargraph.yml
---
include:
  - "**/*.rb"
exclude:
  - "**/spec/**/*"
  - test/**/*
  - vendor/**/*
  - ".bundle/**/*"
domains: []
plugins:
  - solargraph-rspec
  - solargraph-rails
rspec:
  let_methods:
    - let_it_be
reporters:
  - rubocop
  - require_not_found
formatter:
  rubocop:
    cops: all
max_files: 20000

```

Coming from the https://github.com/castwide/solargraph/pull/1028, I initially thought the issue might be related to the indexing store update. However after using the excellent profiler, [vernier](https://github.com/jhawthorn/vernier), I came to find that the bottleneck seemed to be the `Solargraph::ApiMap#resolve_method_aliases` particularly its recursive usage of `#get_method_stack` inside `Solargraph::ApiMap#resolve_method_alias`

https://github.com/castwide/solargraph/blob/91bdc3244c4ad8762a89b6b37d144b7fe7d2fa0b/lib/solargraph/api_map.rb#L970

<details><summary>Vernier screenshot</summary>
image
</details> 

---------------

### Solution

Replace recursive `get_method_stack` and linear search with direct indexed method lookup via `Store` and `Index`.

- ~~Add indexed lookups for methods and aliases by name~~
	- 	~~Assume that the method names are relatively distinct enough to reduce the search space~~ -> [turned out redundant.](https://github.com/castwide/solargraph/pull/1035/commits/49fa70429048761a6a22aae341773c01af5f61d2)
- Reuse Cached ancestor traversal to avoid repeated computations via `get_method_stack`
- Add fast path for creating resolved alias pins without individual lookups

#### Before

[Vernier Profile](https://www.dropbox.com/scl/fi/hx7h4cl3z8huy476c2lsb/before_solargraph_verier_profile_20250821202548.json.gz?rlkey=oak0rdlxfzsyht4zb0oxgh6v5&st=a6mq9e37&dl=0)

<details><summary>Screenshots</summary>

#### First request after LSP boot

 - Total Go-to-definition completion time (~47s)

image

- `Solargraph::Library#definitions_at` time (~4.3s)

image

> [!NOTE]
> As we can see, most of the time is spent on cataloging initially, which makes sense. 

---------------

#### First request after opening a new file

- Total Go-to-definition completion time (~9.6s)

image

-  `Solargraph::Library#definitions_at` time (~7.9s)

image

</details> 

#### After

[Vernier Profile](https://www.dropbox.com/scl/fi/fqiyclwnsgcxqvixmksoe/after_solargraph_verier_profile_20250821195536.json.gz?rlkey=ng4dcyu02h9f71qazpbl5xlac&st=f1l5pcqh&dl=0)

<details><summary>Screenshots</summary>

#### First request after LSP boot

 - Total Go-to-definition completion time (~13.8s)

image

- `Solargraph::Library#definitions_at` time (0.92s)

image

> [!NOTE]
> As we can see, most of the time is spent on cataloging initially, which makes sense. 

---------------

#### First request after opening a new file

- Total Go-to-definition completion time (4.6s)

image

-  `Solargraph::Library#definitions_at` time (0.6s)

image

</details> 

------

### Summary

Performance improvements (measured on a large Ruby project with ~7,100 files):

- First request after LSP boot: ⏱️ 47s → 13.8s (~3.6x faster)

- First request after opening a new file: ⏱️ 9.6s → 4.6s (~2x faster)

- `Solargraph::Library#definitions_at` hot path: ⏱️ 4.3s → 0.9s (~4.6x faster)

Cached responses seem not to be impacted by this change, and they remain fast as before.

### Further potential Improvements

Example completion after file open: 

image

From the samples above (with the new implementation), cataloging accounts for roughly 75-85% of the total time in cold completion responses. There may be room for improvement here - for example, by making cataloging run asynchronously. I recall some recent changes in this area, so the implications aren’t entirely clear. That said, my hunch is that for code completion specifically, developer experience often benefits more from speed than from perfect accuracy (just my personal opinion).

### [Vernier](https://github.com/jhawthorn/vernier) quick how-tos

Assuming that you've already installed the gem locally, to run the solargraph with vernier use:

```
vernier run -- solargraph stdio
```

Use https://vernier.prof/ to upload the profiles above. 

If Vernier is a new tool for you, [this video stream](https://www.youtube.com/watch?v=9nvX3OHykGQ#t=27m40) from Aaron is highly recommended!

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -77,7 +77,6 @@ Layout/ClosingHeredocIndentation:
 # Configuration parameters: AllowForAlignment.
 Layout/CommentIndentation:
   Exclude:
-    - 'lib/solargraph/api_map.rb'
     - 'lib/solargraph/language_server/host.rb'
     - 'lib/solargraph/parser/parser_gem/node_methods.rb'
     - 'lib/solargraph/source_map/mapper.rb'
@@ -2136,13 +2135,6 @@ Style/NegatedIfElseCondition:
     - 'lib/solargraph/type_checker.rb'
 
 # This cop supports safe autocorrection (--autocorrect).
-# Configuration parameters: AllowedMethods.
-# AllowedMethods: be, be_a, be_an, be_between, be_falsey, be_kind_of, be_instance_of, be_truthy, be_within, eq, eql, end_with, include, match, raise_error, respond_to, start_with
-Style/NestedParenthesizedCalls:
-  Exclude:
-    - 'lib/solargraph/type_checker.rb'
-
-# This cop supports safe autocorrection (--autocorrect).
 Style/NestedTernaryOperator:
   Exclude:
     - 'lib/solargraph/pin/conversions.rb'
@@ -2237,7 +2229,6 @@ Style/RedundantBegin:
     - 'lib/solargraph/shell.rb'
     - 'lib/solargraph/source/cursor.rb'
     - 'lib/solargraph/source/encoding_fixes.rb'
-    - 'lib/solargraph/type_checker.rb'
     - 'lib/solargraph/workspace.rb'
 
 # This cop supports safe autocorrection (--autocorrect).
@@ -2674,7 +2665,6 @@ YARD/MismatchName:
     - 'lib/solargraph/pin/until.rb'
     - 'lib/solargraph/pin/while.rb'
     - 'lib/solargraph/pin_cache.rb'
-    - 'lib/solargraph/source/chain.rb'
     - 'lib/solargraph/source/chain/call.rb'
     - 'lib/solargraph/source/chain/z_super.rb'
     - 'lib/solargraph/type_checker.rb'
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -675,13 +675,53 @@ module Solargraph
     # @param visibility [Enumerable<Symbol>]
     # @return [Array<Pin::Base>]
     def resolve_method_aliases pins, visibility = [:public, :private, :protected]
-      with_resolved_aliases = pins.map do |pin|
-        resolved = resolve_method_alias(pin)
-        next nil if resolved.respond_to?(:visibility) && !visibility.include?(resolved.visibility)
-        resolved
-      end.compact
-      logger.debug { "ApiMap#resolve_method_aliases(pins=#{pins.map(&:name)}, visibility=#{visibility}) => #{with_resolved_aliases.map(&:name)}" }
-      GemPins.combine_method_pins_by_path(with_resolved_aliases)
+      # Separate regular pins from aliases
+      regular_pins = []
+      alias_pins = []
+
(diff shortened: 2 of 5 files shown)


lib/solargraph/source_map/clip.rb
```diff
@@ -12,7 +12,7 @@ def initialize api_map, cursor
         @api_map = api_map
         @cursor = cursor
         block_pin = block
-        block_pin.rebind(api_map) if block_pin.is_a?(Pin::Block) && !Solargraph::Range.from_node(block_pin.receiver).contain?(cursor.range.start)
+        block_pin.rebind(api_map) if block_pin.is_a?(Pin::Block) && !Solargraph::Range.from_node(block_pin.receiver)&.contain?(cursor.range.start)
```
lekemula: I think this was accidentally included from my other branch 🙈, I'll remove it.

apiology:
I'd suggest creating a spec demonstrating creating an alias in a subclass to a method defined in the superclass - that'd show that if you can't find the method the easy way, you fall back to the old way

apiology:
> /home/runner/work/solargraph/solargraph/lib/solargraph/api_map/store.rb:55 - Not enough arguments to Kernel#select

Probably another good spec to add - these select methods are on arrays, which implement Enumerable.  Enumerable#select is an alias for Enumerable#find_all, and I'm guessing the logic to find that alias and use it isn't working right with this change.

castwide:
Nice work, thanks! Out of curiosity, I tested this branch against a project that had slow completions due to heavy use of dynamic yield receivers. Performance was noticeably improved there, too.

The linting error should go away with the merge. If it doesn't, I'll deal with the fallout.

castwide:
From what I've determined, there's nothing to be gained from making `Library#sync_catalog` asynchronous. Ruby threads are always executed in a single process, so it will never run concurrently at the OS level (e.g., POSIX). Any other function within the solargraph process that needs the newest maps will need to wait for the library to synchronize anyway. Ruby threads don't make it any faster. The main reason `#sync_catalog` exists is to ensure that clients have access to the most current versions of code maps at the moment they're requested.

PR #1038, opened 2025-08-24
Fix RuboCop linting errors in regular expressions

castwide:


castwide:
Let's ignore rubocop_todo errors for now. In its current state, it's too much of a distraction.

castwide:
Undercover, too. I'd rather address these issues specifically instead of fighting them in unrelated PRs.

PR #1040, opened 2025-08-27
Avoid stack errors when resolving method aliases

castwide:


lekemula:
Well this popped out sooner than I expected 😅 - thanks for the catch and fix!

PR #1047, opened 2025-08-29
Set continue_on_error at job level for undercover and rubocop todo

apiology:
Right now unless someone dives in deep to check, there's no useful feedback to them on whether the undercover or rubocop todo checks succeeded or failed.  The UI for pulling out 'exited with error code' otherwise is comically bad :( There's discussion [here](https://github.com/orgs/community/discussions/11592) on that.

With this the mark will be back to red.

To make it clear which what is mandatory and what is optional, I'd suggest setting a [branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule) on the master branch that requires some (but not undercover and the rubocop todo) checks to pass - I think that will give at least a little better UX to let people know there are things they can dig in on, but that you're not requiring that they do so necessarily - see the 'required' bubble here:

image

I'll leave the Rubocop TODO check here if you're inclined to fiddle with that setting...

Another idea for undercover is that I can probably rig it up to leave comments in the diff of the PR on the uncovered parts - let me know if you'd prefer that approach.


apiology:
@castwide Ready for review

PR #1053, opened 2025-08-30
rubocop -a

apiology:
See the commit list for what's been done - to regenerate (until merged, this will generate conflicts pretty regularly):

```sh
# Error out instead of continuing
set -e
# Reset any existing commits out
git checkout rubocop_a
git fetch -a
git reset --hard origin/master
# Tell RuboCop not to prepend unused arguments with `_`, leaving Solargraph params unlinked, then run autocorrect and push up changes:
gsed -i 's/^  AllowUnusedKeywordArguments: true/  Enabled: false/g' .rubocop.yml
rm .rubocop_todo.yml
touch .rubocop_todo.yml
bundle exec rubocop -a -c .rubocop.yml || true
git checkout .rubocop.yml # revert rule change
bundle exec rubocop --auto-gen-config --no-exclude-limit --no-offense-counts --no-auto-gen-timestamp
git commit -a -m "rubocop -a"
git push -f
```

The only failure here is solargraph typechecking at strong level of every touched file, which is expected.

After this PR is merged, other PRs will need to do the commands above before merging master in to make that merge happen smoothly.


lib/solargraph/api_map/store.rb
```diff
@@ -37,10 +37,10 @@ def update *pinsets
...
           @pinsets[changed + idx] = pins
           @indexes[changed + idx] = if pins.empty?
-            @indexes[changed + idx - 1]
-          else
-            @indexes[changed + idx - 1].merge(pins)
-          end
+                                      @indexes[changed + idx - 1]
+                                    else
+                                      @indexes[changed + idx - 1].merge(pins)
+                                    end
```
apiology: Check against editor defaults

apiology:
Merged into 0.59 release branch

PR #1056, opened 2025-08-31
Qualify include tags [regression]

apiology:
It looks like https://github.com/castwide/solargraph/pull/1048 regressed the ability for ApiMap#get_methods to find the method in this situation:

```ruby
module A
  module B
    # @return [String]
    def foo
      'foo'
    end
  end
end

class A::C
  include B
end
```

It looks like the 'B' in this situation doesn't get qualified like it used to.

I encountered this in practice in #1005 - the gemspec convention stopped working in that PR after master was merged in.


castwide:
I'm working on a change to `ApiMap::Index` that will allow `ApiMap` to qualify references based on their open gates at the point of declaration. That might address the regression you're seeing in #1005.

apiology:
Aha!

I think I for some reason collapsed the auto-generated nested class declaration in an RBS file and replaced it with the A::C format.  I can't for the life of me remember having gone out of my way to do that, but I don't see any reason to think otherwise.

I fixed the syntax in my RBS on that branch and am closing this PR.

If the spirit moves you, would love to have a sentence or two describing the gates concept, maybe in the YARD doc for gates somewhere. It doesn't seem to stick in my head :)

castwide:
Cool, glad you found a fix that works as expected with gates :+1:

The changes to mixins in #1058 (specifically how they're resolved from reference pins) will need explicit and correct support of open gates. I'll keep documentation in mind as I work on it.

PR #892, opened 2025-04-12
Run specs from solargraph-rails configured against current code in CI

apiology:
Depends on https://github.com/iftheshoefritz/solargraph-rails/pull/161

PR #1058, opened 2025-08-31
Refactor reference pin handling

castwide:


apiology:
Hi @castwide!

I'm seeing a regression in rubocop-rails on this branch.  If you merge #892 you'll see this failure: https://github.com/apiology/solargraph/actions/runs/17531329632/job/49788680022?pr=11

Just let me know if these are invalid RBS or something like that we don't want to adapt to - I can make a release with some overrides ahead of the next Solargraph release if so.

apiology:
Whoops - try merging https://github.com/castwide/solargraph/pull/1073 to get the actual list of failures - the solargraph rails specs weren't passing on master due to an old workaround I had added.

apiology:
```
       The following methods could not be found despite being listed in actioncontroller.yml:
       [".log_at", ".permissions_policy"]
```

These look to be solargraph-rails' fault.  It uses 'class ActionController::Base' in https://github.com/iftheshoefritz/solargraph-rails/blob/main/lib/solargraph/rails/annotations/action_controller.rb#L1 along with relative includes.  I'll get a PR out, but I don't think we need to wait on it given the very limited impact - looks like the pins that aren't working are the ones whose includes aren't handled in RBS: 

https://github.com/ruby/gem_rbs_collection/blob/main/gems/actionpack/6.0/actioncontroller.rbs#L6

We can't get these includes out of the YARD info because they do some metaprogramming in the actual source:

https://github.com/rails/rails/blob/main/actionpack/lib/action_controller/base.rb#L230-L267

castwide:
@apiology Cool, thanks for letting me know. I'll probably merge this PR as is and continue the work in a followup.

apiology:
OK, the fix is merged to main in solargraph-rails and things should pass now.

PR #1006, opened 2025-07-15
Improve pin caching

apiology:
This PR has been broken into the following parts:
* https://github.com/castwide/solargraph/issues/1042
* https://github.com/castwide/solargraph/pull/1059
* https://github.com/castwide/solargraph/pull/1057
* https://github.com/castwide/solargraph/pull/1060
* #1061 
* #1067 
* #1064 
* https://github.com/castwide/solargraph/pull/1112
* https://github.com/castwide/solargraph/pull/1113

I'll keep this open and use it after these are merged to make sure the result is what I was looking for, but you can disregard this diff entirely.

# Original PR description: 

Way too much inside - apologies for the large monolithic change:

1. `solargraph gems` caches all cacheable gems/core/stdlibs for a workspace, allowing for CI-based ahead-of-time caching
1. Rebuild logic to deal with external bundles and add specs
   * Interested in whether this helps real-world scenarios!  
   * Big caveat: all gems need to be available in the environment Solargraph runs in - maybe we could dynamically install them in the future?
1. Performance: Define what we need from Parser gem in RBS and stop trying to process it as YARD (one-off for now, but can be made into config preferences for other projects...)
1. Refactor things out of DocMap class:
   * Workspace::Gemspecs: Resolves gemspecs and dependencies from bundle
   * Workspace::RequirePaths: Calculates the require paths from config and .gemspec file if it exists
   * PinCache: Gem building logic
1. 100% spec coverage of all affected lines, including more than a few new scenarios
1. Improved resolution of requires into gems driven by specific problematic cases now in specs
1. More linting (probably too much) to make RuboCop happy
1. Add rspec timeouts, which were valuable figuring out some plugin-related issues in the LSP specs.

## Perf note

Watching the checks, I was a little worried the plugin regression workflows had slowed down.  After isolating things a little more, I think it's possible, but I'm not convinced that more data points won't average this all down to the same thing in the end.  I did separate out the plugin regression workflows to tease out impact of each.  Here's the data I see:

Specs:

* solargraph-rails alone: ~5m
* solargraph-rspec alone: ~7m
* both plugins: ~13m
* no plugins specs alone: ~7m

Typechecking

* solargraph-rails alone: ~2.5m
* solargraph-rspec alone: ~2.5m
* both plugins: ~3m
* no plugins: ~2.5m

I'd suggest we keep an eye on this, but I am comfortable personally moving forward with this PR in the meantime.

## PR complexity note

[...]

PR #1061, opened 2025-09-03
Cache stdlib/core from command line

apiology:
This adds the ability to cache both the core and standard libraries
from the command line:

```
$ bundle exec solargraph help gem
Usage:
  solargraph gems [GEM[=VERSION]...] [STDLIB...] [core]

Description:
  This command will cache the generated type documentation for the specified libraries. While Solargraph will generate this on the
  fly when needed, it takes time. This command will generate it in advance, which can be useful for CI scenarios.

  With no arguments, it will cache all libraries in the current workspace. If a gem or standard library name is specified, it will
  cache that library's type documentation.

  An equals sign after a gem will allow a specific gem version to be cached.

  The 'core' argument can be used to cache the type documentation for the core Ruby libraries.

  Cached documentation is stored in /Users/broz/.cache/solargraph, which can be stored between CI runs.
$
```


apiology:
This PR takes a chunk of complexity out of #1006 - apologies if some of the code looks a little awkward or out of place out of that context, but it should be enough to be a useful chunk of functionality on its own.

apiology:
Overcommit will fail until https://github.com/castwide/solargraph/pull/1059 is merged

lib/solargraph/shell.rb
```diff
@@ -103,12 +103,60 @@ def clear
...
+    def gems *names
+      $stderr.puts("Caching these gems: #{names}")
+      # print time with ms
+      workspace = Solargraph::Workspace.new('.')
+
       api_map = Solargraph::ApiMap.load(Dir.pwd)
-      spec = Gem::Specification.find_by_name(gem, version)
-      api_map.cache_gem(spec, rebuild: options[:rebuild], out: $stdout)
+      if names.empty?
+        api_map.cache_all!($stdout)
```
apiology: I think this is better, as we'll use requires in the project to figure out what gems we have.  Gem::Specification.to_a is both too much (everything installed for all things if you're not using bundler), and too little - it doesn't handle gems which are installed in bundler's area - leading to 'solargraph gems' not fully pre-caching gems.

I think the best approach is to use Bundler itself when we're in a bundle to get the full list of things used by this workspace - that's in a different PR, though.

lib/solargraph/doc_map.rb
```diff
@@ -61,19 +61,27 @@ def initialize(requires, preferences, workspace = nil)
...
     # @return [void]
-    def cache_all!(out)
-      # if we log at debug level:
-      if logger.info?
-        gem_desc = uncached_gemspecs.map { |gemspec| "#{gemspec.name}:#{gemspec.version}" }.join(', ')
-        logger.info "Caching pins for gems: #{gem_desc}" unless uncached_gemspecs.empty?
-      end
-      logger.debug { "Caching for YARD: #{uncached_yard_gemspecs.map(&:name)}" }
-      logger.debug { "Caching for RBS collection: #{uncached_rbs_collection_gemspecs.map(&:name)}" }
+    def cache_all!(out, rebuild: false)
```
apiology, line 67: - [ ] Does not use 'rebuild' flag

apiology:

PR #1062, opened 2025-09-03
Factor out require_paths logic to its own class

apiology:


apiology:
This is another chunk out of #1006

lib/solargraph/workspace.rb
```diff
@@ -32,10 +28,17 @@ def initialize directory = '', config = nil, server = {}
...
       @gemnames = []
-      @require_paths = generate_require_paths
       require_plugins
     end
 
+    # The require paths associated with the workspace.
+    #
+    # @return [Array<String>]
+    def require_paths
+      # @todo are the semantics of '*' the same as '', meaning 'don't send back any require paths'?
```
castwide, line 38: The exceptional handling for '*' can probably be deprecated at some point. I'm having trouble thinking of use cases where it's still necessary and can't be handled in other ways that don't introduce confusion.

PR #1063, opened 2025-09-03
Add new Pin::FactoryParameter

lekemula:
Relates to:  https://github.com/lekemula/solargraph-rspec/pull/18 and https://github.com/lekemula/solargraph-rspec/pull/14
## Problem

I would like to add support for RSpec shared examples via solargraph-rspec plugin:  [solargraph-rspec#18](https://github.com/lekemula/solargraph-rspec/pull/18)

**Current limitation:** When using RSpec shared examples like this:
```ruby
RSpec.shared_examples "some example" do |parameter|
  # example implementation
end

RSpec.describe SomeClass do
  include_examples "some example", "parameter1"
  include_examples "some example", "parameter2" 
end
```

Solargraph cannot:
- Navigate to the shared example definition when clicking on `"some example"`
- Provide autocompletion for available shared example names

## Solution

This PR introduces a new `Pin::FactoryParameter` class that enables plugins (particularly solargraph-rspec) to register method parameters with specific values and their corresponding definitions.

### How it works

The new pin allows plugins to define factory-like method parameters:

```ruby
pin = Solargraph::Pin::FactoryParameter.new(
  method_name: 'it_behaves_like',
  method_namespace: 'RSpec::Core::ExampleGroup', 
  method_scope: :class,
  param_name: 'name',
  value: 'some shared example name', # or symbol :some_shared_example_name
  return_type: nil,
  decl: :arg,
  location: PinFactory.build_location(location_range, source_map.source.filename)
)
```

See [FactoryParameter#initialize method documentation](https://github.com/castwide/solargraph/pull/1063/files#diff-37cede688485d64c77a51d6bbd731ed4db57bcd4b43a2fc28a04a8e0ddc3ee67R35-R42) what each parameter means.

### Key benefits

1. **Go-to-definition support**: Jump directly to shared example definitions
2. **Type inference**: Better return type resolution for factory methods
3. **Parameter completion**: Allow users to get autocompletion as they type
4. **Extensibility**: Foundation for supporting other Ruby DSLs

## Why a new pin type?

This approach creates a flexible foundation that can support multiple Ruby DSL patterns beyond RSpec:

### Immediate applications
- **RSpec shared examples**: `include_examples "example_name"`
- **RSpec shared contexts**: `include_context "context_name"`

### Future possibilities  
- **Factory Bot**: `create(:model_name)` with go-to-definition to factory definitions
	- see [it 'infers types from factory methods' do](https://github.com/castwide/solargraph/pull/1063/files#diff-42d5879928fb4b4d50508cfce76878d252ae0f4bc2c603b7e9234fe76af9ce54R90) as an example
- **Rails associations**: `belongs_to :model_name` with navigation to model files
- **Rails validations**: `validates :method_name` with navigation to method definitions
- **Custom DSLs**: Any method that accepts string/symbol parameters referencing definitions

## Scope of this PR

This PR focuses on implementing the core infrastructure and two primary features:

✅ **Go-to-definition support** - Navigate from parameter values to their definitions
✅ **Return type inference** - Lay groundwork for better type resolution

These features directly address the most requested functionality from the solargraph-rspec.

## Future enhancements

- [ ] **Parameter autocompletion**: Show available factory parameter values while typing
- [ ] **Refactor require calls**: Leverage existing `Solargraph::Source::Chain::Parameter` infrastructure  
- [ ] **Enhanced validation**: Warn when referencing non-existent factory parameters

## Questions for review

1. **API design**: Is `Pin::FactoryParameter` an appropriate name and structure?
2. **Performance**: Are the lookup and indexing mechanisms efficient for large codebases?


lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -130,13 +133,30 @@ def generate_links n
           elsif n.type == :array
             chained_children = n.children.map { |c| NodeChainer.chain(c) }
             result.push Source::Chain::Array.new(chained_children, n)
+          elsif inside_method_call? # If the node is inside a method call, it may be a parameter.
+            lit = infer_literal_node_type(n)
+            if lit
+              method_call_chain = NodeChainer.chain(send_node, @filename, nil, [])
+              literal = Chain::Literal.new(lit, n)
+              arg_index = send_node.children[2..-1].index(n)
+              result.push Chain::Parameter.new(literal, method_call_chain)
```
lekemula, line 161: Initially, I wanted to make this work as a chain of: `method_call > parameter`; however, I struggled to find a good way to pass the last method call resolution to the `Chain::Parameter` as a context here:

https://github.com/castwide/solargraph/blob/48f3d08299fb9dd7e6074c7575f85869a4fcfe45/lib/solargraph/source/chain.rb#L108-L133

So I ended up with `parameter(method_call_chain)`. 

Maybe I don't understand the `Pin::ProxyType` well enough, so if you have a suggestion on how to make it work, I'd be more than happy to refactor this into an "appropriate chain".

lib/solargraph/pin/factory_parameter.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module Solargraph
+  module Pin
+    class FactoryParameter < Base
```
lekemula, line 19: Open to other naming suggestions!

spec/source/source_chainer_spec.rb
```diff
@@ -80,6 +80,21 @@
...
+  it "recognizes method literal parameters" do
+    source = Solargraph::Source.load_string(<<~RUBY)
+      SomeClass
+        .instance
+        .foo(:bar, "baz", 1)
+    RUBY
+    map = Solargraph::SourceMap.map(source)
+    cursor = map.cursor_at(Solargraph::Position.new(2, 10)) # :bar
+    expect(cursor.chain.links.last).to be_a(Solargraph::Source::Chain::Parameter)
+    expect(cursor.chain.links.map(&:word)).to eq(['foo(.., <::Symbol>, ..)'])
```
lekemula, line 94: I'm NOT very happy with this "word notation": `foo(.., <::Symbol>, ..)`, other ideas?

Another one that I thought of was: `foo(name=<::Symbol>)`, maybe?

lekemula:
> I don't think I understood how to actually use them? In the case of factory bot for eg? What do the various params do?

I actually describe it in `spec/source/chain_spec.rb`( see `it 'infers types from factory methods' do`) spec, have you checked it by any chance? 

Let me know if the spec is still unclear.

ShadiestGoat:
Very interesting. I wonder how much better this is than a method override as I proposed in my factory bot pr? Also, I might've just missed this in my quick glance, but is this able to determine other parts of the signature based on the value? Like other params available?

Obviously I'm no maintainer here and I haven't touched this codebase in like at least a month so I have no say, but given examples from other languages like typescript, it makes more sense to me to do overrides w/ different signatures than a different type of pin. Like imo the signature detection and use of it should improve rather than adding smt like this. What do you think?

lekemula:
Thank you so much @ShadiestGoat for your feedback - you’ve raised some excellent questions!  

On the question of _“why introduce a new pin instead of using overload signatures?”_: I see overloads as most useful when a method has a finite set of known input values (for example, enum-like parameters such as `@param [:value1, :value2, …]`). In our case, with calls like `RSpec.include_examples(factory_name=)` or `FactoryBot#create(model_name=)`, the parameter values are effectively open-ended (`@param [String, Symbol]`) and depend on definitions elsewhere — e.g. `RSpec.shared_examples` or `FactoryBot.define :factory_name`.  

As far as I know, no type system can fully capture this sort of dynamic binding. That’s why I believe it’s more appropriate for a solargraph plugin to provide that connection via a new `FactoryParameter` pin.  

> *Also, I might’ve just missed this in my quick glance, but is this able to determine other parts of the signature based on the value? Like other params available?*  

That was outside the scope of this particular ticket (it wasn’t needed for my case), but the current implementation could certainly be extended to support something along those lines.  

As for:  

> *Like the fundamental problem is that solargraph's completion & validation for constants/literals is lacking & that override go to definition is not supported. Those are the problems that need to be fixed. Especially since this is currently valid code according to sg (afaik):*  

I agree with you here - but I see that as a separate issue, more aligned with the case I mentioned earlier where values are known ahead of time from the signature itself.  

I’m not dismissing your suggestion at all; in fact, I’d love to hear what @castwide thinks as well. I just wanted to clarify my reasoning for introducing a new pin instead of relying on overloads, as you did in your [FactoryBot PR](https://github.com/lekemula/solargraph-rspec/pull/14) (which I really appreciate ❤️). And please don’t take this PR as a blocker for your work there!  

**TL;DR:** Using overloads here felt semantically incorrect, and on top of that, they seemed harder to make work (at least per my current understanding of the codebase).

lekemula:
From the practical perspective, another reason why I favoured the new pin over overload is that with the later:

```ruby
RSpec.include_examples("some shared example")
      #^^^^^^^^^^^^^ Uses needs to click on the method name to jump to the definition
      # We lose the method definition
```

Whereas with the new pin

```ruby
RSpec.include_examples("some shared example")
      #                ^^^^^^^^^^^^^^^^^^^^ User clicks on the param
      #^^^^^^^^^^^^^ User can still jump to definition of the "real method"
```

ShadiestGoat:
That's fair - thank you for the expansion. I do agree that it's not exactly ergonomic to use overloads.

Your last point sounds like a really cool use case (though I'm not 100% convinced that it can't be done with overloads).

I'm not sure I understand about your first point - rspec shared examples, factory models, or even rails `belongs_to` *aren't* open ended from the perspective of a plugin. Like a plugin that'd implement factory params would know what are the available values, right?

PS: Just like you, I'm also not suggesting that this is a bad implementation/idea, I'm trying to explore this idea/issue further. I really love the rspec plugin and it has saved me tons of headaches at work :heart:

lekemula:
> Your last point sounds like a really cool use case (though I'm not 100% convinced that it can't be done with overloads).

You're actually right here, it's rather a question of _should we_, and later how hard and efficiently we could do that.

> I'm not sure I understand about your first point - rspec shared examples, factory models, or even rails `belongs_to` *aren't* open ended from the perspective of a plugin. Like a plugin that'd implement factory params would know what are the available values, right?

I meant from the perspective of the method definition. I think, `@overload` is primarily meant to express the acceptable params prior to its usage, and let alone to the relationship of other methods.

Let me use `Hash#[]` as an example. We could in theory track the key assignments of a given variable, and suggest the possible keys and/or their return values  eg. `my_hash["my_key"]`, but would the usage of `@overload` or invention of a "imaginary subclass" of a hash with overloaded `[]` with the possible keys, be the right way? It surely would do the trick, but IMO it's "a trick" nonetheless.

Maybe I'm being over-pedantic here.

> I really love the rspec plugin and it has saved me tons of headaches at work :heart:

This really warms my heart to hear! 🥰

lekemula:
Hey @castwide, I'd like to get back to adding support for RSpec shared examples in the `solargraph-rspec` gem. Do you think we can prioritise this PR for the next release? Are there any big concerns from your side regarding the approach/design, ie. the introduction of the new pin?

lib/solargraph/api_map/store.rb
```diff
@@ -317,9 +325,9 @@ def catalog pinsets, &block
         true
       end
 
-      # @return [Hash{::Array(String, String) => ::Array<Pin::Namespace>}]
+      # @return [Hash{::Array, String, String => ::Array<Pin::Namespace>}]
```
lekemula: Related to https://github.com/ksss/rubocop-yard/issues/43 

Will revert this. should be fixed by bumping `rubocop-yard`. I've also opened a real fix for it in the `yard` gem.

spec/source/source_chainer_spec.rb
castwide: It's probably fine if the `word` is the same as `literal.word`. As a general rule, the `word` attribute is just a context-free representation of the code at that particular link in the chain. The `Call#word` attribute, for example, can be a method name, but it does not include any reference to the recipient.

PR #1064, opened 2025-09-03
Move more pin-caching logic from DocMap to PinCache

apiology:
Changes in logic:
* We treat the parser gem as a special case given its degenerate behavior with YARD, and only ingest RBS pins.  https://github.com/castwide/solargraph/pull/1005 [provides](https://github.com/castwide/solargraph/pull/1005/files#diff-f722e983a0d62e429ce9f99530cb6010dabf86d77ea816d1bc8fb8088606b1b3) enough RBS updates to be able to typecheck solargraph in strong mode, which ought to be enough for our use.  If YARD performance becomes enough of an issue maybe this could be a configuration item someday.
* More nuanced logging for what type of RBS collection is being used (local shims, gem-provided types, or central RBS collection)

Includes for convenience:
* #1102

### DIFF
--- a/lib/solargraph/doc_map.rb
+++ b/lib/solargraph/doc_map.rb
@@ -145,6 +145,7 @@ module Solargraph
         # try to resolve the stdlib name
         deps = workspace.stdlib_dependencies(stdlib_name_guess) || []
         [stdlib_name_guess, *deps].compact.each do |potential_stdlib_name|
+          # @todo Need to support splatting in literal array
           rbs_pins = pin_cache.cache_stdlib_rbs_map potential_stdlib_name
           serialized_pins.concat rbs_pins if rbs_pins
         end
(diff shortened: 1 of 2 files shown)


lib/solargraph/api_map/index.rb
```diff
@@ -169,10 +169,13 @@ def map_overrides
...
-                new_pin.docstring.add_tag(tag)
-                redefine_return_type new_pin, tag
-              end
+              pin.reset_generated!
+
+              next unless new_pin
+
+              new_pin.docstring.add_tag(tag)
+              redefine_return_type new_pin, tag
+              new_pin.reset_generated!
```
apiology, line 175: This fixes an existing bug, where not only are we not always calling reset_generated! on pins (only when return_type is overridden), and we don't reset the parameters.  

As a result, params overloading would only work if #inspect had not been called on the pin before it was overridden, which is bad news for debugging.

See #1066 for the spec behind it; on this branch it fails because some debug logging was added and we're [exercising debug logging](https://github.com/castwide/solargraph/pull/1064/files#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94R29-R31) in specs.

apiology:

PR #1067, opened 2025-09-04
Extract gemspec, bundle and dependency management into its own class

apiology:


apiology:

PR #1071, opened 2025-09-06
Add `solargraph profile` command using vernier

lekemula:
Hi,

I thought of adding a profiling command for convenience, to avoid the hassle of configuring the IDE, switching gem versions, restarting, and all that hassle:

```
❯ solargraph --help profile
Usage:
  solargraph profile [FILE]

Options:
  -d, [--directory=DIRECTORY]                     # The workspace directory
                                                  # Default: .
  -o, [--output-dir=OUTPUT_DIR]                   # The output directory for profiles
                                                  # Default: ./tmp/profiles
  -l, [--line=N]                                  # Line number (0-based)
                                                  # Default: 4
  -c, [--column=N]                                # Column number
                                                  # Default: 10
  -m, [--memory], [--no-memory], [--skip-memory]  # Include memory usage counter
                                                  # Default: true

Profile go-to-definition performance using vernier
```

By default, the `solargraph profile` will run against the current directory, and will find the first Ruby file with the default line/number. 

When testing new local changes, one could use the `bundle exec solargraph --directory=/path/to/a/real-project ..other params`

Here's a quick demo: [here](https://www.dropbox.com/scl/fi/435g82et7nj47tbxsedvn/solargraph_profile_demo.mp4?rlkey=rxj5odeu29gs0k08cz3uz5exm&st=yzzzme68&dl=0)

I hope you find it useful.


lekemula:
I'm not sure how my changes could lead to such breaking of the typechecking step [here](https://github.com/castwide/solargraph/actions/runs/17518614746/job/49759410094?pr=1071). 

@apiology do you recognise anything obvious there by any chance?

lib/solargraph/shell.rb
```diff
@@ -241,6 +241,104 @@ def list
...
+        return
+      end
+
+      hooks = []
+      hooks << :memory_usage if options[:memory]
+
+      directory = File.realpath(options[:directory])
+      FileUtils.mkdir_p(options[:output_dir])
+
+      host = Solargraph::LanguageServer::Host.new
```
lekemula, line 266: I have based the host setup, based on this spec:

https://github.com/castwide/solargraph/blob/872431ca5aed2ad0c054ab4bf4a7242a26ae24be/spec/language_server/message/text_document/definition_spec.rb#L2

Is there any critical step missing that should be included in the profile?

lib/solargraph/shell.rb
```diff
@@ -241,6 +241,104 @@ def list
...
+      host.client_capabilities.merge!({ 'window' => { 'workDoneProgress' => true } })
+      def host.send_notification method, params
+        puts "Notification: #{method} - #{params}"
+      end
```
lekemula, line 273: I thought I'd never find a use-case for this Ruby feature, but I finally found one 😍

apiology: Hey!  This is the line that triggered that assert - looks like Solargraph has never run through that code in its tests or typechecking in CI, either!

https://github.com/apiology/solargraph/blob/master/lib/solargraph/parser/parser_gem/node_processors/defs_node.rb creates some pins, but doesn't pass the 'source' kwarg in.  If you throw "source: :parser" into that "Solargraph::Pin::Namespace.new" call, it should get past that error.

apiology:
> Maybe as a next step, we could select some large open-source Rails projects that we can use to profile and track the merged changes in main through a CI step. Any suggestions for such projects? Personally, I've used https://github.com/gitlabhq/gitlabhq in numerous cases, but https://github.com/mastodon/mastodon also come to mind as good candidates. WDYT?

I've been thinking about doing the same thing as a functionality test.  @aaronc81 has some interesting work here in Sord that we could probably import with credit: https://github.com/AaronC81/sord/blob/master/Rakefile#L121-L138

List here of repos: https://github.com/AaronC81/sord/blob/master/Rakefile#L121-L138

solargraph.gemspec
```diff
@@ -59,6 +59,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'undercover', '~> 0.7'
   s.add_development_dependency 'overcommit', '~> 0.68.0'
   s.add_development_dependency 'webmock', '~> 3.6'
+  s.add_development_dependency 'vernier'
```
apiology:
```suggestion
  s.add_development_dependency 'vernier', '~> 1.8'
```

This will keep a 2.0 release from breaking our build so we can update on our own timeline, and make sure weird things don't happen if a user has an old version installed.

lib/solargraph/shell.rb
lekemula: Thanks again, @apiology, for coming to the rescue! 🙇‍♂️ 

For some reason, I thought the typechecker does not cover the shell area.

solargraph.gemspec
lekemula: Unfortunately both ['~> 1.8'](https://github.com/castwide/solargraph/actions/runs/17657536428/job/50183884510?pr=1071) and ['~> 1'](https://github.com/castwide/solargraph/actions/runs/17657829386/job/50184803707?pr=1071) failed, so it looks like we need `< 2`.

castwide:
See #1171

PR #1072, opened 2025-09-07
Make typechecker errors vim quickfix friendly

lekemula:
A minor change to the typechecker output, but a life-changing one for Vim/Emacs users.
(I mention only Vim in the title, because that's what we vim-ers do :) )

This is similar to: https://docs.rubocop.org/rubocop/formatters.html#emacs-style-formatter

Everything in Vim revolves around something called [quickfix](https://medium.com/@EduardoRodriguesF/vim-start-using-quickfix-lists-e43a7a29ba52). It's essentially a list of "things needed to be fixed" where you can easily navigate between next/prev/first/last, and do a lot more around them. 

Among other things, one can run a terminal command and have this list populated; for example:

`cexpr system("bundle exec solargraph typecheck --level strict")`, and have this convenient list populated:

image

### DIFF
--- a/lib/solargraph/shell.rb
+++ b/lib/solargraph/shell.rb
@@ -259,7 +259,7 @@ module Solargraph
           next if problems.empty?
           problems.sort! { |a, b| a.location.range.start.line <=> b.location.range.start.line }
           puts problems.map { |prob|
-            "#{prob.location.filename}:#{prob.location.range.start.line + 1} - #{prob.message}"
+            "#{prob.location.filename}:#{prob.location.range.start.line + 1}: #{prob.message}"
           }.join("\n")
           filecount += 1
           probcount += problems.length


lekemula:
Hmm.. I'm not sure why overcommit [fails](https://github.com/castwide/solargraph/actions/runs/17533623478/job/49793804162?pr=1072) on typechecking, when the typechecker step itself [succeeded](https://github.com/castwide/solargraph/actions/runs/17533623480/job/49793804160?pr=1072) 🤔

apiology:
It broke because overcommit is an automated tool that relies on the formatting of our messages already - see https://github.com/sds/overcommit/blob/main/lib/overcommit/hook/pre_commit/solargraph.rb

Maybe we can act like rubocop, make this a command line argument to specify the output format and keep the current default?

lekemula:
> It broke because overcommit is an automated tool that relies on the formatting of our messages already - see [sds/overcommit@`main`/lib/overcommit/hook/pre_commit/solargraph.rb](https://github.com/sds/overcommit/blob/main/lib/overcommit/hook/pre_commit/solargraph.rb?rgh-link-date=2025-09-07T23%3A24%3A02Z)
> 
> Maybe we can act like rubocop, make this a command line argument to specify the output format and keep the current default?

@apiology Thanks for the hint - I wonder though how much it make sense to keep both formats when esentialy the differ only by a single character. Would it make more sense to adapt the `MESSAGE_REGEX` on the overcommit side?

lekemula:
@apiology @castwide FYI prepared a draft PR for overcommit: https://github.com/sds/overcommit/pull/873

lekemula:
@castwide do you think we can move forward with this?

castwide:
LGTM! Thanks for the reminder.

PR #1075, opened 2025-09-10
Move Pin::Base#infer deprecation to assert system

### LINKED ISSUE #1052: [WARN] WARNING: Pin #infer methods are deprecated. Use #typify or #probe instead.
```
SOLARGRAPH_ASSERTS=on bundle exec solargraph typecheck --level typed
[WARN] WARNING: Pin #infer methods are deprecated. Use #typify or #probe instead.
Typecheck finished in 38.46436400001403 seconds.
0 problems found in 0 of 237 files.
```

Looks like a call to Pin::Base#infer got added somewhere.  Might be a good idea to move this over to the assertions framework so we catch it more easily...

apiology:
Fixes #1052


apiology:

PR #1081, opened 2025-09-12
Fix combination of directives

apiology:
This was creating two-dimensional arrays; we wanted a one-dimensional array

TODO:
- [ ] Reproduce in spec


apiology:
Dupe of #1124

PR #1083, opened 2025-09-18
Faster constant resolution

castwide:
Optimize constant resolution by making better use of gates, leveraging the store's path indexes, and avoiding unnecessary type inference.

`ApiMap#resolve` is similar to `ApiMap#qualify` in that its purpose is to find fully qualified (absolute) namespaces, with two exceptions:
* `#resolve` is only concerned with real namespaces. It does not support parametrized types or special types like literals, `self`, and `Boolean`.
* `#resolve` never performs type inference.

In general, I would recommend using `#resolve` over `#qualify` where possible. The only place where `#qualify` is still necessary is in complex types. I hope to refactor them to handle qualification internally while further reducing type inference.

### DIFF
--- a/lib/solargraph/pin/base.rb
+++ b/lib/solargraph/pin/base.rb
@@ -513,7 +513,7 @@ module Solargraph
       # @param api_map [ApiMap]
       # @return [ComplexType]
       def typify api_map
-        return_type.qualify(api_map, namespace)
+        return_type.qualify(api_map, *(closure&.gates || ['']))
       end
 
       # Infer the pin's return type via static code analysis.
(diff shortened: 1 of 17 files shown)


lib/solargraph/pin/base.rb
```diff
@@ -574,6 +574,18 @@ def identity
...
       end
 
+      # The namespaces available for resolving the current namespace. Each gate
+      # should be a fully qualified namespace or the root namespace (i.e., an
+      # empty string.)
+      #
+      # Example: Given the name 'Bar' and the gates ['Foo', ''],
+      # the fully qualified namespace should be 'Foo::Bar' or 'Bar'.
+      #
+      # @return [Array<string>]
```
apiology, line 584: ❤️

castwide:
> Any chance you could copy and paste the PR description into the YARD docs in ApiMap::Constants and add a @see from ApiMap's versions? I'm sure my head will constantly confuse the two :)

Yeah, good call. I'll add it in #1084.

PR #1088, opened 2025-09-21
Less-rigorous assertions merging assignment attribute of parameters

apiology:
Default values don't exist in RBS; it just tells you if the arg is optional or not - this was a problem in solargraph-rails specs while assertions were on.

### DIFF
--- a/lib/solargraph/pin/local_variable.rb
+++ b/lib/solargraph/pin/local_variable.rb
@@ -23,14 +23,22 @@ module Solargraph
 
       def combine_with(other, attrs={})
         new_attrs = {
-          assignment: assert_same(other, :assignment),
+          # default values don't exist in RBS; it just tells you if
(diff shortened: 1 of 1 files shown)


lib/solargraph/pin/local_variable.rb
```diff
@@ -10,20 +10,17 @@ def presence_certain?
...
+      def initialize presence: nil, presence_certain: false, **splat
         super(**splat)
-        @assignment = assignment
         @presence = presence
         @presence_certain = presence_certain
       end
 
       def combine_with(other, attrs={})
         new_attrs = {
-          assignment: assert_same(other, :assignment),
```
apiology, line 26: This was redundant with BaseVariable - let's centralize the logic instead of changing two places

apiology:

PR #1094, opened 2025-09-24
[regression] Handle RBS static method aliases

apiology:
Should fix solargraph-rails spec failures


castwide:
Thanks for tracking this down! :100: :shipit:

PR #1097, opened 2025-09-25
[regression] Fix resolution issues with namespaces from YARD 

apiology:
YARD-parsed namespaces weren't correctly setting their gates, leading to unresolved types from methods.

The fix relies on #1096, so I've merged it in


lib/solargraph/pin/base.rb
```diff
@@ -353,8 +353,15 @@ def choose_pin_attr(other, attr)
...
         # arbitrary way of choosing a pin
-        # @sg-ignore Need _1 support
-        [val1, val2].compact.min_by { _1.best_location.to_s }
+        [val1, val2].compact.max_by do |closure|
+          [
+            # maximize number of gates, as types in other combined pins may
+            # depend on those gates
+            closure.gates.length,
+            # use basename so that results don't vary system to system
+            File.basename(closure.best_location.to_s)
```
apiology, line 362: I was hitting a strange 'works on my machine' issue and figured it out that I wasn't seeing it because of a difference in where files lived in CI vs my machine. This resulted in different pin info being selected during combination, as we were using the file location when we had no better way to prefer one over the other.

This should make the result more consistent, making user and CI pin-combination-triggered issues easier to reproduce.

lib/solargraph/api_map/constants.rb
```diff
@@ -25,7 +25,13 @@ def resolve(name, *gates)
...
         flat = gates.flatten
         flat.push '' if flat.empty?
-        cached_resolve[[name, flat]] || resolve_and_cache(name, flat)
+        if cached_resolve.include? [name, flat]
+          cached_result = cached_resolve[[name, flat]]
+          # don't recurse
+          return nil if cached_result == :in_process
+          return cached_result
+        end
+        resolve_and_cache(name, flat)
```
castwide, line 34: :+1:

PR #1098, opened 2025-09-25
[regression] Fix resolution of ambiguous argument types

apiology:
This is a case where incorrect gates result in the wrong type being used - found by strong typechecking on a branch ::Solargraph::Pin::Symbol was resolved as ::Symbol in a generics scenario.


lib/solargraph/source/chain/call.rb
apiology, line 76: call.rb could use a refactor for sure - we should keep that to a separate PR, though.  Adding to rubocop todo file.

PR #1087, opened 2025-09-21
Enable solargraph-rspec tests

lekemula:
https://github.com/castwide/solargraph/actions/runs/17897241460/job/50885481518?pr=1087

Thank you, @apiology, for setting this important workflow. I think it's very useful to make it work also for solargraph-rspec to prevent regressions in the future. I also appreciate the reminder about the `act`, which was a lifesaver!	

I assume the other CI failures are not related to this PR, if otherwise, let me know!

### Future Work

I think we need to revisit caching of gems/yard/rbs files across the CI workflows. See [similar work](https://github.com/lekemula/solargraph-rspec/commit/e492aa0d6ee87e648f973584b4086d1355021a18) on solargraph-rspec.

PR #1100, opened 2025-09-26
Resolve constants in references

### LINKED ISSUE #1099: `include Parser::NodeMethods` not resolving
It looks like while the constant resolution mapping Solargraph::Parser::NodeMethods to Solargraph::Parser::ParserGem::NodeMethods is working great, the `include` lines are generally to `Parser::NodeMethods` - which doesn't resolve.  I suspect we're missing a gate somewhere.

apiology:
Also replaces 'include' logic with call to ApiMap::Constants

Fixes #1099

Includes:
* https://github.com/castwide/solargraph/pull/1087 (saw it break under this PR in my dev branch)


apiology:
Interested to see if this creates the same perf problem as before...

PR #1101, opened 2025-09-27
Fix pin combination consistency issue

apiology:
I was hitting a strange 'works on my machine' issue and figured it out that I wasn't seeing it because of a difference in where files lived in CI vs my machine.  This resulted in different pin info being selected during combination, as we were using the file location when we had no better way to prefer one over the other.

This should make the result more consistent, making user and CI pin-combination-triggered issues easier to reproduce.

### DIFF
--- a/lib/solargraph/pin/base.rb
+++ b/lib/solargraph/pin/base.rb
@@ -354,7 +354,7 @@ module Solargraph
         end
         # arbitrary way of choosing a pin
         # @sg-ignore Need _1 support
-        [val1, val2].compact.min_by { _1.best_location.to_s }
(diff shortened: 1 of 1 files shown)


lib/solargraph/pin/base.rb
```diff
@@ -354,7 +354,7 @@ def choose_pin_attr(other, attr)
         end
         # arbitrary way of choosing a pin
         # @sg-ignore Need _1 support
-        [val1, val2].compact.min_by { _1.best_location.to_s }
+        [val1, val2].compact.max_by { File.basename(_1.best_location.to_s) }
```
apiology, line 357: min vs max is arbitrary, but this choice lets me reproduce the original issue on my machine rather than papering it over :)

PR #1102, opened 2025-09-28
Reduce number of build jobs for faster CI feedback

apiology:
This takes out some lower value combinations - ideally we could keep the number of jobs to <= 20, which is the max that GHA will run simultaneously here.


apiology:

PR #1104, opened 2025-09-28
Make RBS first-class pin data source; apply every @!override overload

apiology:

**Problem:** RBS supplies most of what Solargraph knows about core and stdlib, but LSP's `#documentation` renders the YARD docstring alone, so RBS data only counted once translated into YARD tags. Parsing RBS wrote eight of them, one per kind of pin, flattening a type it already had into tag text:

```ruby
constant_pin.docstring.add_tag(YARD::Tags::Tag.new(:return, '', rooted_tag))
pin.docstring.add_tag(YARD::Tags::Tag.new(:type, '', rooted_tag))
class_pin.docstring.add_tag(YARD::Tags::Tag.new(:abstract, '(RBS interface)'))
```

Even then it did not all render. A method typed by RBS documented none of its parameters, and the abstract marker showed nowhere:

```ruby
# Hovering String#sub, before:
#   Returns:
#   * [::String]
#
# The parameter types are on the pin. Nothing renders them.
```

**Solution:** RBS values are held on the pin as themselves — `return_type:`, `abstract:` — and LSP's `#documentation` reads pin and docstring together: descriptions and tags still come from comments, types now come from the pin. The conversion writes no YARD tags.

**Problem:** An `@!override` with two `@overload` tags applied only the first, and on a method typed by RBS the override was discarded entirely.

```ruby
# @!override Array#first
#   @overload first(n)
#     @param n [Integer]
#     @return [Array<String>]

# Array#first ends up with one signature, (n) -> Array<String>.
# RBS's zero-argument () -> Elem signature is gone.
```

Whether it applied at all depended on what had already read the method's signatures, because the tags were added to a shared, already-materialised pin.

**Solution:** An override is now a pin of its own, ranked above its target and combined through the join below, with the result swapped into the index instead of mutating a pin every consumer already holds.

**Problem:** Because both sources had to share the one docstring, combining or proxying a pin lost data, or produced a pin that documented one source and rendered from another:

- `#combine_with` took comments from one side and the docstring from the other, so a merged pin matched neither, and proxying flipped which side showed
- proxying dropped every docstring tag but the return type, so an `attr_writer`'s `@param` vanished from hover
- replacing a `@return` tag's types dropped its prose description
- a docstring supplied at construction was discarded when there were no comments to reparse from

**Solution:** `#combine_with` becomes the join between sources, each keeping its own representation: a pin may declare authority, winning each field it supplies, with docstrings merged tag-wise rather than one chosen whole.


lib/solargraph/pin/base.rb
```diff
@@ -617,9 +617,9 @@ def probed?
       # @param api_map [ApiMap]
       # @return [self]
       def realize api_map
-        return self if return_type.defined?
+        return self if return_type.defined? && return_type.all_rooted?
         type = typify(api_map)
-        return proxy(type) if type.defined?
+        return proxy(type) if type.defined? && type.all_rooted?
```
apiology, line 622: Could you clarify what if any spec triggered the need for this?  If none, can we add one or more that would fail without the above?

apiology: Good catch — there wasn't one. Added `spec/pin/base_spec.rb` `#realize roots an already-defined but unrooted return type`, which fails without the `all_rooted?` checks (a pin built straight from an unrooted `@return [String]` tag never gets qualified against the ApiMap, so `#realize` short-circuits and returns `String` instead of `::String`). Pushed in 272c0899a.

apiology: **Claude:** (correcting the previous reply, which was posted unsigned)

apiology:
@castwide - ready for review

spec/pin/constant_spec.rb
```diff
@@ -23,4 +23,12 @@ class Foo
...
     expect(pin.symbol_kind).to eq(Solargraph::LanguageServer::SymbolKinds::CONSTANT)
   end
+
+  it 'resyncs the docstring :return tag, not the :type tag other variable pins use' do
+    pin = described_class.new(name: 'BAR', closure: Solargraph::Pin::ROOT_PIN, comments: '@return [String]')
+    realized = pin.realize(Solargraph::ApiMap.new)
+    expect(realized.docstring.tag(:return).types).to eq(['::String'])
+    expect(realized.docstring.tag(:type)).to be_nil
+    expect(realized.return_type.rooted_tags).to eq('::String')
+  end
```
apiology: constants use `@return`?  I would have expected `@type`

lib/solargraph/pin/constant.rb
```diff
@@ -38,6 +38,11 @@ def generate_complex_type
         return ComplexType::UNDEFINED if tags.empty?
         ComplexType.try_parse(*tags)
       end
+
+      # @return [::Symbol]
+      def return_type_tag_name
+        :return
```
apiology: See comment in spec

lib/solargraph/pin/base.rb
```diff
@@ -637,6 +637,10 @@ def proxy return_type
         result = dup
         result.return_type = return_type
         result.proxied = true
+        result.reset_generated!
+        # dup keeps the original docstring, whose tag reflects the old
+        # return_type. Clear it so #docstring resyncs via #parse_comments.
+        result.instance_variable_set(:@docstring, nil)
```
apiology: Why isn't this part of reset_generated! ?

spec/pin/constant_spec.rb
apiology: **Claude:** Checked both, on this branch and at the merge-base (6dcb73338).

YARD has no `@type` tag - Solargraph defines it, in `lib/solargraph/yard_tags.rb`:

```ruby
# Define a @type tag for documenting variables
YARD::Tags::Library.define_tag('Type', :type, :with_types_and_name)
```

```
$ ruby -ryard -e 'puts YARD::Tags::Library.instance.has_tag?(:type)'
false
$ ruby -ryard -e 'require "solargraph"; puts YARD::Tags::Library.instance.has_tag?(:type)'
true
```

So YARD parses `@return` on a constant and drops `@type`:

```ruby
YARD.parse_string("# @return [String]\nFOO = something\n")
# Registry.at('FOO').docstring.tags => [["return", ["String"]]]

YARD.parse_string("# @type [String]\nFOO = something\n")
# Registry.at('FOO').docstring.tags => []
# [warn]: Unknown tag @type in file `(stdin)` near line 2
```

`yard doc` over both renders `FOO` as "Returns: (String)" and `BAR` with no type at all.

Solargraph reads either and infers the same thing - `FOO`, `BAR`, `FOO.upcase`, `BAR.upcase` all give `::String` on this branch and at the merge-base, since `Constant#generate_complex_type` takes `:return` and falls back to `:type`. The tag name only decides what gets written back.

`:return` is also what master already writes. At 6dcb73338 the choice was seven hardcoded calls in `RbsMap::Conversions`, which this PR replaces:

```ruby
# line 369, Pin::Constant
constant_pin.docstring.add_tag(YARD::Tags::Tag.new(:return, '', rooted_tag))
# line 412, Pin::GlobalVariable
pin.docstring.add_tag(YARD::Tags::Tag.new(:type, '', rooted_tag))
```

`Pin::Constant < BaseVariable`, whose `return_type_tag_name` is `:type`, so without the override every RBS constant (`RUBY_VERSION`, `Float::INFINITY`, `ARGV`) would switch tag relative to master. The round-trip holds too: Solargraph emits `@return [::String]`, and plain YARD reads it back with no warning.

One wart either way: `sync_return_type_tag` matches one tag name, so a constant documented by hand with `@type [String]` ends up with both after realize - `[["return", ["::String"]], ["type", ["String"]]]`. Nothing reads the stale one and inference is unchanged; `:type` would just move the duplicate onto `@return`-documented constants.

Also: the new comment on `Base#return_type_tag_name` calls `:type` "the YARD convention for variables" - it is Solargraph's own, per `yard_tags.rb`.

lib/solargraph/pin/constant.rb
apiology: **Claude:** Answered on the spec thread: YARD has no `@type` tag at all (Solargraph defines it in `yard_tags.rb`), it parses `@return` on a constant and warns `Unknown tag @type`, and `:return` is what master already writes for constants via `RbsMap::Conversions` line 369. Keeping `:return`.

lib/solargraph/pin/base.rb
apiology: **Claude:** It cannot move - `@docstring` is not a generated value. It is a constructor attribute that several callers set independently of `#comments`, so there would be nothing to re-derive it from. `#combine_with` picks the two from different sides, nine lines before it resets:

```ruby
comments: choose_longer(other, :comments),   # base.rb:101
docstring: choose(other, :docstring),        # base.rb:103
...
out.reset_generated!                          # base.rb:112
```

Measured, a Method pin with the docstring from one side and comments from the other:

```
combined.comments  = "Zzz other comments."
combined.docstring = "Chosen docstring."     # clear in #proxy
combined.docstring = "Zzz other comments."   # clear in #reset_generated!
```

`reset_generated!` also runs from `Common#closure=`, so an ordinary reparent would drop it - e.g. the `initialize` pin `Convention::StructDefinition` builds with `docstring: docstring` and no `comments:` (struct_definition.rb:37), whose docstring carries the struct's `@param` tags with `comments == ""`.

The suite passes either way (1657 examples, same single pre-existing failure at `spec/pin/method_spec.rb:578`, which also fails at 6dcb73338), so it is not evidence of safety - `ApiMap::Index#map_overrides`, the site the existing comment cites, is at 0% coverage.

That citation is also the wrong example: `map_overrides` resyncs `pin.comments` from the docstring before resetting, so it survives. Worth pointing the comment at `#combine_with` instead.

`#proxy` is where it belongs anyway - it is the operation that invalidates the docstring, because it just replaced `return_type`.

apiology: Surely there's a better way to express this than reaching in to another class' instance variable?

lib/solargraph/pin/base.rb
```diff
@@ -637,6 +637,10 @@ def proxy return_type
         result = dup
         result.return_type = return_type
         result.proxied = true
+        result.reset_generated!
+        # dup keeps the old docstring, whose tag still holds the previous
+        # return_type; nil makes #docstring reparse from comments.
+        result.docstring = nil
```
apiology: "nil makes it reparse from comments" (and RBS info?)

Doesn't that imply that it is generated and reset_generated! would be appropriate?  If not, help me understand why not.

lib/solargraph/pin/base.rb
```diff
@@ -746,6 +750,36 @@ def parse_comments
           @directives = parse.directives
           @macro_names = collect_macro_names
         end
+        return unless @return_type&.defined?
+
+        @docstring ||= Solargraph::Source.parse_docstring("\n").to_docstring
+        sync_return_type_tag
+      end
+
+      # Reads @return_type, not #return_type: BaseVariable's getter reads
```
apiology: Not clear if @return_type means the yard tag in the comment or the instance variable

lib/solargraph/pin/base.rb
```diff
@@ -746,6 +750,36 @@ def parse_comments
...
+          @docstring.add_tag(YARD::Tags::Tag.new(return_type_tag_name, '', rooted_types))
+        end
+      end
+
+      # The docstring tag name that holds this pin's return type. Method
+      # and constant pins use :return; variable pins (ivars, cvars, gvars)
+      # use :type, the YARD convention for variables.
+      #
+      # @return [::Symbol]
+      def return_type_tag_name
```
apiology: Why not 'type tag name' or 'evaluated type tag name'?

We also use 'tag' to mean "YARD-style type" - does the 'tag_name' have precedent here or is there another naming pattern we use?

lib/solargraph/pin/method.rb
```diff
@@ -305,6 +305,12 @@ def typify api_map
 
       def documentation
         if @documentation.nil?
+          # Only force when it can't fall through to #generate_signature,
+          # which assumes real Parameter pins: already-resolved
+          # #signatures, or inline RBS rather than YARD.
+          if (instance_variable_get(:@signatures) || !inline_rbs.empty?) && docstring.tags(return_type_tag_name).empty? && return_type&.defined?
```
apiology: Why are we using instance_variable_get?

lib/solargraph/pin/base.rb
apiology: **Claude:** You are right, and my earlier answer was wrong in a way that matters: the clear is unsafe in `#proxy` too, not only in `reset_generated!`.

`@docstring` is generated only in the sense of *re-derivable from `comments`*, and there are three states, not two:

1. derived from comments - safe to clear;
2. supplied at construction - `docstring:` with empty `comments:`, e.g. the Struct `initialize` pin in `struct_definition.rb:37`;
3. derived, then mutated - `send_node.rb:124`, `attribute_directive.rb:49`, `method.rb:451`, `conversions.rb:311` all `add_tag` onto a comment-derived docstring after construction.

Nothing records which state a pin is in, and state 3 is why checking `comments` for emptiness cannot rescue the clear.

The RBS case you asked about is fine - this PR already moved those pins onto a constructor `return_type:`, so their only tag is the generated one and it regenerates correctly. But the clear drops everything else. Through `#realize`, which is what `library.rb:202` calls for hover:

```ruby
class Bar
  # @return [String]
  attr_writer :thing
end
# before: "Params:\n* value [String] \n\nReturns:\n* [String] \n\nVisibility: public"
# after : "Returns:\n* [::String] \n\nVisibility: public"
```

The `@param value` that `send_node.rb:124` adds is gone. Same for the Struct `initialize` params and the `@abstract` tag on RBS interface pins.

This is new here on two counts: `#proxy` on master keeps every tag (only its `@return` goes stale), and the `&& return_type.all_rooted?` added to `#realize` makes proxy fire on defined-but-unrooted pins master returned `self` for.

What I think the fix is: stop clearing, and instead sync onto an independent copy of the docstring -

```ruby
result.docstring = Solargraph::Source.parse_docstring(docstring.to_raw).to_docstring
```

That round-tripped faithfully for all four cases I tried, and `to_raw` is already used this way in this PR at `index.rb`. It costs a reparse per proxy, and proxy is hot in `Chain::Call`.

The cheaper variant is `docstring.dup` - but `dup` gives a fresh tags array holding the *same* `Tag` objects, so `sync_return_type_tag` would have to stop mutating in place and delete-then-add instead. That changes what happens when two or more return tags are present: today a third is appended, delete-and-add would collapse them to one. Your call on whether that is a change you want.

lib/solargraph/pin/base.rb
apiology: **Claude:** Reworded - no sigils, so there is nothing left to read as a tag:

```ruby
# Reads the return type instance variable, not the reader method:
# BaseVariable derives its reader from this tag, so it would sync
# the tag to itself. Callers force the reader first.
```

lib/solargraph/pin/base.rb
apiology: **Claude:** `tag_name` does have precedent, and it is uniform: all 19 uses of `tag_name` in `lib/` are YARD's `Tag#tag_name` - `tag1.tag_name == tag2.tag_name`, `ref.tag_name == 'param'`, `d.tag.tag_name == 'macro'`. Not one of them means a type string. The type-string sense lives entirely on `ComplexType`: `tag` (192), `tags` (74), `rooted_tag` (73), `rooted_tags` (23), `simple_tags` (1). So the two senses coexist, but they never share an identifier.

`type_tag_name` would break that: `:type` is one of the two values this returns, so the name reads as "the name of the `@type` tag" - wrong for methods and constants.

The real ambiguity you are pointing at is `return_type` sitting next to `tag`, since `return_type.tag` is a real call. Splitting those words fixes it, and `compare_docstring_tags` right below is precedent for the prefix: `docstring_return_tag_name`. 7 call sites, all private, all in `lib/` - say the word and I will do it.

lib/solargraph/pin/method.rb
apiology: **Claude:** It was reading the variable to test whether signatures had *already* been computed - `#signatures` memoizes with `@signatures ||=`, so calling it would generate them and the guard would always be true. `instance_variable_get` was not needed for that: `@signatures` is on `Pin::Method` itself and is always assigned in `initialize`, so a plain read works. Replaced with a named predicate:

```ruby
if (signatures_generated? || !inline_rbs.empty?) && docstring.tags(return_type_tag_name).empty? && return_type&.defined?

# True once #signatures holds a value; unlike calling it, this
# does not generate the signatures.
#
# @return [Boolean]
def signatures_generated?
  !@signatures.nil?
end
```

Typecheck output is identical before and after, so the indirection was not serving the typechecker.

lib/solargraph/api_map/index.rb
```diff
@@ -167,38 +167,123 @@ def map_references klass, hash
...
+      # @param pin [Pin::Base]
+      # @param ovr [Pin::Reference::Override]
+      # @return [void]
+      def apply_override pin, ovr
+        combined = pin.combine_with(override_pin_for(pin, ovr))
+        # combined belongs to us alone, so deleting from it touches nothing shared.
+        # @sg-ignore Wrong argument type for
+        #   YARD::Docstring#delete_tags: name expected String,
+        #   received String, Symbol - delete_tags is ok with a
+        #   _ToS, but we should fix anyway
```
apiology: sg-ignores in this repo are one single line, so they can be audited in rules.rb with grep.  Please see that file and choose an existing name or create a new one if need be

apiology: **Claude:**

Removed rather than renamed. `ovr.delete` holds Symbols and `YARD::Docstring#delete_tags` is typed `String`, so the marker was covering a real mismatch rather than a tool gap. `name.to_s` fixes it and no suppression is needed.

`rules.rb` has no catalogued reason covering this case, so the alternative would have been coining a new one for something that did not need suppressing at all.

Strong typecheck on the file is unchanged at 3 problems, all pre-existing `concat` calls at lines 134, 138 and 142.

lib/solargraph/pin/base.rb
```diff
@@ -746,6 +818,38 @@ def parse_comments
...
+        sync_return_type_tag
+      end
+
+      # Reads the return type instance variable, not the reader method:
+      # BaseVariable derives its reader from this tag, so it would sync
+      # the tag to itself. Callers force the reader first.
+      #
+      # @return [void]
+      def sync_return_type_tag
+        rooted_types = @return_type.items.map(&:rooted_tag)
```
apiology: These are tags, not types.   _type implies ComplexType/UniqueType

apiology: **Claude:** Renamed to `rooted_tags` in 0da4b54c1 - `rooted_tag` returns a `String`, not `ComplexType`/`UniqueType`, so `rooted_types` was misleading.

PR #1107, opened 2025-10-02
Reproduce solargraph-rspec issue

apiology:
I hit a failure in the solargraph-rspec specs while putting together a new merge branch.  It looks like it can be reproduced with the combination of:
* https://github.com/castwide/solargraph/pull/1087
* https://github.com/castwide/solargraph/pull/1100
* https://github.com/castwide/solargraph/pull/1103

Any two together do not reproduce it.


apiology:
OK, I believe that this is a false alarm - https://github.com/lekemula/solargraph-rspec/pull/26 is a spec-only change on the solargraph-rails side that should resolve it.  If @lekemula agrees, we can rerun the test workflow here once the PR is merged.

apiology:
@lekemula and I are closing in on a fix; I'll close this as it doesn't reflect an issue on the Solargraph side as far as I can see.

PR #1110, opened 2025-10-05
Annotation fixes for strong typechecking

apiology:
Includes to fix build:


apiology:
Already merged, minus a bit which is already baked into the needed PR.

PR #1112, opened 2025-10-08
CLI 'solargraph gems' improvements

apiology:
* Consolidate shell.rb's gem resolution behind the Workspace class so it can take advantage of improvements in other PRs (e.g., pulling info from the bundle directly to resolve gemspecs)
* Add documentation on 'solargraph gems'
* Allow request to cache core pins from 'solargraph cache' and 'solargraph gems'
* Ensure stdlib is cached in 'solargraph gems'

Depends on and includes:
* https://github.com/castwide/solargraph/pull/1064

To see a clean diff without #1064, see https://github.com/apiology/solargraph/pull/15


apiology:

PR #1113, opened 2025-10-09
Improve gemspec resolution logic

apiology:
Includes:
* Use RBS stdlib dependencies information to handle situations like YAML, which depends on Psych.
* Default to using the version of a gem from the workspaces' bundle.
* Refactors of the external bundle querying code
* Be careful to avoid relying on Gem::Specification methods for identity / sorting / hashing / etc - the 'bundler' Gem::Specification in particular seems to not have all the bells and whistles in some Ruby configurations.

Depends on and includes:
* https://github.com/castwide/solargraph/pull/1067

See https://github.com/apiology/solargraph/pull/10 for a more useful diff that doesn't include #1067

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -31,9 +31,8 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: 3.4
-          bundler: latest
           bundler-cache: true
-          cache-version: 2025-06-06
+          cache-version: 2026-01-11
 
       - name: Update to best available RBS
         run: |


lib/solargraph/pin/parameter.rb
```diff
@@ -200,7 +205,8 @@ def param_tag
       # @return [ComplexType]
       def typify_block_param api_map
         block_pin = closure
-        if block_pin.is_a?(Pin::Block) && block_pin.receiver
+        if block_pin.is_a?(Pin::Block) && block_pin.receiver && index
+          # @sg-ignore flow-sensivie typing should handle is_a? with &&
```
apiology: Fix spelling

apiology:

PR #1114, opened 2025-10-09
Better flow-sensitive typing

apiology:
Cases now at least partially handled:
* `if foo.nil?`
* `if foo`
* `unless foo.nil? || foo`
* `if unrelated && foo`
* `break if foo.nil?`
* `return foo unless !foo`
* `return if foo.nil?`
* `foo = nilable || not_nilable`
* `foo unless foo.nil? || foo`
* `foo if unrelated && foo`
* `while foo`
* `foo unless unrelated || repr.nil?`
* `break unless is_a?`
* `case ... when ... return if ...`

Includes for functionality:
* https://github.com/castwide/solargraph/pull/1119

Includes for convenience:

Includes for better testing:
* https://github.com/castwide/solargraph/pull/1087

Includes for buildability:

### DIFF
--- a/lib/solargraph/source/chain.rb
+++ b/lib/solargraph/source/chain.rb
@@ -271,7 +271,7 @@ module Solargraph
                else
                  ComplexType.new(types)
                end
-        if context.nil? || context.return_type.undefined?
+        if context.nil? || context.context.undefined?
           # up to downstream to resolve self type
           return type
         end
--- a/spec/type_checker/levels/alpha_spec.rb
+++ b/spec/type_checker/levels/alpha_spec.rb
@@ -93,5 +93,26 @@ describe Solargraph::TypeChecker do
 
       expect(checker.problems.map(&:message)).to eq([])
     end
+
+    it 'resolves self correctly in arguments' do
+      checker = type_checker(%(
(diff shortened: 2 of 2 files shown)


lib/solargraph/pin/if.rb
```diff
@@ -0,0 +1,18 @@
...
+      # @param node [Parser::AST::Node, nil]
+      # @param context [ComplexType, nil]
+      # @param args [::Array<Parameter>]
+      def initialize node: nil, **splat
+        super(**splat)
+        @node = node
+      end
+    end
+  end
+end
```
apiology: Remove - not used

spec/type_checker/levels/strict_spec.rb
```diff
@@ -61,7 +91,7 @@ def bar(a); end
         require 'kramdown-parser-gfm'
         Kramdown::Parser::GFM.undefined_call
       ), 'test.rb')
-      api_map = Solargraph::ApiMap.load_with_cache('.', $stdout)
+      api_map = Solargraph::ApiMap.load '.'
```
apiology: Why?

apiology:

PR #1119, opened 2025-11-01
Support intersection types for internal use

apiology:
Similar to the & type operator in TypeScript - see https://en.wikipedia.org/wiki/Intersection_type

This is a building block needed to go further with flow-sensitive typing, which in turn is a building block for union type enforcement (i.e., nil-checking)

Right now there's no YARD syntax and the RBS syntax isn't mapped in yet, but this is enough to hook into via flow-sensitive typing.

Includes for tests to pass:


apiology:

PR #1124, opened 2025-11-24
Fix type issue combining directives

apiology:
Solargraph found this in another branch while improving typechecking.  I'm not aware of any practical impact from the bug - I don't think we end up combining directives in practice.

Includes for a working build:


apiology:

PR #1125, opened 2025-11-27
Relax bundler dependency version constraint

hsbt:
This PR relaxes the Bundler version constraint in the gemspec from `~> 2.0` to `>= 2.0`.

The next version of bundler is 4.0. This constraint block to use that version.


castwide:
I'll include this in the next release. Should be no later than Jan 1.

PR #1126, opened 2025-11-27
Allow levels to be changed for typechecking rules in .solargraph.yml

apiology:
Includes to be able to build:

---

Suggested updates for https://solargraph.org/guides/configuration:

1) Change block in Defaults section to:

```
include:
- "**/*.rb"
exclude:
- spec/**/*
- test/**/*
- vendor/**/*
- ".bundle/**/*"
require: []
domains: []
reporters:
- rubocop
- require_not_found
formatter:
  rubocop:
    cops: safe
    except: []
    only: []
    extra_args: []
type_checker:
  rules: {}
max_files: 5000
```

2) Add section 'formatter' (hadn't been added)

Specify configuration for code formatters used in the LSP.  Currently
supports 'rubocop' as the subsection, with the following keys:

* `cops`: Valid values: 'all', 'safe'
* `except`: List of cops to pass to the --except argument in the RuboCop
  CLI
* `only`: List of cops to pass to the --only argument in the RuboCop
  CLI
* `extra_args`: Additional arguments to pass to the RuboCop CLI

3) Add section 'type_checker.rules'

Overrides the level at which an individual typecheck rule starts
reporting.

Each key is a rule name (see
[rules.rb](https://github.com/castwide/solargraph/blob/master/lib/solargraph/type_checker/rules.rb)
for the current list). Each value is a level name (`normal`, `typed`,
`strict`, `strong`, `alpha`) that the rule starts reporting at instead
of its default.

```
type_checker:
  rules:
    validate_calls: typed # make 'typed' level more strict without
                          # bringing in other rules
```


castwide:
:+1: :shipit:

PR #1127, opened 2025-11-27
Union type enforcement

apiology:
This flips on enforcement of union types - meaning we get typechecking violations when there is a potential nil dereference!

Depends on and includes:
* https://github.com/castwide/solargraph/pull/803
* https://github.com/castwide/solargraph/pull/928
* https://github.com/castwide/solargraph/pull/1012
* https://github.com/castwide/solargraph/pull/1050
* https://github.com/castwide/solargraph/pull/1005
* https://github.com/castwide/solargraph/pull/1059
* https://github.com/castwide/solargraph/pull/1089
* https://github.com/castwide/solargraph/pull/1098
* https://github.com/castwide/solargraph/pull/1097
* https://github.com/castwide/solargraph/pull/1096
* https://github.com/castwide/solargraph/pull/1094
* https://github.com/castwide/solargraph/pull/1102
* https://github.com/castwide/solargraph/pull/1116
* https://github.com/castwide/solargraph/pull/1123
* https://github.com/castwide/solargraph/pull/1140

PR #1128, opened 2025-11-27
Add ivar support to flow sensitive typing

apiology:
Adds support for doing flow-sensitive typing in instance variables, not just local variables.

Note: This approach assumes single-threaded use for type-checking purposes - it's possible to be clever and subvert it, but as a way to reduce false positives in nil-checking, it seems very effective.

To see the actual diff in advance of merging the below, check out https://github.com/apiology/solargraph/pull/19

Builds on:
* https://github.com/castwide/solargraph/pull/1127


apiology:

PR #1139, opened 2026-01-01
Test plugins with v0.58.0

castwide:
Bumping the gem version broke the solargraph-rails tests in CI. This is a temporary stopgap to make sure the tests still pass before release.

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -35,6 +35,8 @@ jobs:
       run: |
         echo 'gem "solargraph-rails"' > .Gemfile
         echo 'gem "solargraph-rspec"' >> .Gemfile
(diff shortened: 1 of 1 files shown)


apiology:
Verified in a dummy PR that this should pass on master now:

https://github.com/apiology/solargraph/pull/22

I lifted the upper bound entirely in solargraph-rails.

apiology:
...except I merged the latest castwide/master in and now it's not working lol

Looking here

castwide:
@apiology Thanks, I was just about to call attention to that. I don't see anything obvious in #1126 that would cause these particular tests to fail.

apiology:
This should fix - will merge once the tests pass: https://github.com/iftheshoefritz/solargraph-rails/pull/186

PR #1140, opened 2026-01-02
Improve signature combination

apiology:
Use our generated RBS signature from parameters as a key to combine method signatures from RBS/YARD pins.

This is closer to what RBS does than the current technique of using the arity alone, and fixes a key degenerate case in Integer#+ revealed by updated definitions used by recently released RBS gems.  

I saw this issue when updating PR from master; after merging this code in, it goes away:
* #1127 

Includes for convenience:
* #1102


apiology:

PR #1149, opened 2026-01-11
Use correct field for self type resolution

apiology:
Add a regression test and fix for self type resolution issue found on a future branch


apiology:

PR #1151, opened 2026-01-11
Don't log caching for each dependent library

apiology:
This fixes an integration issue between different branches.

We are now logging things entirely in PinCache and don't need to log additionally in RbsMap::StdlibMap, much less for each dependent library as well as the one being cached.

This is based on the 2026-01-06 branch.


apiology:

PR #1158, opened 2026-01-15
Spec performance fixes

apiology:
* Avoid building all gem pins during testing
* Cache installed gems (but not the solargraph cache)
* Split out longer CI checks into smaller individual checks run in parallel

Builds on v0.59 branch


spec/yard_map/mapper_spec.rb
```diff
@@ -32,17 +32,6 @@
...
-    # Using rubocop because it's a known dependency
-    rubocop = Gem::Specification.find_by_name('rubocop')
-    Solargraph::Yardoc.cache([], rubocop)
-    Solargraph::Yardoc.load!(rubocop)
-    pins = Solargraph::YardMap::Mapper.new(YARD::Registry.all).map
-    pins = pins.select { |pin| pin.path == 'RuboCop::Options.new' }
-    expect(pins.map(&:return_type).uniq.map(&:to_s)).to eq(['self'])
-    expect(pins.flat_map(&:signatures).map(&:return_type).uniq.map(&:to_s)).to eq(['self'])
-  end
-
```
apiology, line 46: This is slow and doesn't describe how it tests anything different than the spec above it

lib/solargraph/api_map/constants.rb
```diff
@@ -188,12 +188,12 @@ def collect_and_cache gates
...
       end
 
-      # @return [Hash{Array(String, Array<String>) => String, :in_process, nil}]
+      # @return [Hash{Array, String, Array, String => String, :in_process, nil}]
       def cached_resolve
         @cached_resolve ||= {}
       end
 
-      # @return [Hash{Array<String> => Array<Solargraph::Pin::Namespace, Solargraph::Pin::Constant>}]
+      # @return [Hash{Array, String => Array<Solargraph::Pin::Namespace, Solargraph::Pin::Constant>}]
```
apiology: This pattern of changes was brought in by a buggy version of YARD and the rubocop-yard autocorrect - please update both and revert the mistaken changes.

apiology: Fixed in c789c8f3c. Root cause: these two YARD Hash tags were correct on both merge parents, but got mangled mid-merge by an errant RuboCop autocorrect run (YARD/CollectionStyle). Tested rubocop-yard 1.0.0 through 1.3.0 (latest) — the autocorrect bug reproduces on all of them (turns `Hash{Array(String, Array<String>) => String, ...}` into invalid `Hash{Array, String, Array, String => String, ...}`), so a version bump doesn't fix it. Reverted both tags to correct syntax and disabled YARD/CollectionStyle in .rubocop.yml so autocorrect can't silently re-mangle them.

spec/yard_map/mapper_spec.rb
apiology: Removed in c789c8f3c.

lib/solargraph/api_map/constants.rb
apiology: Found more instances of the same bug, fixed in f89c19190: lib/solargraph/doc_map.rb (6 occurrences) and lib/solargraph/api_map/store.rb (2 occurrences) had the same Array(...) tuple parens stripped by the merge-time autocorrect. Confirmed all now match castwide/v0.59's original correct syntax exactly (diffed to verify zero remaining differences), and re-ran rubocop + affected specs clean.

apiology:
Closing in favor of https://github.com/castwide/solargraph/pull/1237, which targets `master` instead of `v0.59`.

PR #1162, opened 2026-01-18
Refactor type hierarchy

apiology:
Let's see if we can make room for a first-class intersection type


apiology:
Closing in favor of #1244, which ports this refactor onto current master. The `type_hierarchy` branch had diverged too far (~1300 commits) from its base for a direct rebase to be practical; #1244 re-derives just the intentional change (extracting `Solargraph::Type` from the `TypeMethods` mixin) against current master, with tests/rubocop/typecheck matching the master baseline.

—Claude

PR #1163, opened 2026-01-18
Refactor RbsMap::Conversions

apiology:
Let's work more directly with type objects and reduce duplication

Includes 
* #1164


apiology:

PR #1164, opened 2026-01-27
Fix solargraph-rspec spec failure

apiology:
Includes to reproduce the problem:
* #1165


apiology:

PR #1166, opened 2026-01-29
Test verbosity fix

apiology:
Includes for a working build:
* https://github.com/castwide/solargraph/pull/1168

PR #1167, opened 2026-01-29
Run specs in parallel, speed up CI, fix concurrency bugs in specs and code

apiology:
This allows rspec to use more than one CPU.  Really useful on modern dev machines, and also useful in CI, where the default runner size has two CPUs.

See https://github.com/apiology/solargraph/pull/27/changes for a more useful diff and commentary

To do:
- [ ] Track down slow tests and see where time is going
- [ ] Track down rspec + rails specs slowness
- [ ] Add developer doc on things to expect, how to write specs
- [ ] Fix logging caching gems
- [ ] Figure out where doc_map vs workspace fix went

Reduces the actual testing time in CI to about a minute using our small workers:

image

There are also a few fixes here:
* For some reason, the debug level logging I tried to set in rspec wasn't persisting - this seems to do a better job and found a couple of bugs in the process that only reproduce at debug level logging.
* Some tests didn't run individually, but did when run in sequence in the specs.  We may see new failures pop up in the future; the trade-off for faster responsiveness is probably going to be less deterministic behavior in the face of buggy tests.

Includes for mergability:
* #1165 

Includes this build fix:
* https://github.com/castwide/solargraph/pull/1168

Includes the fix for this bug, which parallel running seems to trigger (more often?):
* https://github.com/castwide/solargraph/pull/1166

Includes to show combined effect:
* https://github.com/castwide/solargraph/pull/1158

Includes transitively:
* v0.59 branch


.github/workflows/linting.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: 3.4
-        bundler-cache: false
+        bundler-cache: true
```
apiology, line 106: We're already doing this in the tests for: 
* #1168 

This keeps linting from being the long pole in the build and frees up workers faster for other checks.

.github/workflows/plugins.yml
```diff
@@ -104,9 +164,36 @@ jobs:
...
+      run: |
+        bundle exec solargraph config
+        yq -yi '.plugins += ["solargraph-rspec"]' .solargraph.yml
+    - name: Install gem types
+      run: bundle exec rbs collection update
+    - name: Ensure specs still run
+      run: |
+        PARALLEL=$(nproc --all)
+        export PARALLEL
+        bundle exec rake spec
```
apiology: These were doing both specs and typechecking in the same check, which became the long pole for the build.  Split them in two so they can run in parallel, at least when there are worker slots available.

.github/workflows/plugins.yml
```diff
@@ -217,6 +306,8 @@ jobs:
         bundle info solargraph
         bundle info rbs
         bundle info yard
+        PARALLEL=$(nproc --all)
+        export PARALLEL
```
apiology: `nproc -all` returns 4 for the default open source GitHub action workers, which is the default anyway for `parallel_rspec`, but this adds future proofing

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -313,8 +312,8 @@ def to_rbs
           'bool'
         elsif name.downcase == 'nil'
           'nil'
-        elsif name == GENERIC_TAG_NAME
-          all_params.first&.name
+        elsif name == GENERIC_TAG_NAME && !all_params.empty?
+          all_params.first.name
```
apiology: Leave the RBS represention as `"generic"` rather than `""`, which would be what the nil would turn into.  This is bad input regardless.

lib/solargraph/diagnostics/type_check.rb
```diff
@@ -7,12 +7,12 @@ module Diagnostics
...
     class TypeCheck < Base
       # @return [Array<Hash>]
-      def diagnose source, api_map
+      def diagnose source, api_map, workspace: nil
         # return [] unless args.include?('always') || api_map.workspaced?(source.filename)
         severity = Diagnostics::Severities::ERROR
         level = (args.reverse.find { |a| ['normal', 'typed', 'strict', 'strong'].include?(a) }) || :normal
         # @sg-ignore sensitive typing needs to handle || on nil types
-        checker = Solargraph::TypeChecker.new(source.filename, api_map: api_map, level: level.to_sym)
+        checker = Solargraph::TypeChecker.new(source.filename, api_map: api_map, level: level.to_sym, workspace: workspace)
```
apiology: Helps speed up a slow spec by being able to configure the workspace used.

lib/solargraph/api_map.rb
```diff
@@ -760,6 +760,13 @@ def qualify_superclass fq_sub_tag
       store.qualify_superclass fq_sub_tag
     end
 
+    # @param require_path [String]
+    #
+    # @return [Array<Gem::Specification>, nil]
+    def resolve_require require_path
+      workspace.resolve_require require_path
+    end
```
apiology, line 764: Useful to reach this method from specs, and seems a reasonable thing to export here, given we already export some DocMap Gem::Sepcification methods.

spec/pin/base_spec.rb
```diff
@@ -50,8 +50,14 @@
...
-    Solargraph::Shell.new.uncache('yard')
-    api_map = Solargraph::ApiMap.load_with_cache('.', $stderr)
+    # if this fails you might not have an rbs collection installed
+    api_map = Solargraph::ApiMap.load ''
+
+    spec = Gem::Specification.find_by_name('yard')
+    api_map.cache_gem(spec)
+
+    bench = Solargraph::Bench.new(external_requires: ['yard'])
+    api_map.catalog bench
```
apiology, line 62: Not exactly sure why we were uncaching yard, but avoid slow specs by only caching what we need.

spec/pin/method_spec.rb
```diff
@@ -554,7 +554,9 @@ class Foo
       # on type.  Let's make sure we combine those with anything else
       # found (e.g., additions from the BigDecimal RBS collection)
       # without collapsing signatures
-      api_map = Solargraph::ApiMap.load_with_cache(Dir.pwd, nil)
+      api_map = Solargraph::ApiMap.load(Dir.pwd)
+      bench = Solargraph::Bench.new external_requires: ['bigdecimal']
+      api_map.catalog(bench)
```
apiology: Only ask to cache what we need to speed up specs - goal here is to never do the full build of all solargraph gems, which adds time.

spec/source/chain_spec.rb
```diff
@@ -1,12 +1,12 @@
...
   it "gets empty definitions for undefined links" do
     chain = described_class.new([Solargraph::Source::Chain::Link.new])
-    expect(chain.define(nil, nil, nil)).to be_empty
+    expect(chain.define(nil, nil, [])).to be_empty
   end
 
   it "infers undefined types for undefined links" do
     chain = described_class.new([Solargraph::Source::Chain::Link.new])
-    expect(chain.infer(nil, nil, nil)).to be_undefined
+    expect(chain.infer(nil, nil, [])).to be_undefined
```
apiology, line 9: Passing nil here didn't match our type annotations and also triggered a newly-found nil violation in the debug logs

.envrc
```diff
@@ -1,3 +1,10 @@
+#!/bin/bash
+
+if [[ "$(uname)" == "Darwin" ]]; then
+  WORKERS=$(sysctl -n hw.physicalcpu)
+  export WORKERS
+fi
```
apiology, line 6: Useful for macOS direnv users like me to make sure we use the right number of parallel workers

spec/spec_helper.rb
```diff
@@ -21,16 +22,38 @@
...
+
 RSpec.configure do |c|
   # Allow use of --only-failures with rspec, handy for local development
   c.example_status_persistence_file_path = 'rspec-examples.txt'
 end
+RspecTimeGuard.setup
+RspecTimeGuard.configure do |config|
+  config.global_time_limit_seconds = 120
+  config.continue_on_timeout = false
+end
```
apiology, line 51: Some concurrency issues I hit manifested as deadlock - this helps ensure in CI that we fail the spec rather than let it hang the worker until the worker times out, which is a multiple-hour process.

spec/spec_helper.rb
```diff
@@ -21,16 +22,38 @@
     enable_coverage(:branch) if ENV['SOLARGRAPH_BRANCH_COVERAGE']
   end
 end
+PROJECT_DIRECTORY = File.expand_path('..', __dir__)
```
apiology, line 28: We have a tests that reference `.` as the project directory.  Given we know we have `chdir`s happen in Solargraph, we should start using this instead.  Turns out  the failures that drove me to this were not caused by chdir, but it did eliminate the possible cause and seems like a good practice nonetheless.

spec/spec_helper.rb
```diff
@@ -21,16 +22,38 @@
...
+  end
+end
+
+set_logging
+require 'parallel_rspec'
+
+ParallelRSpec.configure do |config|
+  config.after_fork do |worker_number|
+    warn "ParallelRSpec worker #{worker_number} starting in pid #{Process.pid}"
+    set_logging
```
apiology: I don't know why putting this here reproduced more nil issues in debug statements.  My guess is that something in the specs or code turns off debug logging partway through the specs, but I don't know where.

spec/yardoc_spec.rb
```diff
@@ -62,9 +62,18 @@
...
+        # write the complete file to simulate successful run
+        FileUtils.mkdir_p(gem_yardoc_path)
+        FileUtils.touch(File.join(gem_yardoc_path, 'complete'))
+      end
+
       described_class.build_docs(gem_yardoc_path, [], gemspec)
       described_class.build_docs(gem_yardoc_path, [], gemspec) # second time
-      expect(File.exist?(File.join(gem_yardoc_path, 'complete'))).to be true
+
+      expect(Open3).to have_received(:capture2e).once
```
apiology, line 76: Mock more to speed things up.

spec/yard_map/mapper_spec.rb
```diff
@@ -1,11 +1,14 @@
...
+  # directory
+  before :context do
     @api_map = Solargraph::ApiMap.load('.')
   end
 
   def pins_with require
-    doc_map = Solargraph::DocMap.new([require], @api_map.workspace, out: nil)
-    doc_map.cache_doc_map_gems!(nil)
+    doc_map = Solargraph::DocMap.new([require], @api_map.workspace, out: $stderr)
+    doc_map.cache_doc_map_gems!($stderr)
```
apiology, line 13: I added this logging to find spots where we were doing expensive gem caching multiple times because the needed specs were running in different workers.

Not in love with having logging pop out during specs, and willing to turn it off if. you prefer, but it did help.

spec/language_server/protocol_spec.rb
```diff
@@ -432,9 +432,15 @@ def bar baz
...
   it "handles $/solargraph/documentGems" do
+    status = instance_double(Process::Status)
+    allow(status).to receive(:==).with(0).and_return(true)
+    allow(Open3).to receive(:capture2).with('solargraph', 'gems').and_return(['', status])
+
     @protocol.request '$/solargraph/documentGems', {}
     response = @protocol.response
+
     expect(response['error']).to be_nil
+    expect(Open3).to have_received(:capture2).with('solargraph', 'gems')
```
apiology, line 464: Avoid another case of complete gem caching

spec/language_server/message/extended/check_gem_version_spec.rb
```diff
@@ -42,13 +46,12 @@
...
-        "result" => response['params']['actions'].first
-      }
-      host.receive action
-    }.not_to raise_error
+    action = {
+      "id" => response['id'],
+      "result" => response['params']['actions'].first
+    }
+    host.receive action
+    expect(Open3).to have_received(:capture2).with('gem update solargraph')
```
apiology, line 59: Make sure we don't somehow mess with the solargraph gem while testing it :)

lib/solargraph/source/chain.rb
```diff
@@ -166,7 +167,13 @@ def infer_uncached api_map, name_pin, locals
...
         type = infer_from_definitions(pins, links.last.last_context, api_map, locals)
         out = maybe_nil(type)
-        logger.debug { "Chain#infer_uncached(links=#{self.links.map(&:desc)}, locals=#{locals.map(&:desc)}, name_pin=#{name_pin}, name_pin.closure=#{name_pin.closure.inspect}, name_pin.binder=#{name_pin.binder}) => #{out.rooted_tags.inspect}" }
+        logger.debug do
+          "Chain#infer_uncached(links=#{self.links.map(&:desc)}, " \
+            "locals=#{locals.map(&:desc)}, " \
+            "name_pin=#{name_pin}, " \
+            "name_pin.closure=#{name_pin&.closure.inspect}, " \
+            "name_pin.binder=#{name_pin&.binder}) => #{out.rooted_tags.inspect}"
+        end
```
apiology, line 179: See notes in spec_helper - looks like debug logging works more consistently with the new setup here, reproducing type issues in debug logging as intended more often (not sure that it's perfect yet for sure!).

lib/solargraph/yardoc.rb
```diff
@@ -18,6 +18,21 @@ module Yardoc
...
+        # Wait for up to 60 seconds for another process to finish building
+        timeout = 60
+        start_time = Time.now
+
+        sleep 1 until docs_built?(gem_yardoc_path) || (Time.now - start_time > timeout)
+
+        return if docs_built?(gem_yardoc_path)
+
+        raise "YARD doc build for #{gemspec.name} #{gemspec.version} did not finish in #{timeout} seconds"
+      end
```
apiology, line 34: I saw issues in testing this where we started YARD docs before they were ready - presumably because the caching was started in another process but not done.

spec/language_server/host/dispatch_spec.rb
```diff
@@ -1,5 +1,5 @@
 describe Solargraph::LanguageServer::Host::Dispatch do
-  before :all do
+  before :context do
```
apiology: parallel_rspec is smart enough to not try to parallelize when it sees once-per-context setup like this.

It triggers off of :context, not :all, even those they are synonyms to rspec.

- [ ] Do I need this as :context?

spec/language_server/message/text_document/rename_spec.rb
```diff
@@ -1,10 +1,14 @@
 # frozen_string_literal: true
 
 describe Solargraph::LanguageServer::Message::TextDocument::Rename do
+  let(:temp_file_url) do
+    "file:///#{Dir.mktmpdir}/file.rb"
+  end
```
apiology, line 7: - [ ] Is there something we can mock instead to avoid conflict in the below specs?

spec/rbs_map/conversions_spec.rb
```diff
@@ -94,13 +94,23 @@ def bar: () -> untyped
   end
 
   context 'with standard loads for solargraph project' do
-    before :all do # rubocop:disable RSpec/BeforeAfterAll
-      @api_map = Solargraph::ApiMap.load_with_cache('.')
+    before :context do
```
apiology:
```suggestion
    # Use :context here instead of :all so that parallel_rspec runs these on the same worker and we only have to cache these gems on one worker
    before :context do
```

spec/doc_map_spec.rb
```diff
@@ -124,7 +124,9 @@
   context 'with require as bundle/require' do
     it 'imports all gems when bundler/require used' do
       doc_map_with_bundler_require = described_class.new(['bundler/require'], workspace, out: nil)
-      doc_map_with_bundler_require.cache_doc_map_gems!(nil)
+      if doc_map_with_bundler_require.pins.length <= plain_doc_map.pins.length
+        doc_map_with_bundler_require.cache_doc_map_gems!(nil)
+      end
```
apiology, line 131: Almost always we'll already have cached a gem or two by here, allowing us to avoid caching any more.

apiology:
Superseded by https://github.com/apiology/solargraph/pull/44, which carries this work forward rebased on top of castwide/solargraph#1237.

PR #1168, opened 2026-01-31
Work around erb bundle problem with RBS pre-release

apiology:
https://github.com/castwide/solargraph/actions/runs/21547640233/job/62091456310?pr=1168 is now happening with the latest RBS pre-release, which seems to be reproducing:
* https://github.com/ruby/rbs/pull/2833

Removing some of our previous bundler workarounds and using the setup-ruby action to install the bundler both seems to work around the issue as well as give us faster builds by caching the bundle install step.


apiology:

PR #1171, opened 2026-02-28
Remove Vernier

castwide:
Remove the Vernier dependency due to Windows incompatibility. I realize that Windows is a minority use case, but I'm afraid I require it.


apiology:
If @lekemula doesn't beat me to it, I may try to propose a way to keep this capability handy on non-Windows platform.

castwide:
I'm open to that. My number one concern is that the gem can still be installed on Windows.

lekemula:
@castwide, would @apiology's [suggestion](https://github.com/castwide/solargraph/pull/1071#issuecomment-3283079192) be an option?

castwide:
> @castwide, would @apiology's [suggestion](https://github.com/castwide/solargraph/pull/1071#issuecomment-3283079192) be an option?

Yes, that seems reasonable.

castwide:
Superceded by #1172.

PR #1172, opened 2026-03-02
Allow vernier gem to be used optionally, but don't require it

apiology:
Ensures solargraph can still be used on the Microsoft Windows platform

### DIFF
--- a/lib/solargraph/shell.rb
+++ b/lib/solargraph/shell.rb
@@ -376,7 +376,7 @@ module Solargraph
       begin
         require 'vernier'
       rescue LoadError
-        $stderr.puts 'vernier gem not found. Add this line to your .Gemfile and run `bundle install` to use this command:'
+        $stderr.puts 'vernier gem not found. Please install this dependency:'
         $stderr.puts
         $stderr.puts "  gem 'vernier', '>1.0', '<2'"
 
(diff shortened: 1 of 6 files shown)


lib/solargraph/shell.rb
```diff
@@ -376,7 +376,10 @@ def profile file = nil
       begin
         require 'vernier'
       rescue LoadError
-        warn 'vernier gem not found. Install with: gem install vernier'
+        $stderr.puts 'vernier gem not found. Add this line to your .Gemfile and run `bundle install` to use this command:'
```
castwide: Since `.Gemfile` is nonstandard, this notification might be confusing to users who attempt to profile Solargraph in a different project where it doesn't exist. They'd need to have both `solargraph` and `vermier` in their `Gemfile`. If they're not using bundler, they should only need to have `vermier` installed at the system level.

castwide: Another option that was previously discussed, we could hide the command.

apiology: Yeah, I wasn't thinking about using this on anything but solargraph :)

apiology:
```suggestion
        $stderr.puts 'vernier gem not found. Please install this dependency:'
```

apiology: Rephrased a little more generically - folks who want to profile can figure this out :)

castwide:
LGTM. Confirmed installable on Windows. Thanks for collaborating on this, @lekemula and @apiology!

Issue #1121, opened 2025-11-11
Support for inline RBS syntax


connorshea:
It'd be great to be able to use the inline RBS syntax that is in the RBS gem and is a standardized included-in-the-language alternative to YARD: https://github.com/ruby/rbs/blob/master/docs/inline.md#inline-rbs-type-declaration

It is supported already by Ruby LSP and Sorbet to an extent https://github.com/sorbet/sorbet/pull/8470

Both for the typechecking/indexing and just generally to make it easier to define types in a standard way across all Ruby code.

apiology:
Yes!  I'd love to see this too.  My understanding is that the upcoming version 4 of the RBS gem will provide some level of support - maybe we can build on top of that?

castwide:
The first iteration of [support for inline RBS](https://github.com/castwide/solargraph/pull/1173) is released in version 0.60.0.

PR #1173, opened 2026-03-04
Support for Inline RBS

castwide:
ref #1121


castwide:
This iteration supports two syntaxes:

```ruby
# (short) method signatures

#: (Integer) -> String
def example(number); end

# class inheritance parameters

class Example < Array #[String]
end
```

PR #1175, opened 2026-03-11
Type alias reference pins

castwide:
Create reference pins for RBS type aliases to improve type inference.


castwide:
Superseded by #1181.

PR #1178, opened 2026-03-13
Handle runtime without bundle

castwide:
`Workspace::Gemspecs` throws errors if it tries to find required gemspecs without a configured bundle. This can occur in any use case where the `solargraph` executable is not running in a bundled environment, regardless of whether the workspace itself has a bundler configuration.


castwide:
Superceded by #1180.

PR #1185, opened 2026-04-30
Fix signatureHelp bug that has existed in Solargraph since version 0.54.1:

meilinxiaoxue:
When autocompleting a function and typing a parenthesis triggers signatureHelp, the feature fails (no signatureHelp is shown) because the parenthesis is not closed.

for example:
In gvim ,use solargraph via coc-nvim and coc-solargraph("^"represent cursor location):
foo(^      //no signature help shows
foo(^)    //signature shows

 After the fix, signatureHelp should be displayed even when the parenthesis is not closed.


castwide:
Looks good, thanks!

PR #1189, opened 2026-05-17
Macro fixes

castwide:
This PR makes some changes to how macro directives get cataloged in maps.

* `SourceMap` tracks macros defined in workspace code.
* `ApiMap` processes macros after cataloging the workspace so they can resolve namespaces and paths.
* These changes supersede the need to map ephemeral pins.
* Macros only get applied to method calls that resolve to the specific method where the macro was defined.
* Tags without nesting get expanded along with directives.


spec/api_map_spec.rb
```diff
@@ -971,27 +971,27 @@ def self.property(name, ret_type, docstring)
...
         # @!macro prop
         #   @!parse
         #     module SomeNamespace
-        #       # @return [$6]
+        #       # @return [$3]
         #       def self.$1(value)
         #       end
         #     end
-        def self.property(name, default, type:, comment:)
+        def self.property(name, default, type, comment)
```
lekemula: @castwide I think this change breaks the idea of the spec:

> expands keyword arguments as flat array of params from DSL methods with attached macros

This was sort of my "invention" of providing support for keyword arguments in macros, by treating them as an ordered params. It's not ideal, because keyword args are meant to be called out of order, but this was the simplest way we could at least have something as a start. 

I couldn't find any specification on YARD side about keyword arguments.

That said, I can understand why you could have been confused or might not like it :D. If the latter, then we should change the spec description at the very least.

lekemula:
> These changes supersede the need to map ephemeral pins.

Cool that you came up with a way to avoid the ephemeral pins, I was not a big fan of that either! 🚀

spec/api_map_spec.rb
castwide: I modified the spec so it still tests with keyword arguments and added some `@todo` comments that explain why the expectations are a little wonky.

PR #1190, opened 2026-05-17
Specs Pending on Macro Changes

castwide:
Mark a couple of failing specs as pending while we iterate on the `yard-macros` branch.


castwide:
This PR isn't necessary at the moment, as the changes that break the specs haven't been merged yet. If they need to be marked pending, I'll do it in the PR that affected them.

PR #1194, opened 2026-05-18
Merge YARD macro support

castwide:


spec/fixtures/gem-with-yard-macros/gem-with-yard-macros.gemspec
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative "lib/gem/with/yard/macros/version"
+
+Gem::Specification.new do |spec|
+  spec.name = "gem-with-yard-macros"
+  spec.version = Gem::With::Yard::Macros::VERSION
+  spec.authors = ["Lekë Mula"]
+  spec.email = ["l.mula@finlink.de"]
```
lekemula, line 9:
```suggestion
  spec.authors = ["Lekë Mula"]
  spec.email = ["leke.mula@gmail.com"]
```

or, if you'd prefer dummy creds

```suggestion
  spec.authors = ["John Doe"]
  spec.email = ["john.doe@email.com"]
```

I don't mind either.

castwide: First suggestion is fine with me. If you prefer otherwise, just let me know.

spec/source_map/clip_spec.rb
```diff
@@ -274,6 +274,7 @@ def baz; end
   end
 
   it 'infers types from named macros' do
+    pending 'WIP'
```
lekemula: So I took a look into this one. It seems like we need to account for named macros that are not necessarily attached to methods, right? 
```
  macro_pins = pins_by_class(Pin::Method).select { |pin| pin.macros.any? }
```

Again, my brain takes me to an "additional pin" solution captured somewhere in `lib/solargraph/yard_map/directives.rb`, but I know you probably have better ideas on your mind (?). So if you could give some guidance, I can tackle this one.

lekemula:
Hey @castwide, I tried to test this today on my project and I found something strange. The LSP was stuck in a seemingly endless loop, and it was never finishing cataloging.

These are the last 500lines of the logs, which seem to be kept looping over and over again:

[solargraph_main.log](https://github.com/user-attachments/files/28397989/solargraph_main.log)

One thing that caught my eye, in the whole river of the logs, was the `ApiMap#add_methods_from_reference(type=Kernel)` repetition.

This issue does not seem to appear in the `v0.59.2`. Everything worked fine once I rolled back.

I hope this helps somehow. 🤞 Let me know if I can provide you with further details.

castwide:
@lekemula Did you encounter this with the `solargraph-rspec` project? I tried in VS Code with `solargraph-rspec` on the `main` branch running `solargraph` on the `yard-macros` branch. I didn't encounter an endless loop, but I did notice the following:

* It had to document 115 gems because I cleared the cache. This took a while, but the UI worked while the gems got mapped iteratively.
* After documenting the 115 gems, the progress status got stuck at "Mapping workspace: 25/28 files." After I closed and reopened the editor, it finished mapping and appeared to work normally.
* The UI was slow but responsive.

Is there a particular operation you attempted that triggered the endless loop? If you encountered the problem on a different project, can you provide a reproducible example?

lekemula:
@castwide, sorry for the ambiguity, but I meant the closed-source project at my current job.

So I ran the `solargraph profile`, and that revealed that the issue is indeed `ApiMap#process_macros` taking too long to process:

image


[v0.59.2_definition_benchmark.json.gz](https://github.com/user-attachments/files/28410447/definition_benchmark.json.gz)

[main_definition_benchmark.json.gz](https://github.com/user-attachments/files/28410458/definition_benchmark.json.gz)

Note: The process on `main` was manually interrupted after some time, so the time there could be much longer.

Here are some stats regarding our project:

```
[ERROR][2026-05-29 23:26:17] "rpc"	"solargraph"	"stderr"	"[DEBUG] ApiMap#process_macros: processing macros for 7838 source maps\n"
[ERROR][2026-05-29 23:26:17] "rpc"	"solargraph"	"stderr"	"[DEBUG] ApiMap#process_macros: store has 198 macro method namepins\n"
```

I fear that the **exact** DSL methods inference might be a little too expensive an operation. Maybe we should go with a more "naive approach" for the sake of performance?

---- 

SIDE NOTE:  regarding `solargraph profile`: I noticed that the bulk of the work of cataloging is now on `definition_benchmark.json.gz` (due to `Library#sync_catalog`) instead of `Host#catalog` which the command currently uses for `catalog_benchmark.json.gz` part. That's the reason I attached them here. We should probably adapt the command accordingly.

```
=== Timing Results ===
Parsing & mapping: 32068.54ms
Catalog building: 2.33ms

Profiles saved to:
  - /Users/lekemula/Projects/finlink/loanlink-api/tmp/profiles-main/parse_benchmark.json.gz
  - /Users/lekemula/Projects/finlink/loanlink-api/tmp/profiles-main/catalog_benchmark.json.gz
  - /Users/lekemula/Projects/finlink/loanlink-api/tmp/profiles-main/definition_benchmark.json.gz

```

castwide:
@lekemula Thanks for clarifying. It looks like my attempt to minimize the amount of macro processing performed during catalog operations is insufficient for very large codebases.

> I fear that the exact DSL methods inference might be a little too expensive an operation. Maybe we should go with a more "naive approach" for the sake of performance?

I tend to agree.

Would you like to profile your solution in #1202? If its performance is acceptable, we can merge that instead. AFAICT, the worst case scenario is that the language server might surface some incompletely inferred macros as `undefined` instead of something more useful. I think I'd prefer that minor wart to a critical performance bottleneck. Hopefully I'll be able to address general improvements to type expansion in #1204.

lekemula:
> Would you like to profile your solution in #1202? If its performance is acceptable, we can merge that instead.

@castwide Unfortunately, the #1202 showed no difference. However, while removing the lines that seem to be causing the performance bottleneck based on vernier traces, the performance improved, and there were no failing specs (at least locally). 

https://github.com/castwide/solargraph/pull/1207

I'm wondering what that call was added for?

PR #1197, opened 2026-05-18
Convert RBS implicit nil annotations

castwide:
Check for the `implicitly-returns-nil` annotation in RBS overloads and adjust the return types accordingly.


castwide:
This PR exposed a slew of strong typecheck errors based on doc maps that didn't take implicit nil into account. It's a big enough task to assign to separate issues.

castwide:
~~One major source of false typecheck errors: the typechecker is treating calls to nil values as unresolved when `NilClass` can respond to them, e.g., `#is_a?` and `#class`.~~ It's a little more subtle than that. Undefined variables don't respond to `NilClass` methods, which is actually correct at the strong level. The good news is that multiple cascading errors can be fixed by correcting the variable assignment.

PR #1199, opened 2026-05-19
Cache generated maps in rspec workflows

castwide:
Maintain a cache for documentation across the rspec workflow. This should eliminate the need to regenerate YARD documentation for each run.


castwide:
Not working. I'll revisit this issue later.

PR #1207, opened 2026-06-01
Macros Performance - Remove redundant chain.define (?)

lekemula:
This seems to be causing the performance issues reported https://github.com/castwide/solargraph/pull/1194#issuecomment-4580077778

After removing these lines, specs still were passing locally, and I could not see the performance degradation as reported in the comment above.

image

image

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -150,11 +150,6 @@ module Solargraph
               next
             end
           end
-          pin = chain.define(self, closure, []).first
-          next unless pin&.macros&.any?
-          pin.macros.each do |macro|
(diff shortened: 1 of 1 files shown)


lib/solargraph/api_map.rb
```diff
@@ -150,11 +150,6 @@ def process_macros
               next
             end
           end
-          pin = chain.define(self, closure, []).first
```
lekemula, line 153: @castwide do you remember the original reason this was added for?

castwide: Good catch! That was the original routine I tested for resolving method calls to pins. The newer routine at line 142 does the same thing without the expense of a chain definition. I just neglected to delete the original.

castwide:
Thanks!

PR #1219, opened 2026-07-28
Typecheck attribute return tags against their backing ivar's type

apiology:
## Summary

`TypeChecker#method_return_type_problems_for` skips the declared-vs-inferred
comparison whenever `pin.node.nil?`. `attr_reader`/`attr_writer`/`attr_accessor`
pins are built without a `node:` (`process_attribute` in `send_node.rb`), so
a manually-written `@return` tag on any attribute has never been checked
against how its ivar is actually assigned, at any typecheck level.

`Pin::Method#probe` already infers a type for attributes via `infer_from_iv`
(used elsewhere in this file for the "missing tag" branch); this PR reuses it
for the declared-vs-inferred comparison. Attributes with no locally-discoverable
ivar assignment (mixin, metaprogramming) stay silent - unproven, not a mismatch.

Fixes two concrete `attr_reader :location` mismatches it was masking, in
`Chain::Call` and `Chain::InstanceVariable` (the latter had a
`@todo: Missed nil violation` comment above the bad tag), plus the matching
`@param` on `ApiMap#var_at_location`.

## New signal this surfaces

Full self-typecheck at `strict`/`strong` adds ~7/26 new problems repo-wide
(on top of the existing backlog tolerated via `continue-on-error: true`).
Spot-checked - look like real pre-existing mismatches, flagged as follow-up.

## Test plan

- [x] `spec/type_checker/levels/strong_spec.rb` (2 new cases)
- [x] `spec/type_checker_spec.rb spec/type_checker/levels/* spec/type_checker/rules_spec.rb` - 284 examples, 0 failures
- [x] `rubocop` on changed files - no new offenses
- [x] `strict`/`strong` self-typecheck stays 0 problems on `api_map.rb`,
      `chain/call.rb`, `chain/instance_variable.rb`

https://claude.ai/code/session_01Vo1kewasiF1ywEQAT5Bth9

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -403,11 +403,10 @@ module Solargraph
     # @param candidates [Array<Pin::BaseVariable>]
     # @param name [String]
     # @param closure [Pin::Closure]
-    # @param location [Location]
+    # @param location [Location, nil]
     #
     # @return [Pin::BaseVariable, nil]
     def var_at_location candidates, name, closure, location
-      # @todo Location can be nil if clips have trouble finding node recipients
       return unless location
 
       with_correct_name = candidates.select { |pin| pin.name == name }
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -14,7 +14,7 @@ module Solargraph
         # @return [String]
         attr_reader :word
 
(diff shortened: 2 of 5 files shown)


apiology:
@castwide : Ready for review

castwide:
Looks good, thanks!

PR #1223, opened 2026-07-29
Infer element types for array literals

### LINKED ISSUE #1196: Specious inference in flow-sensitive typing
```ruby
array = [1, 'two']
index = 0
array[index] # Correctly inferred as Integer
index += 1
array[index] # Erroneously inferred as Integer
```

```ruby
array = [1, 'two']
# @type [Integer]
index = nil
array[index] # Erroneously inferred as Integer
```

```ruby
array = [1, 'two']
array.last # Erroneously inferred as Integer
```

```ruby
array = [1, 'two']
array.unshift 'zero'
array[0] # Erroneously inferred as Integer
```

apiology:
**Problem:** Reading an element out of an array literal gives back the wrong type, not merely a missing one.

```ruby
array = [1, 'two']
array[0]      # master: Array, nil   <- the array's own type, where an element belongs
array.first   # master: undefined    <- no type at all

y = [1]
y.push 'two'  # master: 0 problems found
```

`undefined` makes a check give up, which is bad. `::Array` is worse: it is confidently incorrect, so it satisfies checks that should fail and misdirects the ones that do fire. Every array literal is affected, homogeneous ones included.

**Solution:** Literals now infer `Array<T>` when the elements agree and a fixed `Array(A, B)` when they do not, so a literal index resolves precisely; anything that cannot be pinned to a literal — a reassigned index, `#last`, a position-shifting mutator — widens to the element union instead of guessing.

Restarg arguments are checked against the receiver's element type, so `y.push('two')` above is now flagged.

Known limitation, regression-tested: bare mutation with no reassignment is not tracked — after `array.unshift('zero')`, `array[0]` still reports `Integer`.

Fixes #1196.

https://claude.ai/code/session_01MmhuR7kn8L3wW1Y3F7UcSV


spec/source_map/clip_spec.rb
```diff
@@ -2116,12 +2211,10 @@ def signatures_at
 
     clip = api_map.clip_at('test.rb', [8, 6])
     type = clip.infer
-    # @todo Ideally this would be 'bot' - RBS isn't sophisticated enough to express this
```
apiology, line 2119: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -2143,12 +2236,10 @@ def signatures_at
 
     clip = api_map.clip_at('test.rb', [8, 6])
     type = clip.infer
-    # @todo Ideally this would be just ':foo' - RBS isn't sophisticated enough to express this
```
apiology, line 2146: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -2162,17 +2253,14 @@ def signatures_at
     api_map = Solargraph::ApiMap.new.map(source)
     clip = api_map.clip_at('test.rb', [4, 6])
     type = clip.infer
-    # @todo Ideally this would be just 'String' - RBS isn't sophisticated enough to express this
```
apiology, line 2165: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -2162,17 +2253,14 @@ def signatures_at
...
     expect(type.to_s).to eq('String, :foo')
 
     clip = api_map.clip_at('test.rb', [6, 6])
     type = clip.infer
-    # @todo Ideally this would be just 'Integer' - RBS isn't sophisticated enough to express this
```
apiology, line 2170: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -2162,17 +2253,14 @@ def signatures_at
...
     expect(type.to_s).to eq('Integer, :foo')
 
     clip = api_map.clip_at('test.rb', [8, 6])
     type = clip.infer
-    # @todo Ideally this would be just ':foo' - RBS isn't sophisticated enough to express this
```
apiology, line 2175: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -2060,14 +2159,12 @@ def signatures_at
 
     clip = api_map.clip_at('test.rb', [8, 6])
     type = clip.infer
-    # @todo Ideally this would be 'nil' - RBS isn't sophisticated enough to express this
```
apiology, line 2063: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -2089,12 +2186,10 @@ def signatures_at
...
```
apiology, line 2092: Claude: Leave this comment in place, please

spec/source_map/clip_spec.rb
```diff
@@ -3080,31 +3349,4 @@ class Desc < Base
...
-    clip = api_map.clip_at('test.rb', [4, 13])
-    paths = clip.complete.pins.map(&:path)
-    expect(paths).to include('String#upcase')
-    expect(paths).to include('Integer#abs')
-
-    clip = api_map.clip_at('test.rb', [7, 12])
-    paths = clip.complete.pins.map(&:path)
-    expect(paths).to include('String#upcase')
-    expect(paths).to include('Integer#abs')
-  end
```
apiology, line 3169: Claude: Why was this test removed?

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — restored in f5e1afb75. This comment predates this PR (it's from master's already-pending version of this test); I'd dropped it when rewriting the assertions, but the value it annotates is unchanged, so I just restored the comment.

spec/source_map/clip_spec.rb
apiology: Claude here — it was dropped (not deliberately, just missed) when tuple.rbs was first reverted back to union-only, before this session's work. Restored in f5e1afb75, with its first assertion updated: `foo[0]` now resolves precisely to 'String' (was the union) since this PR restores literal-indexed overloads, so the completion check there no longer expects Integer#abs alongside String#upcase. The second assertion (block param via #each, still a union) is unchanged.

castwide:
Inference is still making too many assumptions about literal values.

```ruby
x = 0
x += 1
x # => inferred as 0

y = [1]
y.push 'two'
y # => inferred as Array<Integer>
```

I'm not sure we should make those assumptions unless a YARD tag or an RBS signature makes the intent explicit.

apiology:
Thanks!

The first one is not what I intended for sure - that should have been handled as part of the 'track reassignment'  part of the PR.

#2 should have failed as a type violation on .push() - will look into that as well.

apiology:
Update on both examples:

**Example 1** (`x = 0; x += 1; x` → `0, Integer`) - this was never unsound: `0` is a subtype of `Integer`, so `0, Integer` already behaved correctly, it just rendered a redundant literal (`Array<1, Integer>` is really just `Array<Integer>` - a literal in a union with its own base type adds no information). Fixed in `67bd423bd`: `Pin::BaseVariable#probe` now drops that redundant literal when the base type is already present in the union. `x` now displays as `Integer`.

apiology:
**Example 2** (`y = [1]; y.push 'two'; y` → `Array<Integer>`, missing the pushed `String`): inference still can't track the mutation (documented limitation, has a regression spec), but `8e94b1cfa` now catches the bad argument at the call site - `y.push 'two'` is flagged as expecting `Integer`, got `String`, since restarg params were previously skipped entirely by argument type checking. (Surfaced and fixed a related bug where `RbsTranslator` discarded restarg/kwrestarg element types entirely.) RBS itself declares `push` as `(*E objects) -> self` on `Array[unchecked out E]`, so `E` is a real generic parameter - Steep, being RBS-based, resolves `E` against the receiver the same way we now do for the same signature. TypeScript and mypy check the pushed value against the declared element type too, though via their own type systems rather than RBS. None of them track the post-mutation container type, so that part isn't a gap unique to us.

apiology:
Separately, `edb944d5e` widens the RBS return type of the tuple's position-shifting mutators (`#unshift`, `#insert`, `#sort!`, etc.) so the reassignment idiom is sound too:

```ruby
array = [1, 'two']
array = array.unshift('zero')
array[0]  # now Integer, String, nil - was the stale, wrong Integer
```

Bare-statement mutation (`array.unshift('zero')`, no reassignment) is still the documented #1196 scenario-4 limitation - RBS's own maintainers hit and retreated from this same wall (https://github.com/ruby/rbs/commit/aae95840), and our fix stays sound since we're narrowing to an already-known superset type rather than introducing a new type variable.

spec/source/chain_spec.rb
```diff
@@ -235,7 +235,7 @@ class NotCorrect; end
     # chain = Solargraph::Source::NodeChainer.chain(node, 'test.rb')
     chain = Solargraph::Parser.chain(node, 'test.rb')
     type = chain.infer(api_map, Solargraph::Pin::ROOT_PIN, [])
-    expect(type.tag).to eq('Boolean')
+    expect(type.tag).to eq('true')
```
apiology, line 238: Let's check the non-literal type of this and ensure it is Boolean, which better matches the intent of this spec

lib/solargraph/source/chain/call.rb
```diff
@@ -186,6 +186,44 @@ def inferred_pins pins, api_map, name_pin, locals
...
+        # NilClass/TrueClass/FalseClass), but they're singletons, not
+        # dispatch-relevant values the way `0` vs `1` are for tuple
+        # indexing - excluding them keeps ordinary `T?`/nilable params
+        # (extremely common, e.g. String#split's `(Regexp | string |
+        # nil pattern)`) from tripping #literal_param_arg_matches?.
+        #
+        # @param unique_type [ComplexType::UniqueType]
+        # @return [Boolean]
+        def dispatch_literal? unique_type
+          unique_type.literal? && !%w[nil true false].include?(unique_type.name)
```
apiology: Can we create/use a predicate in the type hierarchy declaring whether something is a singleton?

lib/solargraph/source/chain/call.rb
```diff
@@ -111,7 +111,7 @@ def inferred_pins pins, api_map, name_pin, locals
                                                         gates: name_pin.gates,
                                                         source: :chain)
                 atype = atypes[idx] ||= arg.infer(api_map, arg_name_pin, locals)
-                unless param.compatible_arg?(atype, api_map) || param.restarg?
+                unless (param.compatible_arg?(atype, api_map) && literal_param_arg_matches?(param, atype, api_map)) || param.restarg?
```
apiology: Is there any case where we'd call compatible_arg? and not also want to check literal_param_arg_matches?  If not, would it make sense to move this into param's class?

lib/solargraph/pin/base_variable.rb
```diff
@@ -176,7 +184,18 @@ def return_types_from_node parent_node, api_map
...
+          items = assignment_types.flat_map(&:items).uniq
+          # Drop a literal item (e.g. `0`) when its non-literal base
+          # type (e.g. `Integer`) is also present in the same union -
+          # a later, wider assignment (`index += 1`) already
+          # subsumes it, so keeping both is redundant and reads as if
+          # the literal value were still reachable.
+          non_literal_names = items.reject(&:literal?).map(&:name)
+          items = items.reject { |item| item.literal? && non_literal_names.include?(item.non_literal_name) }
+          type_from_assignment = ComplexType.new(items)
+        end
```
apiology, line 203: This feels like it should be in the type hierarchy, doesn't it?  Some sort of simplify operation?

lib/solargraph/rbs_translator.rb
```diff
@@ -25,16 +25,45 @@ def self.to_complex_type(type)
...
+                    when :restarg
+                      # @sg-ignore RBS type understanding issue - see to_complex_type
+                      RbsTranslator.to_restarg_return_type(param_type.type)
+                    when :kwrestarg
+                      # @sg-ignore RBS type understanding issue - see to_complex_type
+                      RbsTranslator.to_kwrestarg_return_type(param_type.type)
+                    else
+                      # @sg-ignore RBS type understanding issue - to_complex_type's own param type is too narrow
+                      RbsTranslator.to_complex_type(param_type.type)
+                    end
```
apiology, line 35: Refactoring this into something that requires @sg-ignores doesn't feel like a positive move.  Maybe file an issue that we don't handle case statements well and suppress any rubocop issues with that issue PR as an explanation?

lib/solargraph/shell.rb
```diff
@@ -190,7 +190,7 @@ def gems *names
         names.each do |name|
           if name == 'core'
             # @sg-ignore cache_core and core? are dynamically defined
-            PinCache.cache_core(out: $stdout) # if !PinCache.core? || options[:rebuild]
+            PinCache.cache_core(out: $stdout) if !PinCache.core? || options[:rebuild]
```
apiology: Is this related to the core PR?  If not, let's move it to a different one

lib/solargraph/type_checker.rb
```diff
@@ -431,16 +433,28 @@ def argument_problems_for chain, api_map, closure_pin, locals, location
     # @param arguments [Array<Source::Chain>]
     # @param sig [Pin::Signature]
     # @param pin [Pin::Method]
+    # @param receiver_type [ComplexType] the type of the object the
+    #   method is being called on, used to resolve the restarg's
+    #   declared type (e.g. `Elem` for `Array#push`) against the
+    #   receiver's actual generic parameters (e.g. `Integer` for an
+    #   `Array<Integer>` receiver)
```
apiology: Is this specific to restargs?  Seems like in general this is about being able to resolve any generic parameter type?

spec/source/chain_spec.rb
apiology: Claude here — added in 42f564338: `expect(type.simplify_literals.tag).to eq('Boolean')` alongside the existing literal-type assertion.

lib/solargraph/source/chain/call.rb
apiology: Claude here — added `UniqueType#singleton?` (nil/true/false) in complex_type/type_methods.rb and switched dispatch_literal? to use it instead of the hardcoded name array, in 42f564338.

lib/solargraph/source/chain/call.rb
apiology: Claude here — moved literal_param_arg_matches? into Pin::Parameter#compatible_arg? (its only caller) in 42f564338. That also removes the redundant second `param.typify(api_map)` call literal_param_arg_matches? was doing on its own — compatible_arg? now computes ptype once and reuses it.

lib/solargraph/pin/base_variable.rb
apiology: Claude here — pulled this into `ComplexType#without_redundant_literals` in 42f564338; #probe now just does `ComplexType.new(items).without_redundant_literals`.

lib/solargraph/rbs_translator.rb
apiology: Claude here — the `@param type [RBS::Types::Bases::Base]` annotations were the actual bug: `RBS::Types::Bases::Base` is only the ancestor for `RBS::Types::Bases::{Bool,Void,Any,Nil,Top,Bottom,Self,Instance,Class}`; the classes this case statement actually dispatches on (Optional, Tuple, Literal, Union, Variable, Intersection, ClassInstance, Alias, Interface, ClassSingleton, ...) don't inherit from it at all. The RBS gem ships its own alias for the real union - `RBS::Types::t` (see its sig/types.rb). Switching the annotations on to_complex_type/to_restarg_return_type/to_kwrestarg_return_type to `RBS::Types::t` removes these 3 sg-ignores entirely - a real fix, not a suppression. That does surface a second, genuine gap once the param type is accurate: case/when doesn't narrow `type`'s class inside each branch of type_to_tag, so accessing branch-specific methods (#type, #types, #literal, #name, #args) there still needs an ignore. Filed https://github.com/castwide/solargraph/issues/1241 for that and pointed the remaining sg-ignores at it. All in 42f564338.

lib/solargraph/shell.rb
apiology: Claude here — not related to this PR, reverted in 42f564338 to match master exactly.

lib/solargraph/type_checker.rb
apiology: Claude here — yes, restarg-specific as implemented here: fixed-arity params still get their type from `params` (built by param_details_from_stack -> signature_param_details, via `param.typify(api_map)`), which doesn't take receiver_type into account either - so generics on fixed-arity params aren't resolved against the receiver at all right now, restarg or not. That's a separate, pre-existing gap this PR doesn't touch. Reworded the docstring in 42f564338 to say so explicitly instead of implying receiver_type is already general.

apiology: Please file an issue on this if one doesn't exist already

apiology: Claude here — filed https://github.com/castwide/solargraph/issues/1242 to track generalizing receiver-aware generic resolution to fixed-arity params (currently restarg-only, as discussed above).

lib/solargraph/shell.rb
apiology: Claude here — filed https://github.com/castwide/solargraph/issues/1248. Turns out this line is currently broken regardless of the commented-out condition: `PinCache.cache_core` doesn't exist (only `deserialize_core`/`serialize_core`/`uncache_core` do) - `solargraph gems core` raises a NoMethodError on master today. The `@sg-ignore cache_core and core? are dynamically defined` comment is inaccurate; neither is defined anywhere. Filed separately from this PR since it's a pre-existing bug, not something this PR's diff touches beyond the reverted one-liner.

apiology:
@castwide: Ready for re-review

apiology:
*🤖 Filed by Claude, not the account owner — acting on their behalf via their GitHub credentials.*

## Bug: restarg type-check produces a blank "expected" type for bare (non-generic) per-element restargs

`restarg_problems_for` (added in this PR, `type_checker.rb`) assumes a restarg's `par.return_type` is always generic-wrapped (e.g. `Array<Integer>` for `Array#push`, unwrapped one level via `.subtypes` to get `Integer`). For a restarg declared with a bare per-element type in RBS — e.g. `String#start_with?: (*string) -> bool` — `par.return_type` comes back as a plain, non-generic `Array` with `@subtypes=[]`. Unwrapping that gives `ComplexType.new([])`, an empty type.

The `return errors if ptype.nil? || ptype.undefined?` guard doesn't catch this empty `ComplexType`, so it falls through to the error message, which renders as blank:

```
Wrong argument type for String#start_with?: prefixes expected , received String
```

(note the blank between "expected" and ",")

### Reproduction

```ruby
# Gemfile
gem 'solargraph', github: 'apiology/solargraph', branch: '2026-08-04'
```

```ruby
# example.rb
# @param line [String]
# @return [Boolean]
def check(line)
  line.start_with?('x')
end
```

```
$ bundle exec solargraph typecheck example.rb --level strong
example.rb:4: Wrong argument type for String#start_with?: prefixes expected , received String
Typecheck finished in 0.045454 seconds.
1 problem found.
```

Confirmed via `git bisect` (8 steps, clean) that `8e94b1cf` ("Check restarg argument types against the receiver's element type") is the first bad commit; the immediately preceding commit is clean on this same repro.

### Suggested fix

Either:
- Have `restarg_problems_for` fall back to `par.return_type` itself (skipping the `.subtypes` unwrap) when the unwrap produces an empty type, or
- Fix `ComplexType#undefined?` to recognize `ComplexType.new([])` (zero items) as undefined, so the existing guard catches it and bails out cleanly instead of emitting a malformed message.

Happy to help track down why `RbsTranslator#to_parameter_pin` produces a bare `Array` (rather than the real per-element type) for this specific restarg shape, if useful.

apiology:
Bug fixed, ready for re-review, @castwide

apiology:

On **rbs 4.0.3**, this branch regresses a single-argument `Hash#fetch` with a literal-key receiver. Merge parent `6741f7fbf` is clean; the merge `771c3d02b` reports `Float, generic<X>`. A non-literal key type (`Hash{String => Float}`) is a passing control. On rbs 4.1.3 this branch leaks on *both* arms.

```ruby
# @param period [Hash{"Index" => Float}]
# @return [Float]
def literal_key(period)
  period.fetch('Index')   # => Float, generic<X>
end
```

I don't think this is a defect in this PR's own logic — keeping literal keys from widening to `String` is the point of the change. It's that doing so exposes the `Hash::_Key` nominal-vs-structural check described in #1231's specs, which #1266 addresses. The ask: **this branch shouldn't land ahead of a `_Key` fix that covers rbs 4.0.x.** #1231's spec guards that prerequisite as "on RBS >= 4.1.x", but our integration branch has #1266 merged and rbs 4.0.3 still leaks — so 4.0.x appears uncovered. A version-guarded pending spec here for the 4.0.x case would also make the dependency visible.

To reproduce, delete `core.ser` for the version under test and give each build its own `SOLARGRAPH_CACHE`; without `SOLARGRAPH_FORCE_VERSION` set, separate checkouts report the same version string and share one core-pin cache.

lib/solargraph/pin/base_variable.rb
```diff
@@ -176,7 +184,14 @@ def return_types_from_node parent_node, api_map
       # @return [ComplexType, ComplexType::UniqueType]
       def probe api_map
         assignment_types = assignments.flat_map { |node| return_types_from_node(node, api_map) }
-        type_from_assignment = ComplexType.new(assignment_types.flat_map(&:items).uniq) unless assignment_types.empty?
+        unless assignment_types.empty?
+          # @type [Array<ComplexType::UniqueType>]
+          items = assignment_types.flat_map(&:items).uniq
+          # A later, wider assignment (e.g. `index += 1`) can leave a
```
apiology:
```suggestion
          # A later, wider assignment (e.g. `index += n`) can leave a
```

lib/solargraph/pin/base_variable.rb
```diff
@@ -164,8 +164,16 @@ def return_types_from_node parent_node, api_map
             # Use the return node for inference. The clip might infer from the
             # first node in a method call instead of the entire call.
             chain = Parser.chain(node, nil, nil)
+            # Exclude the pin(s) this exact assignment belongs to from the
+            # candidates available to resolve its own RHS - a self-reference
+            # (e.g. `a = a`, or `index += 1` desugared to `index = index +
+            # 1`) must resolve against the variable's *other* assignments,
+            # not against the not-yet-computed value being derived here.
+            self_excluded_locals = clip.locals.reject do |candidate|
+              candidate.respond_to?(:assignments) && candidate.assignments.include?(parent_node)
```
apiology: Do we need duck-typing here?  What type of local wouldn't have assignments?  If candidate.assignments is valid, but a solargraph limitation (like resolution against the slots of a union) cause a typechecking error, let's sg-ignore that with an appropriate comment from rules.rb to demonstrate and quantify a limitation we should fix.

lib/solargraph/pin/base_variable.rb
```diff
@@ -223,6 +238,17 @@ def typify api_map
...
 
+      # A merged multi-assignment variable pin and its earliest constituent
+      # assignment pin share the same #choose-d (earliest) location but differ
+      # in presence, so discriminating on it keeps Chain's recursion guard from
+      # conflating "resolving the merged pin" with "resolving one of its
+      # narrower assignments" and dropping a legitimate recursive lookup.
+      #
+      # @return [String, nil]
+      def identity_discriminator
+        presence&.inspect
```
apiology: The comment is very helpful, but the code obscures what is being used to discriminate the different pins, and creates an undocumented contract for the presence#inspect method to maintain exactly the level of  uniqueness this code requires (but no more?).

lib/solargraph/pin/parameter.rb
```diff
@@ -221,17 +221,28 @@ def typify api_map
...
-      def compatible_arg? atype, api_map
+      # @param require_literal [Boolean] Whether a literal-typed
+      #   parameter requires the argument to also be a literal (see
+      #   #literal_arg_matches?). Callers doing overload *selection*
+      #   should retry with this set to false if no overload matches
+      #   with it true, so a literal-typed parameter with no
+      #   non-literal sibling overload to fall back to (e.g. a Hash's
+      #   key type resolved to a literal from its receiver's declared
+      #   type) doesn't reject every candidate outright.
+      def compatible_arg? atype, api_map, require_literal: true
```
apiology: I think this defeats the point of inferring the key type to a literal, and violates the principle of least astonishment - a worst-of-both-worlds situation.  

Could you discuss trade-offs and ideas on how to get the user's intent correct beyond this approach?

Does this PR implement inference of Hash literals based on their initial value?  If not, let's change the example to one more relevant at this time; if there is no such example, we should defer the decision until the last responsible moment.  

Does this question map to a spec, regardless of the desired behavior decided on?

lib/solargraph/type_checker.rb
```diff
@@ -440,16 +442,33 @@ def argument_problems_for chain, api_map, closure_pin, locals, location
...
+    # @param receiver_type [ComplexType] the type of the object the
+    #   method is being called on. Resolving a signature's generics
+    #   (e.g. `Elem`) against the receiver's actual generic
+    #   parameters (e.g. `Integer` for an `Array<Integer>` receiver)
+    #   is a general problem, but this is currently only plumbed
+    #   through to the restarg path below (see #restarg_problems_for)
+    #   - fixed-arity params still get their types from `params`
+    #   (built by #param_details_from_stack), which doesn't resolve
+    #   against the receiver. Generalizing that is tracked as a
+    #   follow-up, not attempted here.
```
apiology: What things are we trying to communicate to the next developer, and where is the best location for each one - param docstring, inline comment, commit message, new issue text, or PR description?

lib/solargraph/type_checker.rb
```diff
@@ -509,6 +528,102 @@ def signature_argument_problems_for location, locals, closure_pin, params, argum
...
+      errors
+    end
+
+    # @param sig [Pin::Signature]
+    # @param arguments [Array<Source::Chain>]
+    # @param idx [Integer] the restarg's index within sig.parameters
+    # @return [Array<Source::Chain>] the call-site arguments absorbed
+    #   by the restarg at idx
+    # @sg-ignore flow sensitive typing incorrectly includes an
+    #   intermediate local variable's type in the inferred return type
```
apiology: Please keep sg-ignore comments to one line and align with or add a new entry in rules.rb so we can track the impact of each class of issues.

spec/type_checker/levels/strict_spec.rb
```diff
@@ -49,6 +49,77 @@ def foo str; end
...
+      # ComplexType with zero items, which #undefined? failed to
+      # recognize (ComplexType#method_missing only delegates to
+      # #items.first, so it returns nil - not true - when #items is
+      # empty), so the empty type fell through to the error message
+      # instead of being skipped, rendering as "expected , received".
+      checker = type_checker(%(
+        1.instance_exec(2) { }
+      ))
+      expect(checker.problems.map(&:message)).to eq([])
+    end
```
apiology, line 73: I forget which level we do/should complain about the use of an untyped interface.  Your example is in the strict typechecking spec, implying we start complaining about it at the strong level.

- Does it?   
- Does it on the master branch?
- Should it?  Your choice of example strongly implies not - but surely there are lots of areas where we /should/ - e.g., exposing use of a gem without sufficient type declarations so that the user can provide them.

spec/type_checker/levels/strict_spec.rb
```diff
@@ -49,6 +49,77 @@ def foo str; end
...
+      # receiver's literal key type (e.g. "Index" from a
+      # `Hash{"Index" => Float}` @param tag), giving Hash#fetch a single
+      # candidate overload with no non-literal sibling to fall back to.
+      # The overload-selection gate that requires an exact-literal match
+      # (added to make tuple's literal-indexed overloads win over its
+      # safe catch-all) rejected that candidate outright, since the
+      # call's plain string argument isn't itself literal-typed - so no
+      # overload matched at all, and inference fell back to the union of
+      # every overload's declared return type instead of the one real
+      # match.
```
apiology: This behavior is certainly right - e.g., a literal string should typecheck as if the literal type has a superclass of the String class, meaning it would typecheck just fine being used to call a method iwth a String parameter.

spec/type_checker/levels/strict_spec.rb
```diff
@@ -49,6 +49,100 @@ def foo str; end
...
+      expect(checker.problems.map(&:message)).to eq([])
+    end
+
+    it 'catches a bad #<< argument against a fixed-arity generic parameter resolved against the receiver (#1242)' do
+      # Array#<< has a single fixed-arity parameter typed as the
+      # generic `E`. Before this fix, only restarg parameters (like
+      # Array#push's, above) had their generic types resolved against
+      # the receiver's actual element type - fixed-arity params like
+      # this one still typed themselves from the unresolved generic
+      # declaration and never flagged a mismatch.
```
apiology: Changelogging

spec/type_checker/levels/strict_spec.rb
```diff
@@ -49,6 +49,100 @@ def foo str; end
...
+    it 'does not report a bogus problem for a restarg typed as RBS `untyped` (#1223)' do
+      # Reported at https://github.com/castwide/solargraph/pull/1223#issuecomment-5198820509 -
+      # BasicObject#instance_exec is declared `(*untyped, **untyped)`, so its
+      # restarg has no per-element type to check arguments against.
+      # #restarg_problems_for used to unwrap that down to a
+      # ComplexType with zero items, which #undefined? failed to
+      # recognize (ComplexType#method_missing only delegates to
+      # #items.first, so it returns nil - not true - when #items is
+      # empty), so the empty type fell through to the error message
+      # instead of being skipped, rendering as "expected , received".
```
apiology: changelogging

lib/solargraph/type_checker.rb
```diff
@@ -433,23 +435,56 @@ def argument_problems_for chain, api_map, closure_pin, locals, location
...
       #   when possible, and when not, ensure provably
       #   incorrect situations are detected.
       sig.parameters.each_with_index do |par, idx|
-        return errors if par.decl == :restarg # bail out and assume the rest is valid pending better arg processing
+        if par.decl == :restarg
+          # A restarg absorbs every remaining positional argument at
+          # the call site - check each of them against the restarg's
+          # own declared/resolved type instead of bailing out on the
+          # whole signature. This is what catches e.g. `y.push('two')`
+          # on a `y: Array<Integer>`.
```
apiology: reduce to two lines

lib/solargraph/type_checker.rb
```diff
@@ -433,23 +435,56 @@ def argument_problems_for chain, api_map, closure_pin, locals, location
...
     # @param params [Hash{String => undefined}]
     # @param arguments [Array<Source::Chain>]
     # @param sig [Pin::Signature]
     # @param pin [Pin::Method]
+    # @param receiver_type [ComplexType] the type of the object the
+    #   method is being called on. Used to resolve a signature's
+    #   generics (e.g. `Elem`) against the receiver's actual generic
+    #   parameters (e.g. `Integer` for an `Array<Integer>` receiver),
+    #   for both fixed-arity params (below, and in #kwarg_problems_for)
+    #   and restargs (see #restarg_problems_for).
```
apiology: Limit to 3 lines

lib/solargraph/pin/base_variable.rb
```diff
@@ -223,6 +238,17 @@ def typify api_map
         adjust_type(api_map, raw_return_type)
       end
 
+      # A merged multi-assignment variable pin and its earliest constituent
+      # assignment pin share the same #choose-d (earliest) location but differ
+      # in presence, so discriminating on it keeps Chain's recursion guard from
+      # conflating "resolving the merged pin" with "resolving one of its
+      # narrower assignments" and dropping a legitimate recursive lookup.
```
apiology: 3 lines at max

lib/solargraph/type_checker.rb
```diff
@@ -509,6 +546,101 @@ def signature_argument_problems_for location, locals, closure_pin, params, argum
...
+    # @return [Array<Problem>]
+    def restarg_problems_for location, locals, closure_pin, arguments, sig, pin, receiver_type, par, idx
+      errors = []
+
+      # par.return_type is the type of the local variable the
+      # restarg is captured into inside the method body (e.g.
+      # `Array<Integer>`, not the unwrapped per-element `Integer`) -
+      # resolve any remaining generics against the receiver, then
+      # unwrap one level to get the type each individual argument
+      # must conform to.
```
apiology: 2 lines

lib/solargraph/source/chain/call.rb
```diff
@@ -182,6 +182,11 @@ def inferred_pins pins, api_map, name_pin, locals
             sorted_overloads = with_block + without_block
             # @type [Pin::Signature, nil]
             new_signature_pin = nil
+            # A literal-typed overload parameter (e.g. tuple's `(0 index)
+            # -> A`) requires an exact literal argument match - see
+            # Pin::Parameter#literal_arg_matches? - so a real sibling
+            # catch-all (e.g. tuple's `(int index) -> ...`) wins over it
+            # for a non-literal argument.
```
apiology: How does this relate to the code below?

lib/solargraph/type_checker.rb
```diff
@@ -485,6 +520,7 @@ def signature_argument_problems_for location, locals, closure_pin, params, argum
               return errors
             end
             ptype = params.key?(par.name) ? params[par.name][:qualified] : ComplexType::UNDEFINED
+            ptype = resolve_param_type_against_receiver(ptype, pin, receiver_type)
```
apiology: existing method?  new method?

lib/solargraph/pin/base_variable.rb
```diff
@@ -164,8 +164,16 @@ def return_types_from_node parent_node, api_map
             # Use the return node for inference. The clip might infer from the
             # first node in a method call instead of the entire call.
             chain = Parser.chain(node, nil, nil)
+            # Exclude the pin(s) this exact assignment belongs to from the
+            # candidates available to resolve its own RHS - a self-reference
+            # (e.g. `a = a`, or `index += 1` desugared to `index = index +
+            # 1`) must resolve against the variable's *other* assignments,
+            # not against the not-yet-computed value being derived here.
```
apiology: 3 lines

lib/solargraph/type_checker.rb
```diff
@@ -433,23 +435,56 @@ def argument_problems_for chain, api_map, closure_pin, locals, location
       result
     end
 
+    # Resolves any remaining generics in a fixed-arity parameter's
+    # already-qualified declared type (e.g. `Elem` for a made-up
+    # `Array#insert_first(v: Elem): void`) against the receiver's
+    # actual generic parameters (e.g. `Integer` for an
+    # `Array<Integer>` receiver). This is the fixed-arity counterpart
+    # to the unwrap-and-resolve done in #restarg_problems_for for
+    # restargs.
```
apiology: 3 lines

lib/solargraph/type_checker.rb
```diff
@@ -509,6 +546,101 @@ def signature_argument_problems_for location, locals, closure_pin, params, argum
...
+    # @param arguments [Array<Source::Chain>]
+    # @param idx [Integer] the restarg's index within sig.parameters
+    # @return [Array<Source::Chain>] the call-site arguments absorbed
+    #   by the restarg at idx
+    # @sg-ignore flow sensitive typing incorrectly includes an intermediate local variable's type in the inferred return type
+    def restarg_arguments sig, arguments, idx
+      # A restarg can be followed by trailing positional parameters
+      # (`def foo(*path, baz)`) - those consume the last N call-site
+      # arguments, so they don't belong to this restarg's own
+      # arguments.
```
apiology: 2 lines

lib/solargraph/type_checker.rb
```diff
@@ -509,6 +546,101 @@ def signature_argument_problems_for location, locals, closure_pin, params, argum
...
+      trailing_positional_count = trailing_positional_params.count { |p| p.decl == :arg }
+      # @type [Array<Source::Chain>]
+      # @sg-ignore flow sensitive typing issue with the ternary above
+      restargs = trailing_positional_count.zero? ? arguments[idx..] || [] : arguments[idx...-trailing_positional_count] || []
+
+      # A trailing bare hash argument (`foo(*args, key: val)`) is
+      # parsed as an implicit kwargs hash appended to the call's
+      # arguments - it belongs to the signature's keyword parameters,
+      # not the restarg.
+      # @type [Source::Chain, nil]
```
apiology: Remove `@type`

spec/type_checker/levels/strict_spec.rb
```diff
@@ -49,6 +49,100 @@ def foo str; end
...
+      # `Hash{"Index" => Float}` @param tag), giving Hash#fetch a single
+      # candidate overload with no non-literal sibling to fall back to.
+      # The overload-selection gate that requires an exact-literal match
+      # (added to make tuple's literal-indexed overloads win over its
+      # safe catch-all) rejected that candidate outright, since the
+      # call's plain string argument isn't itself literal-typed - so no
+      # overload matched at all, and inference fell back to the union of
+      # every overload's declared return type instead of the one real
+      # match.
+      checker = type_checker(%(
```
apiology: Check above PR comment, fix docs

lib/solargraph/type_checker.rb
```diff
@@ -509,6 +536,84 @@ def signature_argument_problems_for location, locals, closure_pin, params, argum
...
+      trailing_positional_count = trailing_positional_params.count { |p| p.decl == :arg }
+      restargs = trailing_positional_count.zero? ? arguments[idx..] || [] : arguments[idx...-trailing_positional_count] || []
+
+      # A trailing bare hash argument (`foo(*args, key: val)`) is
+      # parsed as an implicit kwargs hash appended to the call's
+      # arguments - it belongs to the signature's keyword parameters,
+      # not the restarg.
+      # @sg-ignore flow sensitive typing issue with the ternary above
+      last_arg = restargs.last
+      # @sg-ignore flow sensitive typing issue with the ternary above
```
apiology: Not a great new countable slug.  Look at rules.rb; if nothing matches, please investigate and write down your best assumption on what is going wrong that this can't be typechecked and yet is obviously correct.

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -458,10 +461,20 @@ def resolve_generics definitions, context_type
...
+                ComplexType::UNDEFINED
+              end
+            elsif context_type.all_params[idx]
+              context_type.all_params[idx]
+            elsif definitions.generic_defaults[generic_name]
+              # Tuples declare later positional generics (e.g. C, D, ...)
+              # as defaults in terms of earlier ones (e.g. C = A | B).
+              # Resolve those defaults against the same context instead of
+              # returning them as unresolved generic placeholders.
+              definitions.generic_defaults[generic_name].resolve_generics(definitions, context_type)
```
apiology: Is this a new feature in this PR?

lib/solargraph/type_checker.rb
apiology: **Claude:** New — introduced by this PR. `resolve_param_type_against_receiver` is defined at line 448 and called from two sites (513 and 645).

It is not an extraction of pre-existing inline logic: neither `resolve_generics` nor `receiver_type` appears anywhere in master's `type_checker.rb`, so signature checking on master did no receiver-generic resolution at all.

lib/solargraph/type_checker.rb
apiology: **Claude:** Nothing in `rules.rb` matched, and the ternary turns out not to be involved. Replaced with:

```
# @sg-ignore exclusive range literals infer as undefined, so Array#[] resolves to the element type
```

The trigger is the exclusive range in `arguments[idx...-trailing_positional_count]`. `NodeMethods` maps `:irange` to `::Range` but has no case for `:erange`, so every exclusive range literal infers as `undefined`. Probed against this branch, with receiver `@param a [Array<String>]`:

```ruby
(1..3)     # => "Range"                (1...3)    # => "undefined"
a[1..3]    # => "Array<String>, nil"   a[1...3]   # => "String, nil"
a[i, 2]    # => "Array<String>, nil"   a[i...-n]  # => "String, nil"
```

`Array#[]` then selects the `(Integer) -> T` overload instead of `(Range) -> Array[T]`, putting a bare `Source::Chain` into `restargs` — which is why `#last`, `#links` and `#[]` are all unresolved on it.

A second defect also contributes: `reduce_to_value_nodes` flattens an `:or` node in a conditional branch, so the `|| []` fallback never strips `nil`. Both fixes are needed before these three markers can be removed; neither alone suffices. With both applied, only the separate `@return` marker survives, so that fourth slug stands as written.

No upstream issue or PR exists for the `:erange` gap.

lib/solargraph/complex_type/unique_type.rb
apiology: **Claude:** Partly. `Pin::Closure#generic_defaults` and its population in `rbs_map/conversions.rb` are both already on master — master collected the defaults but never read them.

Consuming them here during `resolve_generics` is new, replacing master's `@todo Treating parameterized classes and tuples the same for now`. Covered by two new specs in `spec/complex_type_spec.rb`: "resolves generic parameters on a tuple using ()" and "resolves generic parameters on a tuple using <()>".

Issue #1042, opened 2025-08-28
[regression] Crash while generating activesupport pins


apiology:
solargraph-rails/spec/solargraph-rails/devise_spec.rb:15 is failing against the master branch in solargraph - I suspect after #1035.

https://github.com/castwide/solargraph/blob/master/lib/solargraph/api_map/store.rb#L75 calls `ComplexType.parse(fq_tag)` and I think with the new alias resolution code ends up being called with `Hash<String>', which is then rejected by the parse:

```
Failures:

  1) Solargraph::Rails::Devise includes devise modules
     Failure/Error: map = Solargraph::ApiMap.load_with_cache('./', STDERR)
     
     Solargraph::ComplexTypeError:
       Bad hash type: name=Hash, substring=<Symbol> - must have exactly two parameters
     # /Users/broz/src/solargraph/lib/solargraph/complex_type/unique_type.rb:57:in `parse'
     # /Users/broz/src/solargraph/lib/solargraph/complex_type.rb:379:in `block in parse'
     # /Users/broz/src/solargraph/lib/solargraph/complex_type.rb:318:in `each'
     # /Users/broz/src/solargraph/lib/solargraph/complex_type.rb:318:in `parse'
     # /Users/broz/src/solargraph/lib/solargraph/api_map/store.rb:75:in `get_superclass'
     # /Users/broz/src/solargraph/lib/solargraph/api_map/store.rb:225:in `get_ancestors'
     # /Users/broz/src/solargraph/lib/solargraph/api_map.rb:1008:in `resolve_method_alias'
     # /Users/broz/src/solargraph/lib/solargraph/api_map.rb:690:in `block in resolve_method_aliases'
     # /Users/broz/src/solargraph/lib/solargraph/api_map.rb:688:in `map'
     # /Users/broz/src/solargraph/lib/solargraph/api_map.rb:688:in `resolve_method_aliases'
     # /Users/broz/src/solargraph/lib/solargraph/api_map.rb:565:in `get_path_suggestions'
     # /Users/broz/src/solargraph/lib/solargraph/api_map.rb:573:in `get_path_pins'
     # /Users/broz/src/solargraph/lib/solargraph/gem_pins.rb:59:in `block in combine'
     # /Users/broz/src/solargraph/lib/solargraph/gem_pins.rb:57:in `map'
     # /Users/broz/src/solargraph/lib/solargraph/gem_pins.rb:57:in `combine'
[...]

PR #1224, opened 2026-07-29
Fix stale Hash::_Key expectations from upstream RBS core drift

apiology:
## Summary

While babysitting CI on #1219, I found the `rails`, `rspec`, `regression`, and
`run_solargraph_rspec_specs` jobs were already red on `master` before my change, due to an
unrelated upstream drift: RBS core's `Hash#[]` now types its key parameter as the `_Key`
duck-type interface rather than the generic `K`. That means instantiating `Hash{Symbol =>
untyped}` no longer substitutes `Symbol` for the key param type.

Two specs added for #1042 hardcoded the old `Symbol` expectation. This PR updates them to the
now-correct `::Hash::_Key`, with a comment explaining why.

## Verification

- Confirmed both specs fail identically on unmodified `master` (before this change) and pass
  after this fix.
- `bundle exec rspec spec/convention/activesupport_concern_spec.rb spec/rbs_map/conversions_spec.rb`
  → 24 examples, 0 failures.
- Full `bundle exec rspec` run locally: 1619 examples, 0 failures (aside from 2 machine-local
  failures in `spec/shell_spec.rb` caused by a gem conflict on my own workstation, unrelated to
  any code here — confirmed those also fail identically on unmodified `master`).

## Test plan

- [x] Targeted specs pass locally
- [x] Full local spec suite passes (excluding pre-existing machine-local failures, see above)
- [ ] CI green

---
*Generated with [Claude Code](https://claude.com/claude-code)*

### DIFF
--- a/spec/convention/activesupport_concern_spec.rb
+++ b/spec/convention/activesupport_concern_spec.rb
@@ -187,11 +187,17 @@ describe Solargraph::Convention::ActiveSupportConcern do
       end
 
       it 'finds superclass method pin parameter type' do
-        # RBS core's Hash#[] takes its key as the _Key duck-type interface, not
-        # the generic K, so instantiating Hash{Symbol => untyped} does not
-        # substitute the param type - see ruby/rbs core/hash.rbs.
+        # RBS core's Hash#[] started taking its key as the _Key duck-type
+        # interface instead of the generic K as of RBS 4.1.0, so instantiating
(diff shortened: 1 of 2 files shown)


castwide:
Modifying the tests is fine for now. We might want the type checker to report a problem when we pass an argument of the wrong type, but we can address that issue separately.

PR #1226, opened 2026-07-29
Infer element types for array literals and NilClass for nil literals

apiology:
## Summary

Two pre-existing type-inference gaps identified while investigating solargraph-rspec plugin CI failures (see castwide/solargraph#1224):

- Array literals (e.g. `[1, 2, 3]`) inferred as bare `Array` instead of `Array<Integer>` — `Chain::Array#resolve` never inferred element types from the literal's children.
- `nil` literals inferred as the internal pseudo-type tag `"nil"` instead of `NilClass` when passed through `simplify_literals` (used for e.g. hover/display purposes).

## Changes

- `lib/solargraph/source/chain/array.rb`: infer each array element's type and attach the union as the `Array`'s generic parameter; falls back to a bare `Array` when the literal is empty or an element's type can't be determined.
- `lib/solargraph/complex_type/unique_type.rb`: `simplify_literals` now maps the `nil` pseudo-type to `NilClass`, matching how other literals are simplified to their class name. Widened `to_rbs`'s nil check to also recognize `NilClass` so RBS output still renders the `nil` keyword type rather than `::NilClass`.
- Removed two `pending` markers in specs that this fix resolves (`spec/complex_type/conforms_to_spec.rb`, `spec/type_checker/levels/strict_spec.rb`) and updated two `simple_tags` assertions that encoded the old `"nil"` output.

## Test plan

- [x] `bundle exec rake spec` — full suite green except 2 pre-existing local-machine-only failures in `shell_spec.rb` (`Gem::ConflictError` from a stale globally-installed `solargraph` gem conflicting with `bundler`; reproduces identically on unmodified master, not a CI concern).
- [x] Verified against the downstream `apiology/solargraph-rspec` plugin test suite (the two target specs — `infers type for some_array` and `infers type for some_nil` — now pass; no new failures elsewhere).

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -451,6 +451,11 @@ module Solargraph
     # @param deep [Boolean] True to include superclasses, mixins, etc.
     # @return [Array<Solargraph::Pin::Method>]
     def get_methods rooted_tag, scope: :instance, visibility: [:public], deep: true
+      if rooted_tag.start_with? 'Array('
+        # Array() are really tuples - use our fill, as the RBS repo
+        # does not give us definitions for it
+        rooted_tag = "Solargraph::Fills::Tuple(#{rooted_tag[6..-2]})"
+      end
       rooted_type = ComplexType.try_parse(rooted_tag)
       fqns = rooted_type.namespace
       namespace_pin = store.get_path_pins(fqns).select { |p| p.is_a?(Pin::Namespace) }.first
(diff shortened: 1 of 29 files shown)


apiology:
Closing in favor of apiology/solargraph#40, which is the same branch rebased onto #1223 (`apiology-1196-literal-inference`) and opened as a PR against that branch directly. GitHub does not allow setting a PRs base to a branch that only exists in the head fork, so this could not be done as a stacked PR here upstream — the fork-internal PR shows just the incremental diff instead of a diff duplicating #1223s changes.

Will re-target the fork PR to `master` and reopen upstream once #1223 merges.

PR #1228, opened 2026-07-30
Allow arguments to satisfy RBS interface-typed parameters

### LINKED ISSUE #1227: Hash#fetch on Hash{Symbol => Class<X>} miscomputes generics when rbs >= 4.1.0

While verifying the fix for #1222 on Solargraph 0.60.2, we found a related-but-distinct regression: bumping the `rbs` gem alone (Solargraph version held constant) reintroduces a spurious "does not match inferred type ... generic<X>" error for `Hash#fetch` on a `Hash{Symbol => Class<X>}`-typed value.

### Repro

`Gemfile`:
```ruby
source 'https://rubygems.org'
gem 'solargraph', '0.60.2'
gem 'rbs', '4.1.0'
```

`repro.rb`:
```ruby
# typed: true
# frozen_string_literal: true

class Foo; end

class Holder
  # @return [Hash{Symbol => Class<Foo>}]
  def registry
    { x: Foo }
  end

  # @return [Foo]
  def use_it
    # @type [Class<Foo>]
    clazz = registry.fetch(:x)
    clazz.new
  end
end
```

```
repro.rb:15: Declared type Class<Foo> does not match inferred type Class<Foo>, generic<X> for variable clazz
Typecheck finished in 0.06s
1 problem found.
```

### Isolation

Solargraph version held fixed at 0.60.2 in both runs; only `rbs` changes:

| solargraph | rbs   | result |
|---|---|---|
| 0.60.2 | 4.0.2 | 0 problems |
| 0.60.2 | 4.1.0 | 1 problem (`generic<X>` false positive above) |

So the regression tracks the `rbs` version, not the Solargraph version. Filing against Solargraph rather than `ruby/rbs` since the failure only manifests through Solargraph's typecheck output (haven't confirmed whether the bug is in Solargraph's consumption of rbs's parsed types, or in rbs's own generated signatures for `Hash#fetch`) — happy to move it if it turns out to belong upstream in `ruby/rbs`.

Found while working on https://github.com/apiology/plate-spinner/pull/2149, which pins `rbs` to `< 4.1.0` to work around this.

apiology:

Fixes #1227.

## Problem

As of RBS 4.1.0, `Hash#fetch`'s single-argument overload takes its key as the `::Hash::_Key` duck-type interface instead of the generic `K` (see `ruby/rbs` `core/hash.rbs`). `Pin::Parameter#compatible_arg?` had no way to confirm an argument satisfies an interface without an explicit `include` declaration — which core gems don't add for every class that happens to satisfy `Hash::_Key`'s `#hash`/`#eql?` contract — so the single-arg overload was rejected as a non-match for a plain `Symbol` argument.

`Call#inferred_pins` then fell back to merging the return types of *all* of `Hash#fetch`'s overloads (including the unresolved `generic<X>` placeholder from the block-taking overload) instead of resolving the one overload that actually matched the call, producing a spurious typecheck error like:

```
Declared type Class<Foo> does not match inferred type Class<Foo>, generic<X> for variable clazz
```

on any `Hash{Symbol => Class<X>}#fetch(:key)` call under `rbs >= 4.1.0`.

## Fix

Pass `:allow_unmatched_interface` into the conformance check in `Pin::Parameter#compatible_arg?`, so an argument is treated as compatible with an interface-typed parameter rather than rejecting the overload outright. This mirrors how `TypeChecker` already treats unmatched interfaces when comparing declared vs. inferred types (`type_checker.rb:80`).

## Testing

- Added a regression spec in `spec/type_checker/levels/strong_spec.rb` reproducing the exact repro from #1227; confirmed it fails without the fix and passes with it.
- `bundle exec rspec`: 1618 examples, 0 failures (65 pre-existing pending, unrelated to this change).
- `bundle exec rubocop` on changed files: no offenses.
- `bundle exec rake typecheck`: 548 problems (down from 551 on the same base commit without this change) — the remaining problems are pre-existing RBS 4.1.0 drift unrelated to this fix, out of scope here.

### DIFF
--- a/lib/solargraph/pin/parameter.rb
+++ b/lib/solargraph/pin/parameter.rb
@@ -227,10 +227,17 @@ module Solargraph
         ptype = typify api_map
         return true if ptype.undefined?
 
+        # RBS interfaces (e.g., Hash::_Key) describe duck types that
+        # Solargraph can't verify structurally without an explicit
+        # `include`, which core gems don't declare for every class that
+        # happens to satisfy them (e.g., every object responds to
+        # Hash::_Key's #hash and #eql?). Treat an argument as compatible
+        # with an interface-typed parameter rather than rejecting the
+        # overload outright.
         return true if atype.conforms_to?(api_map,
                                           ptype,
                                           :method_call,
-                                          %i[allow_empty_params allow_undefined])
+                                          %i[allow_empty_params allow_undefined allow_unmatched_interface])
         ptype.generic?
       end
 
--- a/lib/solargraph/pin/parameter.rb
+++ b/lib/solargraph/pin/parameter.rb
@@ -227,13 +227,9 @@ module Solargraph
         ptype = typify api_map
         return true if ptype.undefined?
 
-        # RBS interfaces (e.g., Hash::_Key) describe duck types that
-        # Solargraph can't verify structurally without an explicit
-        # `include`, which core gems don't declare for every class that
-        # happens to satisfy them (e.g., every object responds to
-        # Hash::_Key's #hash and #eql?). Treat an argument as compatible
-        # with an interface-typed parameter rather than rejecting the
-        # overload outright.
+        # RBS interfaces (e.g., Hash::_Key) describe duck types core gems
+        # don't `include` on every satisfying class, so Solargraph can't
+        # verify them structurally; treat a match as compatible instead.
         return true if atype.conforms_to?(api_map,
                                           ptype,
                                           :method_call,
--- a/spec/type_checker/levels/strong_spec.rb
+++ b/spec/type_checker/levels/strong_spec.rb
@@ -892,5 +892,35 @@ describe Solargraph::TypeChecker do
       # an error when trying to declare sub as Subclass
       expect(checker.problems.map(&:message)).not_to include('Unresolved call to bar on Base')
     end
+
+    it 'resolves Hash#fetch return type on Hash{Symbol => Class<X>} without leaking a generic placeholder' do
+      # https://github.com/castwide/solargraph/issues/1227
+      #
+      # As of RBS 4.1.0, Hash#fetch's single-argument overload takes
+      # its key as the ::Hash::_Key duck-type interface instead of the
+      # generic K (see ruby/rbs core/hash.rbs). Solargraph couldn't
+      # prove a Symbol argument satisfies that interface, so it fell
+      # back to merging the return types of all of Hash#fetch's
+      # overloads (including the unresolved block-form's `generic<X>`)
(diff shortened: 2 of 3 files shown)


apiology:
*🤖 Comment from Claude, not Vince — acting on his behalf via his GitHub credentials.*

CI status: 28/29 checks pass, including `Solargraph / strong` (the typecheck job this issue is about).

The one failure, `run_solargraph_rspec_specs`, is pre-existing on `master` and unrelated to this change — it is the downstream `solargraph-rspec` fork's own Array-generics inference test, already failing on the last 5 pushes to `master`, including the commit this PR is based on. This PR does not touch that code path.

See https://github.com/castwide/solargraph/actions/runs/30486979645 for that same failure on master.

apiology:
@castwide This one's ready for review

lib/solargraph/pin/parameter.rb
```diff
@@ -227,10 +227,17 @@ def compatible_arg? atype, api_map
         ptype = typify api_map
         return true if ptype.undefined?
 
+        # RBS interfaces (e.g., Hash::_Key) describe duck types that
+        # Solargraph can't verify structurally without an explicit
+        # `include`, which core gems don't declare for every class that
+        # happens to satisfy them (e.g., every object responds to
+        # Hash::_Key's #hash and #eql?). Treat an argument as compatible
+        # with an interface-typed parameter rather than rejecting the
+        # overload outright.
```
apiology: 3 lines - and is this PR redundant with the structural type PR?

spec/type_checker/levels/strong_spec.rb
```diff
@@ -997,5 +997,35 @@ def baz(bases)
...
+    it 'resolves Hash#fetch return type on Hash{Symbol => Class<X>} without leaking a generic placeholder' do
+      # https://github.com/castwide/solargraph/issues/1227
+      #
+      # As of RBS 4.1.0, Hash#fetch's single-argument overload takes
+      # its key as the ::Hash::_Key duck-type interface instead of the
+      # generic K (see ruby/rbs core/hash.rbs). Solargraph couldn't
+      # prove a Symbol argument satisfies that interface, so it fell
+      # back to merging the return types of all of Hash#fetch's
+      # overloads (including the unresolved block-form's `generic<X>`)
+      # instead of picking the single-argument overload.
```
apiology: No changeloging.  Explain why your example is valid in the spec name, not with a bunch of comments that are quickly irrelevant.

apiology:
**Claude:** Superseded by #1266 — same fix, done via real structural conformance instead of a blanket allow.

PR #1230, opened 2026-07-30
TEMP: verify CI reproduces #1227 without fix (will close)

apiology:
Throwaway PR — do not merge.

This branch has only the regression spec from https://github.com/castwide/solargraph/pull/1228, without the fix, to confirm CI actually reproduces the bug under rbs >= 4.1.0 in the real CI environment rather than relying solely on local verification. Will close once CI shows the expected failure.


apiology:
Confirmed: CI reproduces the exact bug from #1227 without the fix (see https://github.com/castwide/solargraph/actions/runs/30538778613/job/90858343538 — single failure: `Declared type Class<Foo> does not match inferred type Class<Foo>, generic<X> for variable clazz`). Fix is in #1228. Closing this throwaway verification PR.

PR #1236, opened 2026-08-01
Fix exponential blowup in Pin::Method#combine_same_type_arity_signatures

### LINKED ISSUE #1235: solargraph typecheck --level strong hangs indefinitely (exponential blowup in Pin::Method#combine_same_type_arity_signatures)
## Summary

`solargraph typecheck --level strong` can hang indefinitely with unbounded memory growth (observed >11GB RSS on macOS, still climbing after 17+ minutes; >6.7GB on Linux x86_64 after ~10 minutes) when Solargraph's own codebase is typechecked against RBS collection data. Confirmed reproducing on both macOS ARM64 and Linux x86_64 (Docker `ruby:3.4`, `linux/amd64` platform), using the exact steps from `.github/workflows/typecheck.yml`'s `solargraph_typed` job against a clean v0.59 checkout (commit `735eaa5d`, no modifications).

Note: castwide/solargraph's own CI passes reliably in 5-14 minutes on `ubuntu-latest` — the hang was intermittent/environment-sensitive enough that I initially suspected platform (ARM64 vs x86_64) as the differentiator, but reproducing under Linux x86_64 (via QEMU emulation, since my only x86_64 access was emulated) ruled that out. Likely trigger is gem-version drift in `bundle update --pre rbs` (which pulls whatever RBS/gem-type versions are "latest" at install time) surfacing a pre-existing algorithmic bug rather than a platform-specific issue.

## Root cause

`Solargraph::Pin::Method#combine_same_type_arity_signatures` (`lib/solargraph/pin/method.rb:507-534`) merges same-type-arity overload signatures via:

```ruby
def combine_same_type_arity_signatures same_type_arity_signatures
  # This is an O(n^2) operation, so bail out if n is not small
  return same_type_arity_signatures if same_type_arity_signatures.length > 10

  same_type_arity_signatures.reduce([]) do |old_signatures, new_signature|
    next old_signatures + [new_signature] if old_signatures.empty?
    old_signatures.flat_map do |old_signature|
      potential_new_signature = old_signature.combine_with(new_signature)
      if potential_new_signature.type_arity == old_signature.type_arity
        old_signatures - [old_signature] + [potential_new_signature]
      else
        old_signatures + [new_signature]
      end
    end
  end
end
```

The inner `flat_map` iterates once per element of the accumulator `old_signatures`, but **each branch returns the entire accumulator array** (with one element swapped/added) rather than just that element's own contribution. Since `flat_map` concatenates every branch's return value, if none of N existing signatures merge with `new_signature`, the result is `N*(N+1)` elements instead of `N+1`. That larger array becomes the accumulator for the *next* `reduce` step, so the array size compounds **multiplicatively** with each additional signature processed — not just quadratically, as the comment on line 508 assumes.

The guard on line 509 (`return ... if same_type_arity_signatures.length > 10`) only checks the length of the *initial* input array. It doesn't catch this blowup because the explosion happens incrementally, inside the `reduce`, as later signatures are folded in — the array can start well under 10 elements and still explode a few iterations in.

This is hit in practice via `Solargraph::GemPins.combine_method_pins` when RBS collection data supplies many overloaded signatures with the same type arity for a single method (the included regression test uses `Integer#+` as a real-world example — RBS ships many dispatch-by-type overloads for it).

## Evidence

Live Ruby-level stack sampling via [rbspy](https://github.com/rbspy/rbspy) (attached to the running, non-instrumented process) showed an **identical stack** across repeated samples taken 15 seconds apart, confirming zero forward progress rather than merely slow legitimate work:

```
Enumerable#reduce [c function] - (unknown)
Array#each [c function] - (unknown)
(unknown) [c function] - (unknown)
block in combine_same_type_arity_signatures - /repo/lib/solargraph/pin/method.rb:533
Enumerable#flat_map [c function] - (unknown)
Array#each [c function] - (unknown)
(unknown) [c function] - (unknown)
block (2 levels) in combine_same_type_arity_signatures - /repo/lib/solargraph/pin/method.rb:532
```

Full callsite chain: `Shell#typecheck` → `TypeChecker#problems` → `#variable_type_tag_problems` → `Pin::BaseVariable#probe` → `Chain#infer` → `Chain::Call#resolve` → `ApiMap#get_method_stack`/`#get_methods` → `GemPins.combine_method_pins` → `Pin::Method#combine_with` → `#combine_signatures` → `#combine_signatures_by_type_arity` → `#combine_same_type_arity_signatures`.

RSS climbed steadily and continuously (no leveling off) during the hang: on macOS, >11GB after 17+ minutes; on Linux x86_64, 0 → 6.7GB in ~10 minutes, at which point the process was killed to conclude the investigation.

## To reproduce

```
git clone https://github.com/castwide/solargraph.git
cd solargraph
git checkout v0.59
echo "gem 'vernier', '>1.0', '<2'" >> Gemfile
bundle install
bundle update --pre rbs
bundle exec rbs collection install
SOLARGRAPH_ASSERTS=on bundle exec solargraph typecheck --level strong
```

Whether it hangs appears to depend on which RBS/gem versions `bundle update --pre rbs` resolves to at install time (i.e., how many same-type-arity overloads end up bundled for some method) — it reproduced consistently for me across Ruby 3.2.2, 3.2.6, and 3.4.5, on both macOS ARM64 and Linux x86_64, when run "fresh" (today's gem versions), which is consistent with a version of RBS or a gem's RBS signatures having accumulated enough same-type-arity overloads for some method to trip the exponential path.

## Fix

PR incoming with a fix that replaces the per-element `flat_map` (which re-emits the whole accumulator from within a block scoped to a single element) with a single `Array#map` pass that independently combines or keeps each element once, appending the new signature only if nothing merged — preserving the intended `O(n)` per-signature / `O(n^2)` overall behavior without the accidental multiplicative growth. Includes a regression test.

apiology:
Fixes #1235.

## Problem

`Pin::Method#combine_same_type_arity_signatures` (`lib/solargraph/pin/method.rb`) merges same-type-arity overload signatures via `reduce` + inner `flat_map`. The `flat_map` block returns the *entire* accumulator array from within a block scoped to a single element, so when N existing signatures fail to merge with a new one, the result balloons to `N*(N+1)` elements instead of `N+1`. That larger array becomes the accumulator for the next `reduce` step, so array size compounds multiplicatively with each additional signature — not the `O(n^2)` the existing guard comment assumes. The guard (`return ... if same_type_arity_signatures.length > 10`) only checks the *initial* array length and doesn't catch growth that happens incrementally inside the `reduce`.

In practice this causes `solargraph typecheck --level strong` to hang indefinitely with unbounded RSS growth when combining large families of same-type-arity overloads from RBS collection data (e.g. `Integer#+`, which RBS ships with many dispatch-by-type overloads). See #1235 for full repro steps, rbspy stack evidence (identical stuck stack across repeated samples), and RSS growth data on both macOS ARM64 and Linux x86_64.

## Fix

Replace the per-element `flat_map` (re-emitting the whole accumulator per element) with a single `Array#map` pass: each element of `old_signatures` is independently combined-or-kept exactly once, and `new_signature` is appended only if nothing merged. This preserves the original intent (attempt to merge `new_signature` into each existing signature; keep unmerged ones as-is) and the intended `O(n)`-per-signature / `O(n^2)`-overall cost, without the accidental multiplicative blowup.

## Testing

- Added a regression test (`spec/pin/method_spec.rb`) using doubles that never successfully merge, verifying the combined result length tracks the input length (`n`) rather than exploding. This test would hang/OOM against the old code.
- `bundle exec rspec spec/pin/method_spec.rb`: 48 examples, 0 failures (Ruby 3.4, Linux x86_64/Docker).
- Ran the full CI recipe from `.github/workflows/typecheck.yml` (`solargraph_typed` job) against this branch on Linux x86_64: `SOLARGRAPH_ASSERTS=on bundle exec solargraph typecheck --level strong` now **completes** in ~6.5 minutes (previously hung indefinitely, RSS unbounded). Exit code was 1 with 49 problems found in 14 of 241 files — I believe this is unrelated gem/RBS-version drift from `bundle update --pre rbs` picking up whatever is latest as of today rather than what was current when v0.59's CI last passed, not a regression introduced by this change (the fix only removes duplicate/exploded array entries; it doesn't change which signatures successfully merge). Flagging this explicitly rather than asserting a clean typecheck, since I haven't diffed against a same-day baseline without this fix (that baseline hangs, by definition, so a like-for-like comparison isn't available).


apiology:
This PR targeted the `v0.59` tag, which has diverged from `master` significantly (different commit lineages, ~46 commits on master not in v0.59 and 10 commits on v0.59 not in master since their common ancestor) — hence the huge conflict set here.

More importantly, current `master` already stubbed out `combine_same_type_arity_signatures` entirely in #1186 as a stopgap for this exact hang ("debug an infinite loop bug in Ruby 3.x"), so the code this PR patches is currently dead on master. Superseding this with a fresh PR based on current master that removes the stub and applies the real fix: see the new PR (linked from #1235).

PR #1237, opened 2026-08-01
Spec performance fixes

apiology:
Spec performance fixes: replace full-project gem-pin caching with targeted single-gem caching in slow specs. That path is new - the old bulk-cache-everything approach never exercised it - and exposed two latent correctness bugs below, both required to make it work.

## Spec performance

Replaced `ApiMap.load_with_cache` (builds pins for every gem) with `ApiMap.load` + targeted `catalog(bench)`/`cache_gem` for just the gems each test needs (`api_map_method_spec.rb`, `pin/base_spec.rb`, `pin/method_spec.rb`, `rbs_map/conversions_spec.rb`). Plus `bundler-cache: true` in CI and pinning `bundler: 2.5.23` to match `Gemfile.lock`.

## Correctness fixes (needed by the above)

`Workspace::Gemspecs#gemspec_or_preference` returned whatever spec type it was given instead of normalizing it:

```ruby
# gemspec: Gem::Specification | Bundler::LazySpecification | Bundler::StubSpecification
def gemspec_or_preference(gemspec)
  return gemspec unless preference_map.key?(gemspec.name) # returned as-is, unnormalized
  ...
end
```

`DocMap#cache` gates on `uncached_yard_gemspecs.include?(gemspec)` - an equality check a mismatched type silently failed, no-op'ing `cache_gem` for specs from the new `ApiMap#resolve_require`. Fixed via `to_gem_specification`.

The pattern also requires `catalog(bench)` - which registers a gem as required - to run *before* `cache_gem`. Two specs had it backwards, silently caching nothing. Reordered both.

## Minor

- Disabled `YARD/CollectionStyle` (rubocop-yard autocorrect mangles nested-generic `Hash` tags).
- `Diagnostics#diagnose` accepts an optional `workspace:` kwarg.
- Removed a redundant `mapper_spec.rb` test.

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -40,10 +40,6 @@ jobs:
       with:
         ruby-version: ${{ matrix.ruby-version }}
         bundler-cache: true
-        # see https://github.com/castwide/solargraph/actions/runs/19391419903/job/55485410493?pr=1119
-        #
-        # match version in Gemfile.lock and use same version below
-        bundler: 2.5.23
     - name: Set rbs version
       run: echo "gem 'rbs', '${{ matrix.rbs-version }}'" >> .Gemfile
     #  /home/runner/.rubies/ruby-head/lib/ruby/gems/3.5.0+2/gems/rbs-3.9.4/lib/rbs.rb:11:
@@ -71,15 +67,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: '3.4'
-        # see https://github.com/castwide/solargraph/actions/runs/19391419903/job/55485410493?pr=1119
-        #
-        # match version in Gemfile.lock and use same version below
-        bundler: 2.5.23
         bundler-cache: true
-    - name: Install gems
-      run: |
-        bundle _2.5.23_ install
-        bundle update rbs # use latest available for this Ruby version
     - name: Update types
       run: bundle exec rbs collection update
     - name: Run tests
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -40,6 +40,10 @@ jobs:
       with:
         ruby-version: ${{ matrix.ruby-version }}
         bundler-cache: true
+        # see https://github.com/castwide/solargraph/actions/runs/19391419903/job/55485410493?pr=1119
+        #
+        # match version in Gemfile.lock and use same version below
+        bundler: 2.5.23
     - name: Set rbs version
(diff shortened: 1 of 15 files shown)


.rubocop.yml
```diff
@@ -92,6 +92,11 @@ Metrics/PerceivedComplexity:
   Max: 40
 RSpec/ExampleLength:
   Max: 310
+# Autocorrect mangles short-style Hash tags with nested generics/parens
+# (e.g. Hash{Array(String, Array<String>) => String}) into invalid syntax.
+# Confirmed broken through rubocop-yard 1.3.0 (latest as of this writing).
```
apiology: Please also note YARD version

apiology: Added in 4939fd01b - noted yard 0.9.45 (also latest as of this writing) alongside the rubocop-yard version.

lib/solargraph/api_map.rb
```diff
@@ -800,6 +800,14 @@ def qualify_superclass fq_sub_tag
       store.qualify_superclass fq_sub_tag
     end
 
+    # @param require_path [String]
+    #
+    # @return [Array<Gem::Specification>, nil]
+    def resolve_require require_path
+      # @sg-ignore Unresolved call to directory on Solargraph::Workspace, nil
+      Workspace::Gemspecs.new(workspace.directory).resolve_require require_path
```
apiology, line 808: Let's add error handling here (throw an exception) rather than adding an sg-ignore

.github/workflows/rspec.yml
```diff
@@ -67,9 +71,18 @@ jobs:
...
+        bundler: 2.5.23
         bundler-cache: true
+    - name: Install gems
+      run: |
+        bundle _2.5.23_ install
+        bundle update rbs # use latest available for this Ruby version
     - name: Update types
-      run: bundle exec rbs collection update
+      run: |
+        bundle exec rbs collection update
```
apiology: Let's revert this change in 'Update types'

lib/solargraph/api_map.rb
apiology: **Claude:** Fixed in 070890e31 - now raises a clear error when called without a workspace instead of suppressing the typecheck warning.

.github/workflows/rspec.yml
apiology: **Claude:** Reverted in 070890e31 - back to the single-line `run:` form.

apiology:
@castwide - ready for review

PR #1238, opened 2026-08-01
Fix exponential blowup in Pin::Method#combine_same_type_arity_signatures

apiology:
Fixes an O(n^2)-turned-exponential bug in `Pin::Method#combine_same_type_arity_signatures`: the non-merging branch of the inner `flat_map` returned the *entire* accumulator array plus the new signature, instead of just that element's own contribution. Since `flat_map` concatenates every branch's return value, N non-merging existing signatures blow the result up to N*(N+1) elements instead of N+1, and that larger array becomes the next reduce step's accumulator -- so it compounds multiplicatively.

The existing "bail out if n is not small" guard only checks the length of the *initial* input array, so it doesn't catch this blowup, which happens incrementally inside the reduce.

This PR previously also carried a large stack of `Typecheck cleanup batch N` commits clearing the `solargraph typecheck --level strong` backlog. Those have been split out into a separate, dedicated PR consolidating non-essential type-annotation cleanup across several open PRs, so this PR stays scoped to the actual bug fix.

### DIFF
--- a/lib/solargraph/pin/method.rb
+++ b/lib/solargraph/pin/method.rb
@@ -509,11 +509,6 @@ module Solargraph
       #
       # @return [Array<Pin::Signature>]
       def combine_same_type_arity_signatures same_type_arity_signatures
-        # @todo Stubbing this method while we debug an infinite loop bug in Ruby 3.x.
-        #   The body below is intentionally preserved for when the stub is removed.
-        return same_type_arity_signatures
-
-        # rubocop:disable Lint/UnreachableCode
         # This is an O(n^2) operation, so bail out if n is not small
         return same_type_arity_signatures if same_type_arity_signatures.length > 10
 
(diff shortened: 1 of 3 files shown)


castwide:
I'd rather not expand PRs to include so many unrelated typecheck changes. Of the ~~52~~ 87 files changed here, only 2 appear to address the original issue. It's especially distracting when the behavior of the typechecker is in flux. We've already seen situations, e.g., where `@param` and `@return` types flip-flopped between `Array` and `Enumerable` throughout the codebase across consecutive PRs.

apiology:
Worth flagging separately (not fixed here, out of scope for this PR): `Gemfile.lock` and `rbs_collection.lock.yaml` are both gitignored, so every CI run resolves whatever gem/RBS-collection versions are current at install time. I hit this directly — a re-verification of this branch in a brand-new Docker clean-room (fresh `bundle install` + fresh `bundle exec rbs collection install`) surfaced an `Unresolved constant Gem::StubSpecification` that my long-lived local dev bundle (built earlier in the same session) didn't have, purely from gem/RBS version drift between the two installs. Fixed with a targeted `@sg-ignore` in the last commit, but the underlying non-determinism is structural — a *different* spot could fail to resolve on a different day for the same reason, independent of any code change. Might be worth considering committing lockfiles (or at least `rbs_collection.lock.yaml`) at some point to make strong-mode typecheck CI reproducible, now that it's a hard gate again.

apiology:
@castwide : Fixed - sorry about that!  This one got a little out of control 😅

castwide:
Thanks

PR #1231, opened 2026-07-30
Add intersection (A & B) types, including Hash-based record support

apiology:
## Summary

Adds intersection types (`A & B`) to Solargraph, in both plain YARD tags (`@param`/`@return`/`@type`) and inline RBS signatures (`#: () -> (A & B)`). Also adds the `|` union operator and `[...]` grouping brackets, matching https://github.com/lsegal/yard/pull/1700.

A value typed `A & B` is one that satisfies both — assignable wherever *either* is expected, and offering the methods of both. The individual members are its *conjuncts*. Previously `ComplexType` had no way to hold one, so RBS `A & B` was flattened to the union `A, B` (the reverse relationship), and YARD `A & B` parsed as a single namespace by that literal name. #1229 shows the flattening in the error text:

```
Wrong argument type for Consumer#project_to_h: project_obj expected Asana::Resources::Project, received Mocha::Mock, Asana::Resources::Project
```

Fixes https://github.com/castwide/solargraph/issues/1229. YARD has no official intersection syntax (https://github.com/lsegal/yard/issues/1644), so the tag form is a Solargraph extension pending upstream guidance.

## What changed

- **`ComplexType::UniqueType::Intersection`**, with `conforms_to?` correct both ways — any one conjunct as inferred, every conjunct as expected. `ComplexType.parse` gains top-level `&`, `|` (binds looser) and `[...]` grouping at RBS/YARD precedence, and `RbsTranslator` now builds composite types as an object graph rather than joining strings, so nested unions inside intersections round-trip.
- **Dispatch on intersection receivers.** `Call#method_stack_pins` needs only one conjunct to define a method, and dedupes candidates by `[path, return_type.tag]` rather than path alone, so `Box<Integer> & Box<String>` no longer collapses to whichever resolved first. Generics inside conjuncts now resolve too: `Intersection` inherited a `resolve_generics_from_context` that searches `#subtypes`/`#key_types`, which an intersection doesn't use, so `Class<generic<T>> & #new` never bound `T` — and an unresolved generic satisfies every conformance check, so a wrong `@return` was accepted silently.
- **`narrow_with`/`narrowed_return_type`**, renamed from #1119's `intersect_with`/`intersection_return_type` to free the name, and now building a real `Intersection` instead of discarding a mix-in narrowing when one side is a confirmed module.

## Precise Hash "record" dispatch

[...]

PR #1240, opened 2026-08-02
Fix and re-enable strong-level typechecking in CI

apiology:
Re-enables `solargraph typecheck --level strong` as an enforced CI gate in all four places it runs.

Most of the diff is `@sg-ignore` comments documenting type gaps strong mode can't resolve on its own: flow-sensitive-typing limits (postfix nil guards, `is_a?` narrowing across branches, reassignment tracking), the `nil`-vs-`NilClass` representation mismatch, guard-then-fetch patterns on `Hash#[]`/`Array#first`/`#last` that don't narrow, and a few RBS overload/type-alias gaps.

A handful of real fixes are included (missing/wrong `@return`/`@param` tags — e.g. `Host#pending_completions?` was tagged the nonexistent `Bool`). Genuine nil-safety and return-value fixes are deferred to #1245 and marked here with `@sg-ignore ... pending in #1245`, keeping this PR annotation- and CI-gate-only.

## Test plan

- [x] `bundle exec rspec`: 1618 examples, 0 failures
- [x] `SOLARGRAPH_ASSERTS=on bundle exec solargraph typecheck --level strong`: only the gaps explicitly deferred to #1245 remain — verified via before/after diff that the revert introduced no other new problems

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,12 @@ task :console do
   sh 'pry -I lib -r solargraph.rb'
 end
 
+# `vernier` is an optional dependency that `solargraph profile` requires
+# lazily, so a checkout without it reports three `Unresolved constant Vernier`
+# findings in shell.rb. CI appends this to the gitignored `.Gemfile` before
+# every typecheck run; do the same locally:
+#
+#   echo "gem 'vernier', '>1.0', '<2'" >> .Gemfile && bundle install
 desc 'Run the type checker'
 task typecheck: [:typecheck_strong]
 
--- a/Rakefile
+++ b/Rakefile
@@ -10,11 +10,7 @@ task :console do
   sh 'pry -I lib -r solargraph.rb'
 end
 
-# `vernier` is an optional dependency that `solargraph profile` requires
-# lazily, so a checkout without it reports three `Unresolved constant Vernier`
-# findings in shell.rb. CI appends this to the gitignored `.Gemfile` before
-# every typecheck run; do the same locally:
-#
+# `vernier` is an optional dependency typecheck needs; CI adds it before every run, so do the same locally:
 #   echo "gem 'vernier', '>1.0', '<2'" >> .Gemfile && bundle install
 desc 'Run the type checker'
 task typecheck: [:typecheck_strong]
@@ -51,9 +47,9 @@ task :full_spec do
   warn 'ending spec'
   # move coverage/full-new to coverage/full on success so that we
   # always have the last successful run's 'coverage info
-  # @sg-ignore Need a downcast here
+  # @sg-ignore https://github.com/castwide/solargraph/issues/1255
   FileUtils.rm_rf('coverage/full')
-  # @sg-ignore Need a downcast here
+  # @sg-ignore https://github.com/castwide/solargraph/issues/1255
   FileUtils.mv('coverage/full-new', 'coverage/full')
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -45,7 +45,9 @@ task :full_spec do
   warn 'ending spec'
   # move coverage/full-new to coverage/full on success so that we
   # always have the last successful run's 'coverage info
+  # @sg-ignore Need a downcast here
   FileUtils.rm_rf('coverage/full')
+  # @sg-ignore Need a downcast here
   FileUtils.mv('coverage/full-new', 'coverage/full')
 end
 
--- a/lib/solargraph/api_map/cache.rb
+++ b/lib/solargraph/api_map/cache.rb
@@ -62,7 +62,7 @@ module Solargraph
       # @param name [String]
       # @param context [String]
       # @return [String, nil]
-      # @sg-ignore Need a downcast here
+      # @sg-ignore https://github.com/castwide/solargraph/pull/1223
       def get_qualified_namespace name, context
         @qualified_namespaces["#{name}|#{context}"]
       end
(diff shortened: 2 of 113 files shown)


apiology:
Let's separate code changes into a separate PR and reference the PR # in @sg-ignore comments so we can keep this one annotation-only

lib/solargraph/api_map/cache.rb
```diff
@@ -71,11 +72,12 @@ def get_qualified_namespace name, context
       # @param value [String, nil]
       # @return [void]
       def set_qualified_namespace name, context, value
+        # @sg-ignore Need a downcast here
         @qualified_namespaces["#{name}|#{context}"] = value
```
apiology, line 76: Why does this and the above need a downcast?  What's the actual solargraph error?

apiology:
Done. Split the actual behavior-touching hunks out into #1245 (nil-guard fixes, return-value corrections, a dead duplicate-method removal in rbs_map/conversions.rb, and Sorbet-narrowing local-variable refactors) — see that PR for the full breakdown and test results.

This branch now reverts those 28 hunks back to their pre-cleanup form and marks each resulting strong-typecheck gap with an @sg-ignore comment referencing #1245, so it's annotation- and CI-gate-only as requested. Verified via solargraph typecheck --level strong that the revert introduces zero net-new problems (normalized diff against the pre-revert state), and the full test suite still passes (1618 examples, 0 failures).

—Claude

apiology:
Investigated whether the sg-ignore count growth (373 → 745 in my last update) reflects a regression rather than organic accumulation.

Found one: PR #1201 ("Ignore literal values in type inference," merged 2026-05-21) disabled tuple/literal element-type inference wholesale to fix specious-inference reports (#1196). Open PR #1223 restores it properly. That disabling turns out to be the root cause of a broad swath of downstream nil-check/downcast/overload-resolution gaps (e.g. `Open3.capture3`/`capture2e` overload selection), not just tuple indexing.

Determined the exact affected set empirically rather than guessing by category name: test-merged #1223's branch on top of this one and diffed `solargraph typecheck --level strong` output before/after (stripping line numbers to avoid false positives from line-count shifts). 79 lines came back "Unneeded @sg-ignore comment" — each one rewritten here to `# @sg-ignore https://github.com/castwide/solargraph/pull/1223`, replacing whatever reason (or blank comment) was there before. One of them was in this branch's own `#1245`-deferred set (`pin/block.rb`), which turns out to be downstream of the same root cause rather than independent.

Updated the count doc in `TypeChecker::Rules` to reflect this as a third bucket. Verified via full test suite (1618 examples, 0 failures) and a normalized before/after typecheck diff that this is a pure comment-text change with no side effects.

—Claude

apiology:
@castwide - ready for review

lib/solargraph/api_map/cache.rb
apiology: **Claude:**

```
lib/solargraph/api_map/cache.rb:65: Declared return type ::String, nil does not match inferred type ::String, ::NilClass, nil for Solargraph::ApiMap::Cache#get_qualified_namespace
lib/solargraph/api_map/cache.rb:74: Wrong argument type for Hash#[]=: value expected String, NilClass, received String, nil
```

Both errors are the same shape: `@qualified_namespaces`'s inferred value
type is `String, NilClass, nil` - `NilClass` (the class) and `nil` (the
literal) both present as separate union members, rather than collapsed
to one. That looks like the same class of duplicate-nil-representation
issue as the `ComplexType::UniqueType#tuple?` dead code I found and
flagged on
[castwide/solargraph#1231](https://github.com/castwide/solargraph/pull/1231)
(`return false` before the real logic, from
[8c40692](https://github.com/castwide/solargraph/commit/8c406923760c908748968db4d87bbac00758b2cf),
#1201) - haven't traced this one to a specific line yet.

Separately: `get_qualified_namespace`/`set_qualified_namespace` have no
callers anywhere in `lib/` or `spec/` - this pair looks unused.

Rakefile
```diff
@@ -45,7 +51,9 @@ task :full_spec do
   warn 'ending spec'
   # move coverage/full-new to coverage/full on success so that we
   # always have the last successful run's 'coverage info
+  # @sg-ignore Need a downcast here
   FileUtils.rm_rf('coverage/full')
```
apiology: Explain comment to me

Rakefile
```diff
@@ -45,7 +51,9 @@ task :full_spec do
...
+  # @sg-ignore Need a downcast here
```
apiology: Explain comment to me

lib/solargraph/workspace/gemspecs.rb
```diff
@@ -188,7 +186,12 @@ def to_gem_specification specish
                                                             # Specification
                                                             specish
                                                           end
-                                                        # @sg-ignore Unresolved constant Gem::StubSpecification
+                                                        # Do not remove on a local "Unneeded @sg-ignore"
+                                                        # verdict. Whether this constant resolves varies by
+                                                        # Ruby patch release, and CI runs a newer one than
```
apiology: This is based on rubygems version I suspect.  Maybe add an update command so that folks can make this go away.

lib/solargraph/workspace/gemspecs.rb
```diff
@@ -207,8 +210,10 @@ def query_external_bundle command
             "require 'bundler'; require 'json'; Dir.chdir('#{directory}') { puts begin; #{command}; end.to_json }"
           ]
           o, e, s = Open3.capture3(*cmd)
+          # @sg-ignore https://github.com/castwide/solargraph/pull/1223
           if s.success?
             Solargraph.logger.debug "External bundle: #{o}"
+            # @sg-ignore o.split("\n") is non-empty here because !o.empty?
```
apiology: sg-ignores shouldn't be explanations; they should be short descriptions of features to add or things to change.  See rules.rb for examples.  In this case, double check that split has an override that doesn't return nil.

PR #1244, opened 2026-08-02
Introduce a shared Type interface for Typedef (Concrete, Union)

apiology:
## Summary

This branch's base, castwide/solargraph#1204, gives `Typedef::Type` and `Typedef::Typeset` no shared ancestor — they're unrelated classes that happen to implement overlapping method names by convention. A call site that needs to accept "any kind of type value" has to enumerate both, and a new kind (an intersection, a type alias) means touching every one of those call sites by hand.

This PR adds `Solargraph::Type`, an abstract interface (`@!method` stubs) both implement, hoisted to the top level so `@return [Type]` needs no `Typedef::` prefix anywhere in the app. It also splits `Typedef::Typeset` into an abstract composite base plus a concrete `Union`, since "a set of types" doesn't say *how* the elements combine — that ambiguity matters once a second composite (intersection) exists.

```
Solargraph::Type
│
├── Typedef::Concrete            leaf: runtime class with optional
│     type parameters, e.g. String, Array<Integer>
│     └── Typedef::Tuple         (A, B) not [A, B]
│
├── Typedef::Typeset             abstract composite base
│     ├── Typedef::Union         ANY element (OR)
│     └── Typedef::Intersection  ALL elements (AND)  (planned,
│                                              castwide/solargraph#1231)
│
├── Typedef::Alias                                   (planned)
│     a named RBS type alias, lazily expanded
│
├── Typedef::Undefined                                (planned)
└── Typedef::Bottom                                    (planned,
                                              castwide/solargraph#1277)
```

Aggregation-ambiguous predicates (`generic?`) are split into explicit `any_generic?`/`all_generic?` rather than one method silently picking a reduction — a real inconsistency in the old `ComplexType` system (an `Intersection` there picked `any?` for `generic?` but `all?` for `rooted?`, with nothing forcing that choice to be deliberate) motivated this.

## Test plan

- `bundle exec rspec`: 1799 examples, 1 pre-existing failure (environmental — RBS core reports `Array`'s generic parameter as `E`, not `Elem`), 69 pending — matches this branch's own baseline
- `SOLARGRAPH_ASSERTS=on bundle exec solargraph typecheck --level strong`: 762 problems in 117 of 274 files — matches baseline exactly


apiology:
@castwide Ready for review - this one, the typedef changes, and the intersection/union/grouping changes are likely to conflict against each other - may be worth deciding in advance which order they go in and rebase in merge order in advance

castwide:
I'm tempted to let #1204 supersede this PR altogether. The eventual goal is to deprecate `ComplexType`. Changes that fix current bugs and improve functionality are fine, with the caveat the changes will need to be ported to `Typedef`. Large architectural changes are probably not worth the risk of unexpected side effects.

lib/solargraph/shell.rb
```diff
@@ -43,7 +43,7 @@ def socket
         end
         # @sg-ignore Wrong argument type for Backport.prepare_tcp_server: adapter expected Backport::Adapter, received Module<Solargraph::LanguageServer::Transport::Adapter>
         Backport.prepare_tcp_server host: options[:host], port: port, adapter: Solargraph::LanguageServer::Transport::Adapter
-        $stderr.puts "Solargraph is listening PORT=#{port} PID=#{Process.pid}"
+        warn "Solargraph is listening PORT=#{port} PID=#{Process.pid}"
```
castwide: The language server should always emit this line regardless of the log level. Clients need it to connect after it starts.

lib/solargraph/shell.rb
```diff
@@ -60,7 +60,7 @@ def stdio
         end
         # @sg-ignore Wrong argument type for Backport.prepare_stdio_server: adapter expected Backport::Adapter, received Module<Solargraph::LanguageServer::Transport::Adapter>
         Backport.prepare_stdio_server adapter: Solargraph::LanguageServer::Transport::Adapter
-        $stderr.puts "Solargraph is listening on stdio PID=#{Process.pid}"
+        warn "Solargraph is listening on stdio PID=#{Process.pid}"
```
castwide: See R46.

apiology:
> I'm tempted to let #1204 supersede this PR altogether. The eventual goal is to deprecate `ComplexType`. Changes that fix current bugs and improve functionality are fine, with the caveat the changes will need to be ported to `Typedef`. Large architectural changes are probably not worth the risk of unexpected side effects.

@castwide: Sounds good.  I went ahead and rebased this on top of #1204.  You can probably ignore the code for now and instead focus on the diagram in the PR description - would love to get on the same page on names and hierarchy.

PR #1245, opened 2026-08-02
Fix a nil crash and missing nil checks across the parser, type checker and RBS handling

apiology:
**Problem:** Strong-mode typechecking flags missing nil checks across the parser, type checker and RBS handling, one of which is a real crash.

The known crash involves parsing inline RBS:

```ruby
class Foo
  #: # invalid RBS
  def bar; end
end
```

```
pin/method.rb:733:in `return_type_from_inline_rbs':
  undefined method `type' for nil:NilClass (NoMethodError)
```

The rest are paths the checker cannot prove safe; the guards are cheap and the declared types hold either way.

**Solution:** Guard the nils, return `[]` where an `Array` is declared, and call `#fetch` where the key cannot be missing so a bad index raises at the bug rather than downstream.

Alongside those: other typechecker-based linting.

### DIFF
--- a/lib/solargraph/diagnostics/rubocop_helpers.rb
+++ b/lib/solargraph/diagnostics/rubocop_helpers.rb
@@ -25,12 +25,7 @@ module Solargraph
           specs = e.specs
           raise InvalidRubocopVersionError,
                 "could not find '#{e.name}' (#{e.requirement}) - " +
-                # @sg-ignore Gem::Specification#version resolves to
-                #   String here instead of Gem::Version, so the second
-                #   #version call is unresolved - "Unresolved call to
-                #   version on String". Likely an RBS stdlib gap for
-                #   Gem::Specification, not traced further. No upstream
-                #   issue filed yet.
+                # @sg-ignore Gem::Specification#version resolves to String here, not Gem::Version
                 "did find: [#{specs.map { |s| s.version.version }.join(', ')}]"
         end
         require 'rubocop'
--- a/lib/solargraph/diagnostics/rubocop_helpers.rb
+++ b/lib/solargraph/diagnostics/rubocop_helpers.rb
@@ -23,9 +23,10 @@ module Solargraph
         rescue Gem::MissingSpecVersionError => e
           # @type [Array<Gem::Specification>]
           specs = e.specs
+          found_versions = specs.map { |s| s.version.version }.join(', ')
           raise InvalidRubocopVersionError,
                 "could not find '#{e.name}' (#{e.requirement}) - " \
-                "did find: [#{specs.map { |s| s.version.version }.join(', ')}]"
+                "did find: [#{found_versions}]"
         end
         require 'rubocop'
       end
--- a/lib/solargraph/diagnostics/rubocop_helpers.rb
+++ b/lib/solargraph/diagnostics/rubocop_helpers.rb
@@ -24,9 +24,8 @@ module Solargraph
           # @type [Array<Gem::Specification>]
           specs = e.specs
           raise InvalidRubocopVersionError,
-                "could not find '#{e.name}' (#{e.requirement}) - " +
-                # @sg-ignore Gem::Specification#version resolves to String here, not Gem::Version
-                "did find: [#{specs.map { |s| s.version.version }.join(', ')}]"
+                "could not find '#{e.name}' (#{e.requirement}) - " \
+                "did find: [#{specs.map { |s| s.version.to_s }.join(', ')}]"
         end
         require 'rubocop'
       end
--- a/lib/solargraph/diagnostics/rubocop_helpers.rb
+++ b/lib/solargraph/diagnostics/rubocop_helpers.rb
@@ -23,10 +23,15 @@ module Solargraph
         rescue Gem::MissingSpecVersionError => e
           # @type [Array<Gem::Specification>]
           specs = e.specs
-          found_versions = specs.map { |s| s.version.version }.join(', ')
           raise InvalidRubocopVersionError,
-                "could not find '#{e.name}' (#{e.requirement}) - " \
-                "did find: [#{found_versions}]"
+                "could not find '#{e.name}' (#{e.requirement}) - " +
+                # @sg-ignore Gem::Specification#version resolves to
+                #   String here instead of Gem::Version, so the second
+                #   #version call is unresolved - "Unresolved call to
+                #   version on String". Likely an RBS stdlib gap for
+                #   Gem::Specification, not traced further. No upstream
+                #   issue filed yet.
+                "did find: [#{specs.map { |s| s.version.version }.join(', ')}]"
         end
         require 'rubocop'
       end
--- a/lib/solargraph/language_server/host.rb
+++ b/lib/solargraph/language_server/host.rb
@@ -702,8 +702,12 @@ module Solargraph
         @client_capabilities ||= {}
       end
 
+      # @sg-ignore tool-limitation:boolish - a bare && chain's return
+      #   type isn't inferred as Boolean without an explicit cast. No
+      #   upstream issue filed yet.
       def client_supports_progress?
-        !!(client_capabilities['window'] && client_capabilities['window']['workDoneProgress'])
+        # @sg-ignore Need to add nil check here
+        client_capabilities['window'] && client_capabilities['window']['workDoneProgress']
       end
 
       private
@@ -853,8 +857,12 @@ module Solargraph
         }
       end
 
+      # @sg-ignore tool-limitation:boolish - a bare && chain's return
+      #   type isn't inferred as Boolean without an explicit cast. No
+      #   upstream issue filed yet.
       def prepare_rename?
-        !!(client_capabilities['rename'] && client_capabilities['rename']['prepareSupport'])
+        # @sg-ignore Need to add nil check here
+        client_capabilities['rename'] && client_capabilities['rename']['prepareSupport']
       end
 
       # @param library [Library]
--- a/lib/solargraph/language_server/host.rb
+++ b/lib/solargraph/language_server/host.rb
@@ -702,12 +702,10 @@ module Solargraph
         @client_capabilities ||= {}
       end
 
-      # @sg-ignore tool-limitation:boolish - a bare && chain's return
-      #   type isn't inferred as Boolean without an explicit cast. No
-      #   upstream issue filed yet.
+      # @sg-ignore need boolish support for ? methods
       def client_supports_progress?
-        # @sg-ignore Need to add nil check here
-        client_capabilities['window'] && client_capabilities['window']['workDoneProgress']
+        window = client_capabilities['window']
+        window && window['workDoneProgress']
       end
 
       private
@@ -857,12 +855,10 @@ module Solargraph
         }
       end
 
-      # @sg-ignore tool-limitation:boolish - a bare && chain's return
-      #   type isn't inferred as Boolean without an explicit cast. No
-      #   upstream issue filed yet.
+      # @sg-ignore need boolish support for ? methods
       def prepare_rename?
-        # @sg-ignore Need to add nil check here
-        client_capabilities['rename'] && client_capabilities['rename']['prepareSupport']
+        rename = client_capabilities['rename']
+        rename && rename['prepareSupport']
       end
 
       # @param library [Library]
--- a/lib/solargraph/language_server/host.rb
+++ b/lib/solargraph/language_server/host.rb
@@ -704,8 +704,8 @@ module Solargraph
 
       # @sg-ignore need boolish support for ? methods
       def client_supports_progress?
-        window = client_capabilities['window']
-        window && window['workDoneProgress']
+        # @sg-ignore missing flow-sensitive typing on hash keys
+        client_capabilities['window'] && client_capabilities['window']['workDoneProgress']
       end
 
       private
@@ -857,8 +857,8 @@ module Solargraph
 
       # @sg-ignore need boolish support for ? methods
       def prepare_rename?
-        rename = client_capabilities['rename']
-        rename && rename['prepareSupport']
+        # @sg-ignore missing flow-sensitive typing on hash keys
+        client_capabilities['rename'] && client_capabilities['rename']['prepareSupport']
       end
 
       # @param library [Library]
--- a/lib/solargraph/language_server/host.rb
+++ b/lib/solargraph/language_server/host.rb
@@ -703,7 +703,7 @@ module Solargraph
       end
 
       def client_supports_progress?
-        client_capabilities['window'] && client_capabilities['window']['workDoneProgress']
+        !!(client_capabilities['window'] && client_capabilities['window']['workDoneProgress'])
       end
 
       private
@@ -854,7 +854,7 @@ module Solargraph
       end
 
       def prepare_rename?
-        client_capabilities['rename'] && client_capabilities['rename']['prepareSupport']
+        !!(client_capabilities['rename'] && client_capabilities['rename']['prepareSupport'])
       end
 
       # @param library [Library]
--- a/lib/solargraph/library.rb
+++ b/lib/solargraph/library.rb
@@ -254,7 +254,7 @@ module Solargraph
 
       result = []
       files = if only
-                [api_map.source_map(filename)]
+                [api_map.source_map(filename)].compact
               else
                 (workspace.sources + (@current ? [@current] : []))
               end
@@ -483,10 +483,10 @@ module Solargraph
       src = workspace.sources.find { |s| !source_map_hash.key?(s.filename) }
       if src
         Logging.logger.debug "Mapping #{src.filename}"
+        mapped_source = Solargraph::SourceMap.map(src)
         # @sg-ignore OK if src.filename is nil
-        source_map_hash[src.filename] = Solargraph::SourceMap.map(src)
-        # @sg-ignore OK if src.filename is nil
-        source_map_hash[src.filename]
+        source_map_hash[src.filename] = mapped_source
+        mapped_source
       else
         false
       end
--- a/lib/solargraph/library.rb
+++ b/lib/solargraph/library.rb
@@ -254,7 +254,7 @@ module Solargraph
 
       result = []
       files = if only
-                [api_map.source_map(filename)].compact
+                [api_map.source_map(filename)]
               else
                 (workspace.sources + (@current ? [@current] : []))
               end
@@ -483,10 +483,8 @@ module Solargraph
       src = workspace.sources.find { |s| !source_map_hash.key?(s.filename) }
       if src
         Logging.logger.debug "Mapping #{src.filename}"
-        mapped_source = Solargraph::SourceMap.map(src)
         # @sg-ignore OK if src.filename is nil
-        source_map_hash[src.filename] = mapped_source
-        mapped_source
+        source_map_hash[src.filename] = Solargraph::SourceMap.map(src)
       else
         false
       end
--- a/lib/solargraph/parser/parser_gem/node_chainer.rb
+++ b/lib/solargraph/parser/parser_gem/node_chainer.rb
@@ -118,8 +118,9 @@ module Solargraph
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
-            result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
-                                       NodeChainer.chain(n.children[1], @filename, n)])
+            or_lhs = NodeChainer.chain(n.children[0], @filename)
+            or_rhs = NodeChainer.chain(n.children[1], @filename, n)
+            result.push Chain::Or.new([or_lhs, or_rhs])
           elsif n.type == :if
             then_clause = if n.children[1]
                             NodeChainer.chain(n.children[1], @filename, n)
--- a/lib/solargraph/parser/parser_gem/node_chainer.rb
+++ b/lib/solargraph/parser/parser_gem/node_chainer.rb
@@ -118,9 +118,14 @@ module Solargraph
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
-            or_lhs = NodeChainer.chain(n.children[0], @filename)
-            or_rhs = NodeChainer.chain(n.children[1], @filename, n)
-            result.push Chain::Or.new([or_lhs, or_rhs])
+            # @sg-ignore tool-limitation:is_a-narrowing:external-gem-class -
+            #   the is_a?(::Parser::AST::Node) guard at method entry
+            #   narrows n for a workspace-local class but not for this
+            #   external gem-sourced one, so n.children below is
+            #   unresolved. No upstream issue filed yet.
+            result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
+                                       # @sg-ignore tool-limitation:is_a-narrowing:external-gem-class
+                                       NodeChainer.chain(n.children[1], @filename, n)])
           elsif n.type == :if
             then_clause = if n.children[1]
                             NodeChainer.chain(n.children[1], @filename, n)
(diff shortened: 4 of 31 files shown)


lib/solargraph/rbs_map/conversions.rb
```diff
@@ -142,7 +142,7 @@ def convert_self_types_to_pins decl, module_pin
       # @return [String]
       def rooted_name type_name
         name = type_name.to_s
-        RBS_TO_CLASS.fetch(name, name)
+        RBS_TO_CLASS[name] || name
```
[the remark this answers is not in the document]
apiology: **Claude:**

The signature was already correct (`Hash{String => String}`) - this is
a real Solargraph limitation, not a signature bug: `Hash#fetch(key,
default)` leaks a spurious `generic<X>` member into the declared return
type. Confirmed with a minimal repro (a plain `Hash{String => String}`
constant, `.fetch(key, key)`). Reverted to `.fetch` and tagged both
occurrences `tool-limitation:hash-fetch-default-generic-leak` in
1c1d0985b - see the reply on the `.fetch` comment below for the literal
error and repro.

lib/solargraph/parser/parser_gem/node_processors/opasgn_node.rb
```diff
@@ -19,8 +19,9 @@ def process
               # @sg-ignore Need a downcast here
               process_vasgn_target(target, operator, argument)
             else
+              target_type = target.type
               Solargraph.assert_or_log(:opasgn_unknown_target,
-                                       "Unexpected op_asgn target type: #{target.type}")
+                                       "Unexpected op_asgn target type: #{target_type}")
```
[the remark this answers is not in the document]
apiology: **Claude:**

Reverted and tagged in 1c1d0985b:

```
lib/solargraph/parser/parser_gem/node_processors/opasgn_node.rb:26: Unresolved call to type
```

`node.children[0]` is untyped, so `target.type` is unresolved wherever
it's called directly - same as the (already-marked) `if`/`elsif`
conditions just above.

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -118,8 +118,9 @@ def generate_links n
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
-            result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
-                                       NodeChainer.chain(n.children[1], @filename, n)])
+            or_lhs = NodeChainer.chain(n.children[0], @filename)
+            or_rhs = NodeChainer.chain(n.children[1], @filename, n)
+            result.push Chain::Or.new([or_lhs, or_rhs])
```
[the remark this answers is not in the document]
apiology: **Claude:**

Reverted and tagged in 1c1d0985b - same is_a?-narrowing gap as PR
[1311](https://github.com/apiology/solargraph/pull/1311): the
`is_a?(::Parser::AST::Node)` guard at method entry narrows `n` for a
workspace-local class but not for this external gem-sourced one, so
`n.children` below is unresolved:

```
lib/solargraph/parser/parser_gem/node_chainer.rb:127: Wrong argument type for Solargraph::Parser::ParserGem::NodeChainer.chain: node expected Parser::AST::Node, received Array, nil
```

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -156,28 +156,7 @@ def fqns type_name
...
-        # @todo Tuples are in flux
-        # tuples have their own class and are handled in other_type_to_type
-        if base == 'Hash' && params.length == 2
-          ComplexType::UniqueType.new(base, [params.first], [params.last], rooted: type_name.absolute?,
-                                                                           parameters_type: :hash)
-        else
-          ComplexType::UniqueType.new(base, [], params.reject(&:undefined?), rooted: type_name.absolute?,
-                                                                             parameters_type: :list)
-        end
+        RBS_TO_CLASS[ns] || ns
```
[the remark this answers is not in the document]
apiology: **Claude:**

Confirmed with a minimal repro (isolated from this method - a plain
`Hash{String => String}` constant, `.fetch(key, key)` with no `Class<X>`
or RBS-version involvement):

```ruby
module Foo
  # @type [Hash{String => String}]
  LOOKUP = {'a' => 'b'}.freeze

  # @return [String]
  def self.get(name)
    LOOKUP.fetch(name, name)
  end
end
```
->
```
Declared return type ::String does not match inferred type ::String, generic<X> for Foo.get
```

Reverted both `rooted_name` and `fqns` to `.fetch` and tagged
`tool-limitation:hash-fetch-default-generic-leak` in 1c1d0985b. Possibly
related to
[castwide/solargraph#1227](https://github.com/castwide/solargraph/issues/1227)
(same `generic<X>` leak symptom on `Hash#fetch`) but that repro is a
single-arg fetch on a `Class<X>`-valued hash under RBS >= 4.1 - a
different shape, not confirmed the same root cause.

lib/solargraph/parser/parser_gem/node_methods.rb
```diff
@@ -183,7 +183,8 @@ def const_nodes_from node
 
         # @param node [Parser::AST::Node]
         def splatted_hash? node
-          Parser.is_ast_node?(node.children[0]) && node.children[0].type == :kwsplat
+          child = node.children[0]
+          !!(child.is_a?(::Parser::AST::Node) && child.type == :kwsplat)
```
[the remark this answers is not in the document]

Reverted to the plain `is_ast_node?(...) && ....type == :kwsplat` form
and tagged in 1c1d0985b - two markers were needed, both attributed to
the `def`/expression:

```
lib/solargraph/parser/parser_gem/node_methods.rb:185: Solargraph::Parser::ParserGem::NodeMethods.splatted_hash? return type could not be inferred
lib/solargraph/parser/parser_gem/node_methods.rb:186: Unresolved call to type on Array, nil
```

The first is `tool-limitation:boolish` (a bare `&&` chain's return type
isn't inferred as Boolean without a cast). The second is the same
predicate-wrapper narrowing gap as PR
[1320](https://github.com/apiology/solargraph/pull/1320) -
`is_ast_node?` wraps an `is_a?` check internally, and flow-sensitive
typing doesn't narrow through a predicate method that way.

lib/solargraph/language_server/host.rb
```diff
@@ -854,7 +854,7 @@ def dynamic_capability_options
       end
 
       def prepare_rename?
-        client_capabilities['rename'] && client_capabilities['rename']['prepareSupport']
+        !!(client_capabilities['rename'] && client_capabilities['rename']['prepareSupport'])
```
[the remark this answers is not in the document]
apiology: **Claude:**

Done in 1c1d0985b - both `client_supports_progress?` and
`prepare_rename?` reverted to plain `&&` chains, tagged
`tool-limitation:boolish`. Each also needed a "Need to add nil check
here" marker for the intermediate `client_capabilities[...]` lookup,
which can return nil.

lib/solargraph/diagnostics/rubocop_helpers.rb
```diff
@@ -23,9 +23,10 @@ def require_rubocop version = nil
         rescue Gem::MissingSpecVersionError => e
           # @type [Array<Gem::Specification>]
           specs = e.specs
+          found_versions = specs.map { |s| s.version.version }.join(', ')
           raise InvalidRubocopVersionError,
                 "could not find '#{e.name}' (#{e.requirement}) - " \
-                "did find: [#{specs.map { |s| s.version.version }.join(', ')}]"
+                "did find: [#{found_versions}]"
```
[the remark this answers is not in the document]
apiology: **Claude:**

Reverted in 1c1d0985b:

```
Unresolved call to version on String
```

`s.version` resolves to `String` here instead of `Gem::Version`, so the
second `.version` call is unresolved - likely an RBS stdlib gap for
`Gem::Specification`, not traced further.

Also caught a real bug while placing the marker: putting a comment
between the two backslash-continued string literals turned the second
one into unreachable code (rubocop's `Lint/UnreachableCode` flagged it -
the message would have silently dropped the "did find: [...]" part at
runtime). Switched the continuation to `+` so a comment can sit between
the operands safely.

lib/solargraph/language_server/host.rb
```diff
@@ -853,7 +857,11 @@ def dynamic_capability_options
         }
       end
 
+      # @sg-ignore tool-limitation:boolish - a bare && chain's return
```
apiology: Wrong convention - see rules.rb

lib/solargraph/diagnostics/rubocop_helpers.rb
```diff
@@ -24,7 +24,13 @@ def require_rubocop version = nil
...
           specs = e.specs
           raise InvalidRubocopVersionError,
-                "could not find '#{e.name}' (#{e.requirement}) - " \
+                "could not find '#{e.name}' (#{e.requirement}) - " +
+                # @sg-ignore Gem::Specification#version resolves to
+                #   String here instead of Gem::Version, so the second
+                #   #version call is unresolved - "Unresolved call to
+                #   version on String". Likely an RBS stdlib gap for
+                #   Gem::Specification, not traced further. No upstream
+                #   issue filed yet.
```
apiology: See rules.rb

lib/solargraph/parser/parser_gem/node_methods.rb
```diff
@@ -182,7 +182,14 @@ def const_nodes_from node
...
 
         # @param node [Parser::AST::Node]
+        # @sg-ignore tool-limitation:boolish - a bare && chain's return
+        #   type isn't inferred as Boolean without an explicit cast. No
+        #   upstream issue filed yet.
         def splatted_hash? node
+          # @sg-ignore tool-limitation:is_a-narrowing:predicate-wrapper -
+          #   is_ast_node? wraps an is_a? check internally, and
+          #   flow-sensitive typing does not narrow through a predicate
+          #   method that way, so .type below is unresolved.
```
apiology: types.rb

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -140,6 +140,10 @@ def convert_self_types_to_pins decl, module_pin
       # @param type_name [RBS::TypeName]
       #
       # @return [String]
+      # @sg-ignore tool-limitation:hash-fetch-default-generic-leak -
+      #   Hash#fetch(key, default) leaks a spurious generic<X> member
+      #   into the declared return type even though RBS_TO_CLASS is a
+      #   plain Hash{String => String}. No upstream issue filed yet.
```
apiology: types.rb

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -151,6 +155,7 @@ def rooted_name type_name
       # @param type_name [RBS::TypeName]
       #
       # @return [String]
+      # @sg-ignore tool-limitation:hash-fetch-default-generic-leak
```
apiology: types.rb

spec/convention/struct_definition_spec.rb
```diff
@@ -135,6 +135,20 @@
...
     end
+
+    it 'supports assignment to a namespaced constant' do
+      source = Solargraph::SourceMap.load_string(%(
+        # @param bar [String]
+        # @param baz [Integer]
+        Foo::Baz = Struct.new(:bar, :baz)
+      ), 'test.rb')
+
+      # @type [Array<Solargraph::Pin::Parameter>]
```
apiology: Don't @type outside of literal initializers.  If we're not smart enough to infer this, add an sg-ignore and find/register the gap in rules.rb

lib/solargraph/type_checker.rb
```diff
@@ -524,6 +524,8 @@ def kwarg_problems_for sig, argchain, api_map, closure_pin, locals, location, pi
       result = []
       kwargs = convert_hash(argchain.node)
       par = sig.parameters[idx]
+      return result if par.nil?
```
apiology: If we don't think this could ever be nil, do .fetch() instead

lib/solargraph/library.rb
```diff
@@ -483,10 +483,10 @@ def next_map
       src = workspace.sources.find { |s| !source_map_hash.key?(s.filename) }
       if src
         Logging.logger.debug "Mapping #{src.filename}"
+        mapped_source = Solargraph::SourceMap.map(src)
         # @sg-ignore OK if src.filename is nil
-        source_map_hash[src.filename] = Solargraph::SourceMap.map(src)
-        # @sg-ignore OK if src.filename is nil
-        source_map_hash[src.filename]
+        source_map_hash[src.filename] = mapped_source
+        mapped_source
```
apiology: What's the purpose of introducing the the intermediate variable?

lib/solargraph/library.rb
```diff
@@ -254,7 +254,7 @@ def references_from filename, line, column, strip: false, only: false
 
       result = []
       files = if only
-                [api_map.source_map(filename)]
+                [api_map.source_map(filename)].compact
```
apiology: Why did we add this?

lib/solargraph/pin/parameter.rb
```diff
@@ -281,8 +281,9 @@ def typify_method_param api_map
             found = p
             break
           end
-          if found.nil? && !index.nil? && params[index] && (params[index].name.nil? || params[index].name.empty?)
-            found = params[index]
+          indexed_param = index.nil? ? nil : params[index]
+          if found.nil? && indexed_param && (indexed_param.name.nil? || indexed_param.name.empty?)
+            found = indexed_param
```
apiology: Is this a real gap, or are we working around an inference gap?

lib/solargraph/pin/block.rb
```diff
@@ -51,7 +51,7 @@ def destructure_yield_types yield_types, parameters
           yield_type = yield_types.first
           return yield_type.all_params if yield_type.tuple? && yield_type.all_params.length == parameters.length
         end
-        parameters.map.with_index { |_, idx| yield_types[idx] || ComplexType::UNDEFINED }
+        parameters.each_with_index.map { |_, idx| yield_types[idx] || ComplexType::UNDEFINED }
```
apiology: Shouldn't the first version work?

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -118,7 +118,13 @@ def generate_links n
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
+            # @sg-ignore tool-limitation:is_a-narrowing:external-gem-class -
+            #   the is_a?(::Parser::AST::Node) guard at method entry
+            #   narrows n for a workspace-local class but not for this
+            #   external gem-sourced one, so n.children below is
+            #   unresolved. No upstream issue filed yet.
             result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
+                                       # @sg-ignore tool-limitation:is_a-narrowing:external-gem-class
```
apiology: types.rb

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -118,7 +118,13 @@ def generate_links n
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
+            # @sg-ignore tool-limitation:is_a-narrowing:external-gem-class -
+            #   the is_a?(::Parser::AST::Node) guard at method entry
+            #   narrows n for a workspace-local class but not for this
+            #   external gem-sourced one, so n.children below is
+            #   unresolved. No upstream issue filed yet.
```
apiology: rules.rb

lib/solargraph/parser/parser_gem/node_processors/opasgn_node.rb
```diff
@@ -20,6 +20,10 @@ def process
               process_vasgn_target(target, operator, argument)
             else
               Solargraph.assert_or_log(:opasgn_unknown_target,
+                                       # @sg-ignore node.children[0] is untyped, so
+                                       #   target.type is unresolved here the same
+                                       #   way it already is at the if/elsif
+                                       #   conditions above.
```
apiology: rules.rb

lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
```diff
@@ -57,7 +57,10 @@ def parameters_from_inline_rbs
             match = source[offset...eol].to_s.match(/\A\s*#\[([^\]]*)\]/)
             return unless match
 
-            code = match[1].strip
+            captured = match[1]
+            return unless captured
+
+            code = captured.strip
```
apiology: Prefer .fetch() if you can prove to yourself the nil issue will not happen

lib/solargraph/language_server/host.rb
```diff
@@ -853,7 +857,11 @@ def dynamic_capability_options
...
+      #   type isn't inferred as Boolean without an explicit cast. No
+      #   upstream issue filed yet.
       def prepare_rename?
+        # @sg-ignore Need to add nil check here
```
apiology: then add one?

lib/solargraph/language_server/host.rb
```diff
@@ -702,7 +702,11 @@ def client_capabilities
         @client_capabilities ||= {}
       end
 
+      # @sg-ignore tool-limitation:boolish - a bare && chain's return
+      #   type isn't inferred as Boolean without an explicit cast. No
+      #   upstream issue filed yet.
       def client_supports_progress?
+        # @sg-ignore Need to add nil check here
```
apiology: Then add one

lib/solargraph/language_server/host.rb
```diff
@@ -702,7 +702,11 @@ def client_capabilities
         @client_capabilities ||= {}
       end
 
+      # @sg-ignore tool-limitation:boolish - a bare && chain's return
+      #   type isn't inferred as Boolean without an explicit cast. No
+      #   upstream issue filed yet.
```
apiology: rules.rb

lib/solargraph/diagnostics/rubocop_helpers.rb
```diff
@@ -24,7 +24,8 @@ def require_rubocop version = nil
           # @type [Array<Gem::Specification>]
           specs = e.specs
           raise InvalidRubocopVersionError,
-                "could not find '#{e.name}' (#{e.requirement}) - " \
+                "could not find '#{e.name}' (#{e.requirement}) - " +
+                # @sg-ignore Gem::Specification#version resolves to String here, not Gem::Version
```
apiology: Can we change this to .to_s?  I suspect different versions of RubyGems have different output types here, but .to_s should handle all.

lib/solargraph/language_server/host.rb
```diff
@@ -702,8 +702,10 @@ def client_capabilities
         @client_capabilities ||= {}
       end
 
+      # @sg-ignore need boolish support for ? methods
       def client_supports_progress?
-        client_capabilities['window'] && client_capabilities['window']['workDoneProgress']
+        window = client_capabilities['window']
+        window && window['workDoneProgress']
```
apiology: This sounds like it should be a sg-ignore documenting that we don't do flow-sensitive-typing on hash keys

lib/solargraph/parser/parser_gem/node_processors/opasgn_node.rb
```diff
@@ -20,6 +20,7 @@ def process
               process_vasgn_target(target, operator, argument)
             else
               Solargraph.assert_or_log(:opasgn_unknown_target,
+                                       # @sg-ignore node.children[0] is untyped, so target.type is unresolved
```
apiology: There's a PR that handles this I believe - let's give the URL here

lib/solargraph/parser/parser_gem/node_processors/resbody_node.rb
```diff
@@ -9,22 +9,24 @@ class ResbodyNode < Parser::NodeProcessor::Base
 
           # @return [void]
           def process
-            if node.children[1] # Exception local variable name
-              here = get_node_start_position(node.children[1])
+            exception_local = node.children[1] # Exception local variable name
```
apiology: Similar to above - the existing code is safe.  Let's instead document that we are missing flow-sensitive typing on array elements

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -118,7 +118,9 @@ def generate_links n
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
+            # @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
             result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
+                                       # @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
                                        NodeChainer.chain(n.children[1], @filename, n)])
```
apiology: Please find the PR which adds this and list the URL instead

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -151,6 +152,7 @@ def rooted_name type_name
       # @param type_name [RBS::TypeName]
       #
       # @return [String]
+      # @sg-ignore Hash#fetch(key, default) leaks a generic<X> into the return type
```
apiology: Pretty sure there's a PR open on this - list the URL

lib/solargraph/workspace/gemspecs.rb
```diff
@@ -180,8 +180,7 @@ def to_gem_specification specish
                                                           # turns a Bundler::StubSpecification into a
                                                           # Gem::StubSpecification if we can
                                                           if specish.respond_to?(:stub)
-                                                            # @sg-ignore flow sensitive typing ought to be able to handle 'when ClassName'
-                                                            to_gem_specification specish.stub
+                                                            to_gem_specification specish.send(:stub)
```
apiology: Should add a .send

apiology: Wrong analysis - "which produces a false alarm in typechecking when rspec-mocks is in the bundle" - and make it a `@todo`

apiology: **Claude:**

Reworded as a `@todo` in 1e9c9ed04:

```ruby
# @todo Dispatched through send to avoid
#   Object#stub from rspec-mocks, which
#   produces a false alarm in typechecking
#   when rspec-mocks is in the bundle
to_gem_specification specish.send(:stub)
```

apiology: **Claude:**

Added in 005240c97, then reworded as a `@todo` in 1e9c9ed04 per the
follow-up below.

lib/solargraph/language_server/host.rb
apiology: **Claude:**

Done in 005240c97. Both `client_supports_progress?` and `prepare_rename?`
go back to the direct double lookup, each carrying
`# @sg-ignore missing flow-sensitive typing on hash keys`:

```
lib/solargraph/language_server/host.rb:707: Unresolved call to [] on Hash{String => Boolean}, nil
lib/solargraph/language_server/host.rb:859: Unresolved call to [] on Hash{String => Boolean}, nil
```

The string is new, so it is registered in rules.rb at 2 alongside the
other counts this branch had left unrecorded.

lib/solargraph/pin/parameter.rb

An inference gap. Reverted to the original expression in 005240c97 - the
truthiness of `params[index]` does not narrow the two reads after it:

```
lib/solargraph/pin/parameter.rb:284: Unresolved call to name on YARD::Tags::Tag, nil
lib/solargraph/pin/parameter.rb:284: Unresolved call to name on YARD::Tags::Tag, nil
```

Marked `# @sg-ignore missing flow-sensitive typing on array elements`, the
string this branch already uses in resbody_node.rb.

lib/solargraph/pin/block.rb
apiology: **Claude:**

It should. Restored `parameters.map.with_index` in 005240c97; the method
then infers as an Enumerator rather than the mapped Array:

```
lib/solargraph/pin/block.rb:48: Declared return type ::Array<::Solargraph::ComplexType> does not match inferred type ::Enumerator<::Array(::Solargraph::Pin::Parameter, ::Array, ::Integer), undefined> for Solargraph::Pin::Block#destructure_yield_types
```

Marked `# @sg-ignore Need better handling of Enumerator#with_index` and
registered in rules.rb. No open upstream issue matches it.

Issue #1246, opened 2026-08-02
Overload resolution by argument type unions signatures instead of narrowing


apiology:
## Summary

When a method has multiple `@overload` tags that differ only by parameter type (and return different `@generic` types), calling it with an argument whose type matches exactly one overload infers the *union* of both overloads' return types instead of narrowing to the one that actually matches.

## Reproduction

```ruby
# @generic A
# @generic B
class Foo
  # @overload find(index)
  #   @param [String] index
  #   @return [generic<A>]
  # @overload find(index)
  #   @param [Symbol] index
  #   @return [generic<B>]
  def find(index); end
end

# @type [Foo(String, Integer)]
m = blah
mb = m.find('foo')
mb # inferred as "String, Integer" -- expected "String"
```

Same for `#[]`:

```ruby
# @overload [](index)
#   @param [String] index
#   @return [generic<A>]
# @overload [](index)
#   @param [Symbol] index
#   @return [generic<B>]
def [](index); end
```

```ruby
mb = m['foo']
mb # inferred as "String, Integer" -- expected "String"
```

## Investigation so far

- Reproduces identically on master's original (pre-refactor) `Chain::Call#inferred_pins`, so it's not specific to any recent refactor of that method.
- Added tracing to the overload-matching loop (`match_overload_type` / the equivalent inline logic on master): for `m.find('foo')`, only the `String`-typed overload is ever checked, and the type-compatibility match against it appears to succeed and short-circuit the loop as expected — i.e. the overload *selection* logic looks correct in isolation.
- Despite that, `clip.infer` for the resulting local variable (`mb`) still returns the union `"String, Integer"`. That points away from `Chain::Call`'s overload matching itself and toward how the local variable's inferred type gets resolved/cached afterward (`Pin::BaseVariable#probe` or similar assignment-tracking machinery) — worth investigating there rather than in the overload-matching code directly.

## Origin

Surfaced via two specs added in castwide/solargraph#1006 ("Improve pin caching", later split — the relevant piece is apiology/solargraph#34) which were checked in as `pending` rather than fixed:

- `spec/source_map/clip_spec.rb` — `'uses types to determine overload to match'`
- `spec/source_map/clip_spec.rb` — `'uses types to determine overload of [] to match'`

PR #1247, opened 2026-08-02
Improve overload resolution and macro handling in Chain::Call

apiology:
Fixes macro-based return type inference when the macro's `$1`/`$2` placeholders reference the call's actual arguments, not just literal text elsewhere:

```ruby
class Foo
  # @!macro
  #   @return [$1]
  def self.bar; end
end
Foo.bar(String)
#        ^^^^^^ should substitute into the return type
```

| | Before | After |
|---|---|---|
| `Foo.bar(String)` inferred type | `nil` | `String` |

Root cause: macro/directive reprocessing only ran when normal overload matching produced no type at all. `bar`'s macro-based signature *did* produce a (wrong) type through the plain matching path, so reprocessing never ran. `Chain::Call#inferred_pins`'s per-overload matching logic is extracted into `match_overload_type`, restructured so macro reprocessing runs whenever the matched type doesn't actually resolve the macro's placeholders against the real arguments.

Also adds spec coverage for related overload/macro edge cases; 8 of the new specs are `pending`, linked to tracking issues rather than left silently broken ([#1223](https://github.com/apiology/solargraph/pull/1223), [#1246](https://github.com/apiology/solargraph/issues/1246)).

**Testing**: `bundle exec rspec spec/source/chain/call_spec.rb spec/source_map/clip_spec.rb` — 203 examples, 0 failures, 31 pending.

### DIFF
--- a/spec/source/chain/call_spec.rb
+++ b/spec/source/chain/call_spec.rb
@@ -96,7 +96,6 @@ describe Solargraph::Source::Chain::Call do
   end
 
   it 'infers types from macros' do
-    pending 'WIP'
     source = Solargraph::Source.load_string(%(
       class Foo
         # @!macro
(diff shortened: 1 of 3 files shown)


apiology:
@lekemula @castwide: Ready for review

spec/source/chain/call_spec.rb
```diff
@@ -428,6 +455,191 @@ def foo(params)
...
+    # valid options here:
+    #   * emit type checker warning when adding [B.new] and type whole thing as '::A::B'
+    #   * type whole thing as '::A::B, A::C'
+    #   * type as undefined
+    expect(type.rooted_tags).to eq('::A::B, ::A::C').or eq('::A::C, ::A::B').or be_undefined
+    expect(type.rooted_tags).not_to eq('::A::C')
+  end
+
+  it 'qualifies types in a second Array#+' do
+    pending 'Array element-type tracking was reverted on master; re-enable when restored (see castwide/solargraph#1223)'
```
apiology: If there's a PR that a spec is pending on, just put the PR URL, no other narrative.

spec/source/chain/call_spec.rb
```diff
@@ -96,7 +96,6 @@ def self.new; end
   end
 
   it 'infers types from macros' do
-    pending 'WIP'
```
apiology, line 99: Are the two headlines from this PR (overload resolution and macro handling) dependent on each other?  What would be the trade-offs of moving the overload code over to 1223?

spec/source/chain/call_spec.rb
apiology: **Claude:**

Done in 59996faa3 - all 6 occurrences of this pending reason simplified
to the bare URL.

spec/source/chain/call_spec.rb
apiology: **Claude:** No, not dependent within this PR. `match_overload_type` (the overload-resolution extraction, pure lift of the existing inline loop body into a method, no behavior change) and `process_macro`/`process_directive` (the new macro/directive reprocessing, added after that loop, gated on `p.macros`/`p.directives`) touch different points in `inferred_pins` and never call into each other. Reverting one wouldn't remove the other.

The real coupling is with castwide/solargraph#1223, not between this PR's own two headlines. #1223 independently rewrites the same `sorted_overloads.each` loop in the same method (wrapping it in a two-pass `require_literal` retry for literal-overload matching), and touches the identical three files this PR touches: `lib/solargraph/source/chain/call.rb`, `spec/source/chain/call_spec.rb`, `spec/source_map/clip_spec.rb`. Whichever PR merges second has to hand-reconcile the loop body against the other's restructuring.

Trade-offs of moving the overload-resolution half of this PR into #1223:
- Pro: the reconciliation happens once, by whoever's already deep in that loop for #1223, instead of being redone independently after #1223 merges.
- Con: #1223 already touches 26 files; folding this in grows it further, and ties the macro fix's landing to #1223's review timeline even though `process_macro`/`process_directive` have no overlap with #1223's changes at all.

I'd leave this PR as-is and let whichever of the two merges second do the reconciliation - the overlap is confined to one loop, not the whole overload-matching mechanism.

PR #1225, opened 2026-07-29
Fix Gemspecs#in_this_bundle? crash when no Gemfile is discoverable

apiology:
## Summary

- `Workspace::Gemspecs#in_this_bundle?` calls `Bundler.definition&.lockfile&.to_s&.start_with?(directory)`, but `Bundler.definition` raises `Bundler::GemfileNotFound` (not `nil`) when no Gemfile is discoverable from the current process's working directory. The safe-navigation chain doesn't help since the exception happens while evaluating `Bundler.definition` itself.
- This crashes any caller of `in_this_bundle?` (e.g. `find_gem`, `all_gemspecs_from_bundle`) whenever Solargraph itself is running outside of any bundle context - e.g. installed and invoked as a standalone gem.
- Rescues the exception and treats the workspace as not being "this bundle" in that case, falling back to the existing external-bundle resolution path.

Note: `Workspace::Gemspecs` is currently unused by the rest of `lib/` (this is a real, isolated bug fix in existing, tested code - not a functional change to Solargraph's current behavior).

## Test plan

- [x] Added a regression spec in `spec/workspace/gemspecs_find_gem_spec.rb` that stubs `Bundler.definition` to raise `Bundler::GemfileNotFound` and asserts `find_gem` falls back cleanly instead of raising
- [x] Verified locally: reverting just the `lib` fix (keeping the new spec) reproduces the exact crash (`Bundler::GemfileNotFound` uncaught at `gemspecs.rb:222`); restoring the fix makes it pass
- [x] `bundle exec rspec spec/workspace/` - 53 examples, 0 failures
- [x] `bundle exec rubocop lib/solargraph/workspace/gemspecs.rb spec/workspace/gemspecs_find_gem_spec.rb` - no offenses
- [x] `bundle exec solargraph typecheck --level strong` - confirmed identical 8 pre-existing findings on baseline (line numbers only shift by the added lines); no new findings introduced

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NdhWz9sGH13UicXmPwb4p9

Issue #41, opened 2018-04-15
Variables aren't appearing in dropdown


garyking:
Not sure why, but for some of my large Ruby projects, variables don't appear in auto-complete dropdowns.

There aren't any hints in the output about this, either.

castwide:
One known limitation is that variables will not appear in completion lists unless they have an explicit assignment. Examples of instance variables with and without assignments:

![image](https://user-images.githubusercontent.com/3704922/38775152-2f087594-4049-11e8-8dc5-41a45db4297e.png)

![image](https://user-images.githubusercontent.com/3704922/38775154-3618b7fe-4049-11e8-9ae7-e571fd8d75ed.png)

Variables in completions will also be limited to those that are determined to be visible within the current context. Examples of variables inside and outside of context:

![image](https://user-images.githubusercontent.com/3704922/38775169-a4790618-4049-11e8-9ee2-9fbd2eaaeb19.png)

![image](https://user-images.githubusercontent.com/3704922/38775175-b8bc5346-4049-11e8-9eed-e11e6749fc36.png)

If you have an example of a variable that ought to be visible but isn't, please let me know.

garyking:
I think the problem I had fixed itself when I closed and re-opened the Ruby workspace. Otherwise, the variable dropdowns do seem to follow the rules you mentioned.

PR #1252, opened 2026-08-03
Rewrite PinCache as an instance-based engine, with wiring and CLI update

apiology:
Rewrites `PinCache` from a class-method-based module into an instance-based gem-caching engine, and updates every consumer to match.

## Why

Fixes a stale-type bug: after adding, updating, or removing an `rbs_collection.yaml` entry for a gem, hover/completion/type-checking could keep showing the *previous* RBS source's results, with no indication anything was stale short of a manual `solargraph cache --rebuild`.

## How

The old cache key conflated a gem's possible RBS sources (bundled types, `rbs_collection`, local override, unresolved) into one entry. Replaced with explicit `CACHE_KEY_GEM_EXPORT`/`_UNRESOLVED`/`_STDLIB`/`_LOCAL` classification plus a content hash of the collection's per-gem lock data. `PinCache` becomes an instance scoped to that config, since the key logic now needs to know which config produced it.

## What's in it

- **`PinCache`**: instantiable class taking `rbs_collection_path`/`rbs_collection_config_path`/`directory`/`yard_plugins`; `Yardoc`, `GemPins`, `RbsMap` updated to match.
- **Wiring**: `DocMap`, `ApiMap`, `Workspace`, `Library` delegate to it; CLI `cache`/`uncache`/`gems` are thin wrappers.
- Also: `Gemspecs#in_this_bundle?` no longer raises with no Gemfile ([#1225](https://github.com/castwide/solargraph/pull/1225)).

An unrelated `RbsMap::Conversions` dead-code fix, in a file this PR already touches, is split out to [#41](https://github.com/apiology/solargraph/pull/41).

**Testing**: `bundle exec rspec` — 1595 examples, 0 failures, 64 pending.

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -118,7 +118,7 @@ module Solargraph
       recreate_docmap = @unresolved_requires != unresolved_requires ||
                         # @sg-ignore Unresolved call to rbs_collection_path on Solargraph::Workspace, nil
                         workspace.rbs_collection_path != bench.workspace.rbs_collection_path ||
-                        @doc_map.uncached_gemspecs.any?
+                        @doc_map.any_uncached?
 
       if recreate_docmap
         @doc_map = DocMap.new(unresolved_requires, bench.workspace, out: nil) # @todo Implement gem preferences
@@ -168,16 +168,6 @@ module Solargraph
     # @return [::Array<Gem::Specification>]
     def uncached_gemspecs
       doc_map.uncached_gemspecs || []
-    end
-
-    # @return [::Array<Gem::Specification>]
-    def uncached_rbs_collection_gemspecs
-      @doc_map.uncached_rbs_collection_gemspecs
-    end
-
-    # @return [::Array<Gem::Specification>]
-    def uncached_yard_gemspecs
-      @doc_map.uncached_yard_gemspecs
     end
 
     # @return [Enumerable<Pin::Base>]
@@ -241,7 +231,7 @@ module Solargraph
     # @param rebuild [Boolean] whether to rebuild the pins even if they are cached
     # @return [void]
     def cache_all_for_doc_map! out: $stderr, rebuild: false
-      doc_map.cache_all!(out, rebuild: rebuild)
+      doc_map.cache_doc_map_gems!(out, rebuild: rebuild)
     end
 
     # @param gemspec [Gem::Specification]
@@ -660,6 +650,7 @@ module Solargraph
     # @param cursor [Source::Cursor]
     # @return [SourceMap::Clip]
     def clip cursor
+      # @sg-ignore Need to add nil check here
       raise FileNotFoundError, "ApiMap did not catalog #{cursor.filename}" unless source_map_hash.key?(cursor.filename)
 
       SourceMap::Clip.new(self, cursor)
@@ -751,10 +742,10 @@ module Solargraph
       logger.debug do
         "ApiMap#resolve_method_aliases(pins=#{pins.map(&:name)}, visibility=#{visibility}) => #{with_resolved_aliases.map(&:name)}"
       end
-      with_resolved_aliases
+      GemPins.combine_method_pins_by_path(with_resolved_aliases)
     end
 
-    # @return [Workspace, nil]
+    # @return [Workspace]
     def workspace
       doc_map.workspace
     end
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -649,7 +649,6 @@ module Solargraph
     # @param cursor [Source::Cursor]
     # @return [SourceMap::Clip]
     def clip cursor
-      # @sg-ignore Need to add nil check here
       raise FileNotFoundError, "ApiMap did not catalog #{cursor.filename}" unless source_map_hash.key?(cursor.filename)
 
       SourceMap::Clip.new(self, cursor)
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -118,7 +118,7 @@ module Solargraph
       recreate_docmap = @unresolved_requires != unresolved_requires ||
                         # @sg-ignore Unresolved call to rbs_collection_path on Solargraph::Workspace, nil
                         workspace.rbs_collection_path != bench.workspace.rbs_collection_path ||
-                        @doc_map.uncached_gemspecs.any?
+                        @doc_map.any_uncached?
 
       if recreate_docmap
         @doc_map = DocMap.new(unresolved_requires, bench.workspace, out: nil) # @todo Implement gem preferences
@@ -168,16 +168,6 @@ module Solargraph
     # @return [::Array<Gem::Specification>]
     def uncached_gemspecs
       doc_map.uncached_gemspecs || []
-    end
-
-    # @return [::Array<Gem::Specification>]
-    def uncached_rbs_collection_gemspecs
-      @doc_map.uncached_rbs_collection_gemspecs
-    end
-
-    # @return [::Array<Gem::Specification>]
-    def uncached_yard_gemspecs
-      @doc_map.uncached_yard_gemspecs
     end
 
     # @return [Enumerable<Pin::Base>]
@@ -241,7 +231,7 @@ module Solargraph
     # @param rebuild [Boolean] whether to rebuild the pins even if they are cached
     # @return [void]
     def cache_all_for_doc_map! out: $stderr, rebuild: false
-      doc_map.cache_all!(out, rebuild: rebuild)
+      doc_map.cache_doc_map_gems!(out, rebuild: rebuild)
     end
 
     # @param gemspec [Gem::Specification]
@@ -659,6 +649,7 @@ module Solargraph
     # @param cursor [Source::Cursor]
     # @return [SourceMap::Clip]
     def clip cursor
+      # @sg-ignore Need to add nil check here
       raise FileNotFoundError, "ApiMap did not catalog #{cursor.filename}" unless source_map_hash.key?(cursor.filename)
 
       SourceMap::Clip.new(self, cursor)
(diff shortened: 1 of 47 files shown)


apiology:
@castwide - ready for review

apiology:
*🤖 Filed by Claude, not the account owner — acting on their behalf via their GitHub credentials.*

## Bug: `Workspace::Gemspecs#resolve_require` fails to resolve gems whose require path differs from their package name

This PR introduces a regression that silently drops gems from `DocMap`/`ApiMap` cataloging whenever a project's `require` path doesn't textually match the gem's RubyGems package name (e.g. `require 'active_support'` for the `activesupport` gem). No error or warning is produced anywhere — the gem's pins simply never get built or cached, and any of *its own* transitive dependencies never get discovered either.

### Root cause

`lib/solargraph/workspace/gemspecs.rb`, `Gemspecs#resolve_require`:

```ruby
def resolve_require require
  ...
  # Determine gem name based on the require path
  file = "lib/#{require}.rb"
  spec_with_path = Gem::Specification.find_by_path(file)
  ...
  gem_names_to_try = [
    spec_with_path&.name,
    require.tr('/', '-'),
    require.split('/').first
  ].compact.uniq
  ...
```

`Gem::Specification.find_by_path` already resolves relative to each gem's own `require_paths` (normally `["lib"]`) internally — it expects the path exactly as it would appear in a `require` statement (e.g. `"active_support"`), **not** pre-prefixed with `"lib/"`. Prepending `"lib/"` here double-prefixes the lookup and it always returns `nil`:

```ruby
Gem::Specification.find_by_path("active_support")        # => #<Gem::Specification activesupport-8.0.5>
Gem::Specification.find_by_path("lib/active_support.rb") # => nil
```

Because `spec_with_path` is always `nil`, `resolve_require` falls through to two name-guessing fallbacks (`require.tr('/', '-')` and `require.split('/').first`), both of which produce `"active_support"`. Neither matches the real gem name `"activesupport"`, so the lookup against `all_gemspecs_from_bundle` also fails. The require path is left unresolved.

This only becomes *visible* for gems where the conventional require path differs from the RubyGems package name — a well-known split in the Ruby ecosystem (`activesupport`/`active_support`, `actionpack`/`action_pack`, etc.). For the common case where they're identical (`asana`, `json`, `faraday`, ...) the fallback guess happens to work, masking the bug.

The pre-refactor code (before this PR, in `DocMap#resolve_path_to_gemspecs`) called `Gem::Specification.find_by_path(path)` directly with **no** `"lib/"` prefix, and worked correctly.

### Downstream effect (how this was found)

In a real project (github.com/apiology/checkoff) this silently broke `Minitest` resolution entirely, with zero errors even at `SOLARGRAPH_LOG=debug`. The causal chain:

1. `activesupport` never resolves as a required gemspec (this bug).
2. `activesupport` has a genuine **runtime** dependency on `minitest` (it ships Minitest-based test helpers) — this transitive dependency is never discovered as a result.
3. The project's only literal `require 'minitest/autorun'` lived in a file excluded from the workspace scan via `.solargraph.yml`.
4. Both possible discovery paths for `minitest` failed simultaneously, so it vanished from the ApiMap with no diagnostic of any kind — `bin/solargraph pin Minitest` returned nothing, exit 1.

### Minimal reproduction

No test framework needed — this reproduces with `activesupport` alone.

`Gemfile`:
```ruby
# frozen_string_literal: true
source 'https://rubygems.org'

gem 'activesupport'
gem 'solargraph',
    github: 'apiology/solargraph',
    branch: '2026-08-04'
```

`example.rb`:
```ruby
require 'active_support'
```

```
$ bundle install
$ bundle exec solargraph pin ActiveSupport
$ echo $?
1
```

No output at all — `ActiveSupport` fails to resolve. On the parent commit (`dba36ec2b2d8b1e5c72dd80b6914d3589b81b044`, right before this PR was merged), the same repro resolves correctly and exits 0.

Bisected with `git bisect` across the 72 commits between `apiology/solargraph@8fda6338` (merge-base with `castwide/solargraph` master) and the `2026-08-04` integration branch tip (`99cc5bf`); this PR's merge commit (`73091464`) is the first bad commit, `dba36ec2` (the immediately preceding "Add PinCache: unified gem pin caching engine" PR) is the last good one.

### Suggested fix

Drop the manual `"lib/"` prefix and pass the require path directly, matching the pre-refactor behavior:

```ruby
# Determine gem name based on the require path
spec_with_path = Gem::Specification.find_by_path(require)
```

(`require_paths.rb`'s existing suffix-matching/`.rb`-appending behavior lives inside `find_by_path` itself, so no other logic here needs to change.)

I'd suggest also adding a regression spec covering a gem whose require path differs from its package name (e.g. `activesupport`/`active_support`) to `spec/workspace/gemspecs_spec.rb`, since the existing spec suite apparently didn't catch this — every currently-tested gem likely has matching require-path/package-name pairs.

lib/solargraph/api_map.rb
```diff
@@ -659,6 +649,7 @@ def locate_pins location
     # @param cursor [Source::Cursor]
     # @return [SourceMap::Clip]
     def clip cursor
+      # @sg-ignore Need to add nil check here
```
apiology: Look in rules.rb; there's something about flow sensitive typing against attribute access whose message we should use here

spec/language_server/host_spec.rb
```diff
@@ -275,7 +275,7 @@ def initialize(foo); end
       expect(symbols).not_to be_empty
     end
 
-    it 'opens a file outside of prepared libraries' do
+    it 'opens a file outside of prepared libraries', time_limit_seconds: 200 do
```
apiology: Let's move the rspec timeout changes into a new PR

lib/solargraph/api_map.rb

Retagged in 24fa7be08 to match `lib/solargraph/type_checker/rules.rb`'s
"flow sensitive typing needs to handle attrs" entry. Then confirmed it
was actually unneeded - removing it entirely and re-checking showed no
error at that line, so I dropped it rather than keep a marker for
nothing.

spec/language_server/host_spec.rb
apiology: **Claude:**

Reverted all 7 `time_limit_seconds:` additions across the 5 spec files
in 24fa7be08, and moved them to
[castwide/solargraph#1331, Add time_limit_seconds guards to 7 slow specs](https://github.com/castwide/solargraph/pull/1331).

lib/solargraph/pin/parameter.rb
```diff
@@ -95,7 +95,7 @@ def arity_decl
 
       # @return [String]
       def type_arity_decl
-        arity_decl + return_type.items.count.to_s
+        arity_decl + return_type.tags
```
apiology: Where is this used?  Is this a breaking change?  Why not rooted_tags?

lib/solargraph/workspace/gemspecs.rb
```diff
@@ -45,7 +45,7 @@ def resolve_require require
 
         # Determine gem name based on the require path
         file = "lib/#{require}.rb"
-        spec_with_path = Gem::Specification.find_by_path(file)
+        spec_with_path = Gem::Specification.find_by_path(require)
```
apiology: Is 'file' used at this point?

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -769,16 +773,17 @@ def alias_to_pin decl, closure
       #
       # @param type [RBS::MethodType]
       # @return [ComplexType]
+      # @param [Object] implicit_nil
```
apiology: Why not bool?  and a keyword arg to make this more self-describing?

lib/solargraph/pin/parameter.rb

Used on one chain only: `Callable#type_arity` -> `Method#combine_signatures_by_type_arity`, where it serves as a Hash key and an `==` comparison. Nothing serializes or exposes it, so no external break.

It does change grouping: the old `items.count` put any two 2-member unions in the same bucket, and the new form separates them by actual type.

`rooted_tags` looks like the better key. `tags` can collide across namespaces, and a collision merges signatures that should stay apart — wrong types, which costs more than the extra signatures that over-splitting would produce.

Two stale things left behind by the change:

- the `type_arity` doc comment still gives `"foo:5"` as its example and describes "the number of unique types in each parameter"; it now emits `"foo:String, Integer"`.
- `full_type_arity` still uses `return_type.items.count.to_s` for the return component while the parameters now use tags.

lib/solargraph/workspace/gemspecs.rb
apiology: **Claude:**

Yes — `file` is still used further down the same method, in the manual fallback scan:

```ruby
spec&.files&.any? { |gemspec_file| file == gemspec_file }
```

`Gem::Specification.find_by_path` matches a requirable path rather than a lib-relative filename, so `require` is the correct argument here, and `file` stays for that literal `spec.files` comparison.

lib/solargraph/rbs_map/conversions.rb
apiology: **Claude:**

Both look right. `implicit_nil` is only ever truth-tested, and the sole caller `parts_of_function` already declares it as `@param implicit_nil [Boolean]`. The tag here is also sitting after `@return`.

```ruby
# @param type [RBS::MethodType]
# @param implicit_nil [Boolean]
# @return [ComplexType]
def extract_method_type_return_type type, implicit_nil:
```

apiology: Go for it.

lib/solargraph/workspace/gemspecs.rb
apiology: Let's move the file definition to where it is used

lib/solargraph/pin/parameter.rb
apiology: Switch to rooted_tags.

I assume you'll update whatever needs to be updated as well

apiology:
**Claude:** Closing. This PR has been split into smaller, independently reviewable pieces, all open:

- castwide/solargraph#1339 — Skip expensive parser gem YARD generation when RBS collection available
- castwide/solargraph#1342 — Cache only bundle and stdlib gems by default
- castwide/solargraph#1343 — Pass the require path to find_by_path
- castwide/solargraph#1344 — Root RBS reference pin and namespace names
- castwide/solargraph#1346 — Fix crash in solargraph gems core
- apiology/solargraph#93 — Cached RBS types leak between open workspace folders
- apiology/solargraph#98 — Serve RBS stdlib pins before the gem cache builds

PR #1253, opened 2026-08-03
Remove duplicate dead methods in RbsMap::Conversions

apiology:
## Why

`RbsMap::Conversions#build_type` and `#parts_of_function` are each defined twice in this file. Ruby silently uses the later definition, so the earlier one is dead code — but it's live-looking dead code: it references two helper methods (`other_type_to_type`, `method_type_to_type`) that don't exist anywhere in `lib/`, so if anything ever called the shadowed version, it would raise `NoMethodError`. Nothing currently reaches it, but it's a landmine for the next refactor that touches this file and doesn't know to check which definition actually runs.

## How

Delete the earlier (shadowed) definitions of both methods. Also fixes a merge artifact nearby: a `# rubocop:enable` comment had gotten concatenated onto the same line as the next method's docstring, breaking both the rubocop directive and the docstring.

## What's in it

- `lib/solargraph/rbs_map/conversions.rb`: -100/+2 lines, no behavior change (dead code removed).

**Testing**: `bundle exec rspec spec/rbs_map/` — 30 examples, 0 failures, 4 pending.

### DIFF
--- a/lib/solargraph/rbs_map/conversions.rb
+++ b/lib/solargraph/rbs_map/conversions.rb
@@ -159,27 +159,6 @@ module Solargraph
         RBS_TO_CLASS.fetch(ns, ns)
       end
 
-      # @param type_name [RBS::TypeName]
-      # @param type_args [Enumerable<RBS::Types::Bases::Base>]
-      # @return [ComplexType::UniqueType]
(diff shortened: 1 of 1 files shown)


apiology:
@castwide - ready for review

PR #1258, opened 2026-08-03
Narrow repeated calls to the same attr_reader-style accessor

### LINKED ISSUE #1249: Flow-sensitive typing doesn't narrow repeated calls to the same attr_reader-style method
## Summary

Flow-sensitive typing narrows nil-checks on local variables (`x = foo; return if x.nil?; x.bar`), but does not narrow repeated calls to the same no-argument accessor method (e.g. an `attr_reader`) even when the method is a pure getter with no side effects. A nil-guard on `obj.attr` does not narrow a later call to `obj.attr` in the same expression or method body — each call is treated as an independent, unnarrowed invocation.

## Repro

```ruby
# typecheck --level strong
class Location
  # @return [String]
  def filename; end
end

class Pin
  # @return [Location, nil]
  attr_reader :location
end

# @param pin [Pin]
def bundled_filename(pin)
  # `pin.location` is nil-guarded here...
  return nil unless pin.location

  # ...but this second call to `pin.location` is still typed
  # `Location, nil`, so `.filename` is flagged as a possible nil
  # dereference even though the guard above already ruled that out.
  pin.location.filename
end
```

Typechecking this at `strong` reports a problem on `pin.location.filename`, even though `pin.location` was already confirmed non-nil on the line above.

## Context

Found while auditing `@sg-ignore` accumulation in castwide/solargraph#1240 (strong-level typecheck cleanup) — this is the largest single category there (30 occurrences), spread across `pin/base.rb`, `pin/block.rb`, `pin/callable.rb`, `type_checker.rb`, and others. Representative real examples:

```ruby
# lib/solargraph/pin/base.rb
def filename
  return nil if location.nil?
  # @sg-ignore flow sensitive typing needs to handle attrs
  location.filename
end
```

```ruby
# lib/solargraph/type_checker.rb
def internal? pin
  return false if pin.nil?
  # @sg-ignore flow sensitive typing needs to handle attrs
  pin.location && api_map.bundled?(pin.location.filename)
end
```

A reasonable fix scope: when a method is a simple, argument-free `attr_reader`-style accessor (or otherwise provably pure/idempotent), treat repeated calls to it within the same flow-sensitive-typing scope the same way a local variable holding its result would be treated for nil-narrowing purposes.

apiology:
## Summary

Flow-sensitive typing already narrows nil-checks on local/instance variables (`x = foo; return if x.nil?; x.bar`), but a nil-guard on `obj.attr` did not narrow a later call to `obj.attr` in the same method body -- each call was treated as an independent, unnarrowed invocation.

`FlowSensitiveTyping` now recognizes receivers that are a dotted chain of simple, argument-less, blockless calls rooted in a tracked local or instance variable (e.g. `pin.location`, `@pin.location`) and records nil-narrowing facts against a synthesized pin for that chain, the same way it already does for a plain variable:

```ruby
# @param pin [Pin]
def bundled_filename(pin)
  return nil unless pin.location
  pin.location.filename  # now typed Location, not Location-or-nil
end
```

`Chain::Call#resolve` looks up those facts by threading a dotted "receiver path" through `Chain#define`/`#infer`, checked before falling back to ordinary method resolution. This required adding an optional `receiver_path` parameter to `Chain::Link#resolve` and its subclasses (mechanical -- all but `Chain::Call` just accept and ignore it).

Also fixes a latent `Pin::BaseVariable#equality_fields` gap found while testing this: downcast copies of the same pin (e.g. the if-true and if-false facts from one guard) shared identical name/location/closure/source and so had identical equality_fields, despite carrying different presence/narrowed types. That let them collide as cache keys in Chain's inference cache, returning a stale, wrongly-narrowed or wrongly-unnarrowed result depending on lookup order -- this made the new narrowing intermittently flaky until fixed.

This intentionally does not cover a bare, implicit-self call (e.g. `location.nil?; location.filename` where `location` is a 0-arg method, not a local) -- narrowing that would need a way to resolve the enclosing self type at indexing time, which FlowSensitiveTyping does not currently have. Left as a natural follow-up.

Fixes #1249

## Test plan

- [x] `bundle exec rspec spec/parser/flow_sensitive_typing_spec.rb` -- new specs cover the truthy-guard, `.nil?`-guard, and ivar-rooted cases from the issue
- [x] `bundle exec rspec` -- full suite, 1621 examples, 0 failures (matches baseline)
- [x] `bundle exec rubocop` on changed files -- no new offenses vs baseline
- [x] `SOLARGRAPH_ASSERTS=on bundle exec solargraph typecheck --level strong` -- no new problems vs baseline (repo-wide strong typecheck is not currently clean and is not a hard CI gate; this PR was checked line-by-line against a baseline run to avoid adding new debt)

https://claude.ai/code/session_01KGu6zb5faStTC754PxMUSA

### DIFF
--- a/lib/solargraph/parser/flow_sensitive_typing.rb
+++ b/lib/solargraph/parser/flow_sensitive_typing.rb
@@ -250,13 +250,51 @@ module Solargraph
         process_and(expression_node, true_ranges, false_ranges)
         process_or(expression_node, true_ranges, false_ranges)
         process_variable(expression_node, true_ranges, false_ranges)
+        process_call_chain(expression_node, true_ranges, false_ranges)
+      end
+
+      # Recognizes receivers of the form 'foo', '@foo', 'foo.bar', or
+      # '@foo.bar.baz' -- a chain of simple, argument-less, blockless
+      # calls/variable references rooted in a local variable, instance
+      # variable, or (for a bare call, e.g. 'foo' referring to a 0-arg
+      # method on self) an as-yet-unresolved name.
+      #
+      # @param node [Parser::AST::Node, nil]
+      # @return [::Array<String>, nil] Dotted-word chain, e.g. ['pin',
+      #   'location'], or nil if `node` doesn't have this shape.
+      def parse_receiver_chain node
+        return unless node.is_a?(::Parser::AST::Node)
+        # @sg-ignore Need to add nil check here
+        return [node.children[0].to_s] if %i[lvar ivar].include?(node.type)
+        # @sg-ignore Need to add nil check here
+        return unless node.type == :send
+        # no arguments
+        # @sg-ignore Need to add nil check here
+        return unless node.children[2..].empty?
+
+        # @sg-ignore Need to add nil check here
+        method_name = node.children[1]
+        # @sg-ignore Need to add nil check here
+        return unless method_name.is_a?(Symbol)
+
+        # @sg-ignore Need to add nil check here
+        receiver = node.children[0]
+        # bare call, e.g. `s(:send, nil, :foo)` - implicit self, so
+        # 'foo' could be a local variable or a 0-arg method on self
+        # @sg-ignore Need to add nil check here
+        return [method_name.to_s] if receiver.nil?
+
+        base = parse_receiver_chain(receiver)
+        return unless base
+
+        base + [method_name.to_s]
       end
 
       # @param call_node [Parser::AST::Node]
       # @param method_name [Symbol]
-      # @return [Array(String, String), nil] Tuple of rgument to
-      #   function, then receiver of function if it's a variable,
-      #   otherwise nil if no simple variable receiver
+      # @return [Array(String, ::Array<String>), nil] Tuple of argument to
+      #   function, then dotted-word chain for the receiver, otherwise nil
+      #   if the receiver isn't a simple chain (see #parse_receiver_chain)
       def parse_call call_node, method_name
         return unless call_node&.type == :send && call_node.children[1] == method_name
         # Check if conditional node follows this pattern:
@@ -267,28 +305,21 @@ module Solargraph
         call_receiver = call_node.children[0]
         call_arg = type_name(call_node.children[2])
 
-        # check if call_receiver looks like this:
-        #  s(:send, nil, :foo)
-        # and set variable_name to :foo
-        if call_receiver&.type == :send && call_receiver.children[0].nil? && call_receiver.children[1].is_a?(Symbol)
-          variable_name = call_receiver.children[1].to_s
-        end
-        # or like this:
-        # (lvar :repr)
-        variable_name = call_receiver.children[0].to_s if %i[lvar ivar].include?(call_receiver&.type)
-        return unless variable_name
+        # @sg-ignore Need to add nil check here
+        chain_words = parse_receiver_chain(call_receiver)
+        return unless chain_words
 
-        [call_arg, variable_name]
+        [call_arg, chain_words]
       end
 
       # @param isa_node [Parser::AST::Node]
-      # @return [Array(String, String), nil]
+      # @return [Array(String, ::Array<String>), nil]
       def parse_isa isa_node
-        call_type_name, variable_name = parse_call(isa_node, :is_a?)
+        call_type_name, chain_words = parse_call(isa_node, :is_a?)
 
         return unless call_type_name
 
-        [call_type_name, variable_name]
+        [call_type_name, chain_words]
       end
 
       # @param variable_name [String]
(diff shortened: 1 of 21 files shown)


lib/solargraph/source/chain.rb
```diff
@@ -125,12 +132,14 @@ def define api_map, name_pin, locals
...
           # binder is now the lhs of the rhs we are evaluating.
           working_pin = Pin::ProxyType.anonymous(name_pin.context, binder: type, closure: name_pin, source: :chain)
+          receiver_path = next_receiver_path(receiver_path, link)
           logger.debug do
             "Chain#define(links=#{links.map(&:desc)}, name_pin=#{name_pin.inspect}, locals=#{locals}) - after processing #{link.desc}, new working_pin=#{working_pin} with binder #{working_pin.binder}"
           end
         end
         links.last.last_context = working_pin
-        links.last.resolve(api_map, working_pin, locals)
+        # @sg-ignore Need to add nil check here
```
apiology: Why is this annotation necessary?

apiology:
@castwide - ready for review

lib/solargraph/source/chain.rb
apiology: **Claude:** Fixed for real in 9445d1f: extracted `links.last` into a local once and added an explicit `return [] if last_link.nil?` guard, instead of suppressing the warning. It's genuinely unreachable (`Chain#initialize` pads an empty `links` array with `UNDEFINED_CALL`, so `links` is never empty), but that invariant should be expressed in the code, not asserted away with a comment. No `@sg-ignore` needed now.

(Also tried teaching `always_leaves_compound_statement?` to recognize a bare `raise` the same way it already recognizes `return`/`next`/`retry` - it currently doesn't, since `raise` is a plain method call in the AST rather than a dedicated node type. Backed that out of this PR since it's already handled in another PR.)

lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -250,13 +250,49 @@ def process_expression expression_node, true_ranges, false_ranges
...
         process_or(expression_node, true_ranges, false_ranges)
         process_variable(expression_node, true_ranges, false_ranges)
+        process_call_chain(expression_node, true_ranges, false_ranges)
+      end
+
+      # Recognizes receivers of the form 'foo', '@foo', 'foo.bar', or
+      # '@foo.bar.baz' -- a chain of simple, argument-less, blockless
+      # calls/variable references rooted in a local variable, instance
+      # variable, or (for a bare call, e.g. 'foo' referring to a 0-arg
+      # method on self) an as-yet-unresolved name.
```
apiology: 3 lines

lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -250,13 +250,49 @@ def process_expression expression_node, true_ranges, false_ranges
...
+
+        receiver = node.children[0]
+        # bare call, e.g. `s(:send, nil, :foo)` - implicit self, so
+        # 'foo' could be a local variable or a 0-arg method on self
+        return [method_name.to_s] if receiver.nil?
+
+        # @sg-ignore a :send node's receiver (children[0]) is nil or an
+        #   AST::Node, never any of the other types children[] can hold
+        #   for other node shapes (e.g. Symbol, Array) -- and
+        #   parse_receiver_chain re-checks node.is_a?(...) itself anyway
```
apiology: Conform to rules.rb standard

lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -267,28 +303,24 @@ def parse_call call_node, method_name
...
-          variable_name = call_receiver.children[1].to_s
-        end
-        # or like this:
-        # (lvar :repr)
-        variable_name = call_receiver.children[0].to_s if %i[lvar ivar].include?(call_receiver&.type)
-        return unless variable_name
+        # @sg-ignore node.children is typed as a broad union (it can hold
+        #   nested arrays/symbols for other node shapes), but
+        #   parse_receiver_chain re-checks node.is_a?(::Parser::AST::Node)
+        #   itself and safely returns nil for anything else
```
apiology: Conform to rules.rb standard

lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -307,18 +339,52 @@ def find_var variable_name, position
...
+      # Finds (for a single tracked local/instance variable) or builds
+      # (for a chain of simple calls off of one, e.g. ['pin', 'location'])
+      # the pin flow-sensitive-typing facts should be recorded against.
+      #
+      # A synthesized pin's type is computed lazily, from `node` itself,
+      # by Pin::BaseVariable#probe the same way a real local variable's
+      # type is computed from its assignment node -- since `node` is the
+      # receiver expression at *this*, necessarily earlier, guard site,
+      # re-inferring it can never see the narrowing facts this same pin
+      # is about to be asked to carry.
```
apiology: 3 lines

lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -433,6 +500,47 @@ def process_variable node, true_presences, false_presences
         process_facts(if_false, false_presences)
       end
 
+      # Handles a bare truthy check on a call chain, e.g. 'pin.location'
+      # in 'return nil unless pin.location'. Bare references to a single
+      # local/instance variable are handled by #process_variable instead;
+      # this only fires once there's an explicit receiver (chain_words
+      # has more than one word).
```
apiology: 1-2 line

lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -433,6 +500,47 @@ def process_variable node, true_presences, false_presences
...
+        # already handled (with inverted true/false semantics) by
+        # process_nilp/process_bang
+        return if %i[nil? !].include?(node.children[1])
+
+        chain_words = parse_receiver_chain(node)
+        return if chain_words.nil? || chain_words.length < 2
+
+        # @sg-ignore Range.from_node is nil only for a node without
+        #   source location info, which doesn't happen for real parsed
+        #   nodes reaching here (same as isa_position/nilp_position above)
```
apiology: Conform to rules.rb

lib/solargraph/pin/base_variable.rb
```diff
@@ -293,6 +293,18 @@ def visible_at? other_closure, other_loc
       # @return [Range]
       attr_writer :presence
 
+      # Flow-sensitive typing downcasts a variable/call pin into multiple
+      # copies that otherwise share name/location/closure/source (see
+      # FlowSensitiveTyping#add_downcast_var) -- they must stay distinct
+      # by presence and narrowed type, or callers that key off of pin
+      # equality (e.g. Chain's inference cache) can conflate two pins
+      # that carry different narrowing facts.
```
apiology: 1 line only

lib/solargraph/source/chain/block_symbol.rb
```diff
@@ -4,7 +4,11 @@ module Solargraph
   class Source
     class Chain
       class BlockSymbol < Link
-        def resolve api_map, name_pin, locals
+        # @param api_map [ApiMap]
+        # @param name_pin [Pin::Base]
+        # @param locals [::Array<Pin::Base>]
+        # @param _receiver_path [::Array<String>, nil]
+        def resolve api_map, name_pin, locals, _receiver_path = nil
```
apiology: Can't this come from name_pin?

lib/solargraph/source/chain/array.rb
```diff
@@ -18,7 +18,8 @@ def word
         # @param api_map [ApiMap]
         # @param name_pin [Pin::Base]
         # @param locals [::Array<Pin::Parameter, Pin::LocalVariable>]
-        def resolve api_map, name_pin, locals
+        # @param _receiver_path [::Array<String>, nil]
+        def resolve api_map, name_pin, locals, _receiver_path = nil
```
apiology: Can't this come from name_pin?

lib/solargraph/source/chain/block_variable.rb
```diff
@@ -4,7 +4,11 @@ module Solargraph
   class Source
     class Chain
       class BlockVariable < Link
-        def resolve api_map, name_pin, locals
+        # @param api_map [ApiMap]
+        # @param name_pin [Pin::Base]
+        # @param locals [::Array<Pin::Base>]
+        # @param _receiver_path [::Array<String>, nil]
+        def resolve api_map, name_pin, locals, _receiver_path = nil
```
apiology: Can't this come from name_pin?

PR #1259, opened 2026-08-04
Fix raise/fail nil guards, root-scoped is_a?, case/when, ||=, and namespace-scope narrowing in flow-sensitive typing

### LINKED ISSUE #1241: Type checker doesn't narrow a case/when subject's type inside each branch
## Summary

When a `case`/`when` subject's static type is a union (e.g. an RBS-alias-derived union like `RBS::Types::t`), the type checker doesn't narrow the subject's type to the matched `when` class inside that branch's body. Every branch still sees the subject typed as the full original union, so any call to a method that only exists on some members of the union (not all of them) is flagged as unresolved/unavailable, even though the `when` clause has already established which member type it is.

## Repro

```ruby
# typecheck --level strong
require 'rbs'

# @param type [RBS::Types::t]
def describe(type)
  case type
  when RBS::Types::Optional
    # `type` is still typed as RBS::Types::t here, not narrowed to
    # RBS::Types::Optional, so #type is "unresolved":
    type.type
  when RBS::Types::Tuple
    # same problem - `type` isn't narrowed to RBS::Types::Tuple:
    type.types
  end
end
```

Typechecking this at `strong` reports `Unresolved call to type` / `Unresolved call to types` on the lines above, even though each is only reachable when `type` has already matched the corresponding class in the `when` clause.

## Context

Found while working on `lib/solargraph/rbs_translator.rb`'s `type_to_tag` (in #1223), which dispatches on RBS's own `RBS::Types::t` union via a `case`/`when` over every member class (`RBS::Types::Optional`, `RBS::Types::Tuple`, `RBS::Types::Literal`, `RBS::Types::Union`, etc.) and calls type-specific accessors inside each branch. Without narrowing, every one of those accessor calls needs an explanatory `@sg-ignore`, which is a lot of noise for what's normal, idiomatic Ruby type-dispatch code.

## Expected

Inside a `when SomeClass` branch (or an `is_a?`/`kind_of?` guard), the subject's type should narrow to the intersection of its prior type and `SomeClass`, the same way flow-sensitive typing already narrows on `is_a?` checks used with `if`/`unless`.

### LINKED ISSUE #1251: Flow-sensitive typing doesn't narrow is_a? checks combined with && or inside elsif branch bodies
## Summary

Flow-sensitive typing can narrow a variable's type from an `is_a?` check in some contexts (per the original "automatically downcast from is_a? calls" feature), but two common shapes still aren't narrowed:

1. An `is_a?` check combined with a second condition via `&&` (or a `next unless`/`return unless` guard using that combination) doesn't narrow the checked expression for later use in the same statement or the following code.
2. An `elsif x.is_a?(SomeClass) && ...` condition doesn't narrow `x` inside that branch's own body.

This is a different, narrower-scoped gap than castwide/solargraph#1241 (which is specifically about `case`/`when` subjects) — this is about `if`/`elsif`/guard-clause conditions built from `is_a?`.

## Repro

```ruby
# typecheck --level strong
# @param node [Object]
def describe(node)
  # `is_ast_node?` is itself an `is_a?`-style predicate
  next unless node.is_a?(::Parser::AST::Node) && node.type == :hash

  # `node` is still typed as `Object` here, not narrowed to
  # `Parser::AST::Node`, so `#children` is "unresolved":
  node.children
end

def describe2(node)
  if node.is_a?(String)
    node.upcase
  elsif node.is_a?(::Parser::AST::Node) && node.type == :casgn
    # `node` is still typed as `Object` in this branch's body too -
    # the narrowing established by the `elsif` condition doesn't
    # carry into the branch:
    node.children[0]
  end
end
```

## Context

Found while auditing `@sg-ignore` accumulation in castwide/solargraph#1240 (strong-level typecheck cleanup) — 12 occurrences across `parser/parser_gem/node_methods.rb` and `parser/parser_gem/node_processors/sclass_node.rb`. Representative real examples:

```ruby
# lib/solargraph/parser/parser_gem/node_methods.rb
node.children.each do |pair|
  next unless Parser.is_ast_node?(pair) && pair.children[0]
  # @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
  result[pair.children[0].children[0]] = simple_convert(pair.children[1])
end
```

```ruby
# lib/solargraph/parser/parser_gem/node_processors/sclass_node.rb
# @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
elsif sclass.is_a?(::Parser::AST::Node) && sclass.type == :casgn
  names = [region.closure.namespace, region.closure.name]
  # @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
  if sclass.children[0].nil? && names.last != sclass.children[1].to_s
```

There's also a disabled (commented-out) block in `lib/solargraph/source/chain/literal.rb` — left off since #1201 — with 4 more `is_a?`-narrowing notes of the same shape, which would become live again if that block is ever re-enabled.

### LINKED ISSUE #1254: Flow-sensitive typing doesn't narrow past a raise/return-based nil guard
**Claude:**

## Summary

Flow-sensitive typing does not narrow a variable's type past a `raise`/`return`-based nil guard — i.e. `raise "..." if x.nil?` or `return ... if x.nil?` (or the `unless` inverse), after which `x` is used either as a call argument or as the receiver of a call. At `strong` level, `x` continues to report as nilable in both cases, even though the guard has already eliminated `nil` for the rest of the method.

This is a different, narrower-scoped gap than the already-filed flow-sensitivity issues:
- #1241 — `case`/`when` subject narrowing
- #1249 — repeated calls to the same attr_reader-style method
- #1250 — reassignment to a different (non-literal) type
- #1251 — `is_a?` checks combined with `&&` or inside `elsif` branch bodies

None of those cover a `raise`/`return`-based early-exit guard specifically. It's also the pattern named but not individually tracked in #1240's description ("flow-sensitive-typing limits (postfix nil guards, ...)").

## Repro

```ruby
# typecheck --level strong

# @param x [String, nil]
# @return [void]
def process(x)
  raise 'x required' if x.nil?

  # `x` is still typed as `String, nil` here, not narrowed to `String` -
  # passing it to a String-only parameter is flagged even though the
  # raise above already eliminates the nil case for the rest of the method:
  accepts_string(x)
end

# @param s [String]
# @return [void]
def accepts_string(s)
  puts s
end

# @param x [String, nil]
# @return [Integer]
def length_of(x)
  return 0 if x.nil?

  # Same gap when `x` is used as a call *receiver* instead of an argument -
  # still typed as `String, nil` here too:
  x.length
end
```

Both `process` and `length_of` report a type mismatch on the guarded variable despite the preceding guard making that case unreachable.

## Context

Found while auditing `@sg-ignore` accumulation in a downstream app (apiology/plate-spinner) — this single pattern accounts for 15 of the app's `@sg-ignore` markers across 10 files, split roughly evenly between the argument-position and receiver-position shapes above. It's also not reliably escapable with the usual `@type` cast workaround: at least 2 of those 15 occurrences still report the mismatch even after adding an explicit `@type` cast on a local immediately following the guard.

apiology:
## Summary

Five related flow-sensitive-typing gaps, fixed together since they touch the same file and spec.

### raise/fail-based nil guards (fixes castwide/solargraph#1254)
- The parser gem parses `raise`/`fail` as a plain `:send` node, not a `:raise` node type, so `always_leaves_compound_statement?` never recognized a raise-based guard and it failed to narrow, unlike an equivalent `return`-based guard. Fix recognizes the `:send` shape (no receiver, method name `:raise`/`:fail`) alongside the existing keyword-based node types.

```ruby
x = maybe_nil        # String, nil
raise 'bad' unless x
x.upcase              # String
```

### root-scoped (::-prefixed) constants in is_a? narrowing (fixes castwide/solargraph#1251)
- `type_name` in `FlowSensitiveTyping` didn't recognize the `:cbase` node the parser gem emits for a leading `::` on a constant (`::Foo` parses as `s(:const, s(:cbase), :Foo)`), so any fully-qualified class name silently disabled `is_a?`-based narrowing. Fix recognizes `:cbase` and renders it as a leading `::`, recursing normally otherwise.

```ruby
x = value                    # Object, nil
raise unless x.is_a?(::Hash)
x.fetch(:k)                   # ::Hash
```

### case/when subject narrowing (fixes castwide/solargraph#1241)
- case/when had no flow-sensitive-typing support at all -- no NodeProcessor was registered for `:case` nodes, so a case subject kept its full (often union) static type in every branch.
- Fix adds a `CaseNode` processor and `FlowSensitiveTyping#process_case`, narrowing the subject to the union of each `when` clause's constant classes, scoped to that branch body only.

```ruby
case x                      # Foo, Bar
when Foo then x.foo_only    # Foo
end
```

### x ||= value on lvars/ivars
- `x ||= value` only assigns when `x` is falsy, so a prior non-nil value should survive -- but `OrasgnNode` rewrote it as a plain `x = value`, and plain-reassignment pins union together rather than override, so the narrowed type was lost.
- Fix adds `FlowSensitiveTyping#process_or_asgn`, reusing the same downcast-pin machinery that already powers `is_a?`/`nil?` narrowing to exclude `nil` from the prior type, scoped to the rest of the enclosing closure; no `@sg-ignore` comments remain anywhere in this fix.
- Conservative and scoped to `lvasgn`/`ivasgn` targets -- doesn't union in the RHS type when it differs from the variable's prior non-nil type.

```ruby
@x = nil     # String, nil
@x ||= compute
@x.upcase     # String
```

### is_a?/nil? guard narrowing at file and class-body scope
- `FlowSensitiveTyping#process_if` extends narrowing past a guard via `enclosing_compound_statement_pin.node`, but the root `Pin::Namespace` (`NodeProcessor.process`, one per file) and the per-class/module `Pin::Namespace` (`NamespaceNode`) never passed `node:`, so that pin's node was always `nil` and narrowing silently never applied there -- unlike `Pin::Method`/`Pin::Block`, which already pass `node:`.
- Fix passes `node:` through at both sites, matching the existing Method/Block pattern; the gap is specific to namespace-shaped compound statements (file root, class/module bodies), not "top-level" generally -- a block at the top level already narrowed correctly. Known follow-on, not fixed here: `SclassNode` (`class << self`) has the same missing `node:` (unverified), and `Pin::Base#combine_with`/`Closure#combine_with` don't carry `node:` forward across a reopened class merged from multiple files (doesn't affect this fix, since narrowing runs before combination).

```ruby
loaded = load_config          # Hash, nil
return unless loaded.is_a?(Hash)
loaded['key']                  # Hash
```

### DIFF
--- a/lib/solargraph/parser/node_processor.rb
+++ b/lib/solargraph/parser/node_processor.rb
@@ -40,9 +40,16 @@ module Solargraph
       # @return [Array(Array<Pin::Base>, Array<Pin::LocalVariable>, Array<Pin::InstanceVariable>)]
       def self.process node, region = Region.new, pins = [], locals = [], ivars = []
         if pins.empty?
+          # node: is required so flow-sensitive typing can find the
+          # end of this compound statement (see
+          # FlowSensitiveTyping#process_if) - without it, an
+          # is_a?/nil? guard at the top level of a file never
+          # narrows, because enclosing_compound_statement_pin.node
+          # is always nil.
           pins.push Pin::Namespace.new(
             location: region.source.location,
             name: '',
+            node: node,
             source: :parser
           )
         end
--- a/lib/solargraph/parser/node_processor.rb
+++ b/lib/solargraph/parser/node_processor.rb
@@ -40,9 +40,6 @@ module Solargraph
       # @return [Array(Array<Pin::Base>, Array<Pin::LocalVariable>, Array<Pin::InstanceVariable>)]
       def self.process node, region = Region.new, pins = [], locals = [], ivars = []
         if pins.empty?
-          # node: lets flow-sensitive typing find this statement's
-          # end (FlowSensitiveTyping#process_if); without it,
-          # top-level is_a?/nil? guards never narrow.
           pins.push Pin::Namespace.new(
             location: region.source.location,
             name: '',
--- a/lib/solargraph/parser/node_processor.rb
+++ b/lib/solargraph/parser/node_processor.rb
@@ -40,12 +40,9 @@ module Solargraph
       # @return [Array(Array<Pin::Base>, Array<Pin::LocalVariable>, Array<Pin::InstanceVariable>)]
       def self.process node, region = Region.new, pins = [], locals = [], ivars = []
         if pins.empty?
-          # node: is required so flow-sensitive typing can find the
-          # end of this compound statement (see
-          # FlowSensitiveTyping#process_if) - without it, an
-          # is_a?/nil? guard at the top level of a file never
-          # narrows, because enclosing_compound_statement_pin.node
-          # is always nil.
+          # node: lets flow-sensitive typing find this statement's
+          # end (FlowSensitiveTyping#process_if); without it,
+          # top-level is_a?/nil? guards never narrow.
           pins.push Pin::Namespace.new(
             location: region.source.location,
             name: '',
--- a/lib/solargraph/parser/parser_gem/node_chainer.rb
+++ b/lib/solargraph/parser/parser_gem/node_chainer.rb
@@ -107,9 +107,13 @@ module Solargraph
             # s(:or_asgn,
             #   s(:ivasgn, :@bar),
             #   s(:int, 123))
+            or_asgn_rhs_node = n.children[1] # s(:int, 123)
             lhs_chain = NodeChainer.chain n.children[0] # s(:ivasgn, :@bar)
-            rhs_chain = NodeChainer.chain n.children[1] # s(:int, 123)
-            or_link = Chain::Or.new([lhs_chain, rhs_chain])
+            # @sg-ignore Need to add nil check here
+            rhs_chain = NodeChainer.chain or_asgn_rhs_node
+            # @sg-ignore Need to add nil check here
+            or_asgn_rhs_never_returns = always_leaves_compound_statement?(or_asgn_rhs_node)
+            or_link = Chain::Or.new([lhs_chain, rhs_chain], rhs_never_returns: or_asgn_rhs_never_returns)
             # this is just for a call chain, so we don't need to record the assignment
             result.push(or_link)
           elsif %i[class module def defs].include?(n.type)
@@ -118,8 +122,14 @@ module Solargraph
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
-            result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
-                                       NodeChainer.chain(n.children[1], @filename, n)])
+            or_rhs_node = n.children[1]
+            # @sg-ignore Need to add nil check here
+            or_lhs_chain = NodeChainer.chain(n.children[0], @filename)
+            # @sg-ignore Need to add nil check here
+            or_rhs_chain = NodeChainer.chain(or_rhs_node, @filename, n)
+            # @sg-ignore Need to add nil check here
+            or_rhs_never_returns = always_leaves_compound_statement?(or_rhs_node)
+            result.push Chain::Or.new([or_lhs_chain, or_rhs_chain], rhs_never_returns: or_rhs_never_returns)
           elsif n.type == :if
             then_clause = if n.children[1]
                             NodeChainer.chain(n.children[1], @filename, n)
--- a/lib/solargraph/parser/parser_gem/node_chainer.rb
+++ b/lib/solargraph/parser/parser_gem/node_chainer.rb
@@ -108,10 +108,11 @@ module Solargraph
             #   s(:ivasgn, :@bar),
             #   s(:int, 123))
             or_asgn_rhs_node = n.children[1] # s(:int, 123)
+            raise "Or-assignment node missing right-hand side: #{n}" if or_asgn_rhs_node.nil?
+
             lhs_chain = NodeChainer.chain n.children[0] # s(:ivasgn, :@bar)
-            # @sg-ignore Need to add nil check here
+            # @sg-ignore flow sensitive typing needs to handle 'raise if'
             rhs_chain = NodeChainer.chain or_asgn_rhs_node
-            # @sg-ignore Need to add nil check here
             or_asgn_rhs_never_returns = always_leaves_compound_statement?(or_asgn_rhs_node)
             or_link = Chain::Or.new([lhs_chain, rhs_chain], rhs_never_returns: or_asgn_rhs_never_returns)
             # this is just for a call chain, so we don't need to record the assignment
--- a/lib/solargraph/parser/parser_gem/node_chainer.rb
+++ b/lib/solargraph/parser/parser_gem/node_chainer.rb
@@ -107,11 +107,8 @@ module Solargraph
             # s(:or_asgn,
             #   s(:ivasgn, :@bar),
             #   s(:int, 123))
-            or_asgn_rhs_node = n.children[1] # s(:int, 123)
-            raise "Or-assignment node missing right-hand side: #{n}" if or_asgn_rhs_node.nil?
-
-            lhs_chain = NodeChainer.chain n.children[0] # s(:ivasgn, :@bar)
-            # @sg-ignore flow sensitive typing needs to handle 'raise if'
+            or_asgn_rhs_node = n.children.fetch(1) # s(:int, 123)
+            lhs_chain = NodeChainer.chain n.children.fetch(0) # s(:ivasgn, :@bar)
             rhs_chain = NodeChainer.chain or_asgn_rhs_node
             or_asgn_rhs_never_returns = always_leaves_compound_statement?(or_asgn_rhs_node)
             or_link = Chain::Or.new([lhs_chain, rhs_chain], rhs_never_returns: or_asgn_rhs_never_returns)
@@ -123,12 +120,9 @@ module Solargraph
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
-            or_rhs_node = n.children[1]
-            # @sg-ignore Need to add nil check here
-            or_lhs_chain = NodeChainer.chain(n.children[0], @filename)
-            # @sg-ignore Need to add nil check here
+            or_rhs_node = n.children.fetch(1)
+            or_lhs_chain = NodeChainer.chain(n.children.fetch(0), @filename)
             or_rhs_chain = NodeChainer.chain(or_rhs_node, @filename, n)
-            # @sg-ignore Need to add nil check here
             or_rhs_never_returns = always_leaves_compound_statement?(or_rhs_node)
             result.push Chain::Or.new([or_lhs_chain, or_rhs_chain], rhs_never_returns: or_rhs_never_returns)
           elsif n.type == :if
--- a/lib/solargraph/parser/parser_gem/node_methods.rb
+++ b/lib/solargraph/parser/parser_gem/node_methods.rb
@@ -181,12 +181,8 @@ module Solargraph
           result
         end
 
-        # Determines whether a clause necessarily leaves its
-        # enclosing compound statement (method body, block, etc.) -
-        # e.g., via an explicit 'return', a loop-control keyword, or
-        # a raise/fail call - so code that depends on execution
-        # actually reaching the statements after it can treat those
-        # statements as unreachable from this clause.
+        # Whether clause_node necessarily leaves its enclosing
+        # compound statement (return/loop-control/raise/fail).
         #
         # @param clause_node [Parser::AST::Node, nil]
         # @return [Boolean]
@@ -195,18 +191,12 @@ module Solargraph
           return true if %i[return next redo retry].include?(clause_node&.type)
           return false if clause_node.nil?
 
-          # A clause with more than one statement (e.g., a raise
-          # preceded by other statements building the error message)
-          # parses as a :begin node - only its last child determines
-          # whether the clause leaves.
+          # A multi-statement clause parses as :begin; only its last child leaves.
           return always_leaves_compound_statement?(clause_node.children.last) if clause_node.type == :begin
 
           return false unless clause_node.type == :send
 
-          # Unlike return/next/redo/retry, `raise` and `fail` are plain
-          # method calls to the parser - `raise 'msg'` parses as
-          # s(:send, nil, :raise, s(:str, "msg")), not a dedicated node
-          # type - so they need to be recognized by shape instead of type.
+          # raise/fail are plain :send calls, not a dedicated node type.
           clause_node.children[0].nil? && %i[raise fail].include?(clause_node.children[1])
         end
 
--- a/lib/solargraph/parser/parser_gem/node_methods.rb
+++ b/lib/solargraph/parser/parser_gem/node_methods.rb
@@ -181,8 +181,7 @@ module Solargraph
           result
         end
 
-        # Whether clause_node necessarily leaves its enclosing
-        # compound statement (return/loop-control/raise/fail).
+        # Whether clause_node necessarily leaves its enclosing compound statement.
         #
         # @param clause_node [Parser::AST::Node, nil]
         # @return [Boolean]
--- a/lib/solargraph/parser/parser_gem/node_methods.rb
+++ b/lib/solargraph/parser/parser_gem/node_methods.rb
@@ -181,6 +181,35 @@ module Solargraph
           result
         end
 
+        # Determines whether a clause necessarily leaves its
+        # enclosing compound statement (method body, block, etc.) -
+        # e.g., via an explicit 'return', a loop-control keyword, or
+        # a raise/fail call - so code that depends on execution
+        # actually reaching the statements after it can treat those
+        # statements as unreachable from this clause.
+        #
+        # @param clause_node [Parser::AST::Node, nil]
+        # @return [Boolean]
+        def always_leaves_compound_statement? clause_node
+          # https://docs.ruby-lang.org/en/2.2.0/keywords_rdoc.html
+          return true if %i[return next redo retry].include?(clause_node&.type)
+          return false if clause_node.nil?
+
+          # A clause with more than one statement (e.g., a raise
+          # preceded by other statements building the error message)
+          # parses as a :begin node - only its last child determines
+          # whether the clause leaves.
+          return always_leaves_compound_statement?(clause_node.children.last) if clause_node.type == :begin
+
+          return false unless clause_node.type == :send
+
+          # Unlike return/next/redo/retry, `raise` and `fail` are plain
+          # method calls to the parser - `raise 'msg'` parses as
+          # s(:send, nil, :raise, s(:str, "msg")), not a dedicated node
+          # type - so they need to be recognized by shape instead of type.
+          clause_node.children[0].nil? && %i[raise fail].include?(clause_node.children[1])
+        end
+
         # @param node [Parser::AST::Node]
         def splatted_hash? node
           Parser.is_ast_node?(node.children[0]) && node.children[0].type == :kwsplat
--- a/lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
@@ -24,6 +24,10 @@ module Solargraph
               comments: comments,
               visibility: :public,
               gates: region.closure.gates.freeze,
+              # node: lets flow-sensitive typing find the end of this
+              # class/module body - see the same node: in
+              # NodeProcessor.process's root namespace pin.
+              node: node,
               source: :parser
             )
             pins.push nspin
--- a/lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
@@ -24,9 +24,6 @@ module Solargraph
               comments: comments,
               visibility: :public,
               gates: region.closure.gates.freeze,
-              # node: lets flow-sensitive typing find the end of this
-              # class/module body - see the same node: in
-              # NodeProcessor.process's root namespace pin.
               node: node,
               source: :parser
             )
--- a/lib/solargraph/source/chain/or.rb
+++ b/lib/solargraph/source/chain/or.rb
@@ -40,7 +40,7 @@ module Solargraph
 
         protected
 
-        # @sg-ignore Fix "Not enough arguments to Module#protected"
+        # @sg-ignore return type could not be inferred
         def equality_fields
           # @sg-ignore literal arrays in this module turn into ::Solargraph::Source::Chain::Array
           super + [@links, @rhs_never_returns]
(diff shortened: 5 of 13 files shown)


apiology:
_Comment drafted by Claude (Anthropic), posted on behalf of @apiology while reviewing this branch._

## Found while running this branch against a real codebase: pre-existing malformed `Hash{}` annotations now crash the whole typecheck run

While using this branch to bulk-audit `@sg-ignore` comments in a downstream project, `solargraph typecheck` crashed partway through the workspace with:

```
Solargraph::ComplexTypeError: Bad hash type: name=Hash, substring=<String, nil, Enumerable<SomeClass>> - must have exactly two parameters
    unique_type.rb:69:in `parse'
```

**Root cause is independent of this PR's own narrowing logic** — it's a pre-existing type-string-generation bug in `ComplexType::TypeMethods#generate_substring_from` that this PR's changes happen to newly exercise (see the follow-up comment for that connection). The defect itself, reproduced directly against the library with no typecheck pipeline involved:

```ruby
require "solargraph"

ct = Solargraph::ComplexType.parse("Hash{String, nil => Enumerable<Integer>}")
ut = ct.items.first
puts "hash_parameters?: #{ut.hash_parameters?}"   # => true
puts "key_types: #{ut.key_types.map(&:tag)}"      # => ["String", "nil"]
puts "subtypes: #{ut.subtypes.map(&:tag)}"        # => ["Enumerable<Integer>"]
```

The annotation `Hash{String, nil => Enumerable<Integer>}` — a `Hash{}` whose **key type is itself an unparenthesized union** (`String, nil`) — parses successfully, but `key_types` ends up with **2 entries** instead of 1 (each union member wrapped as its own `ComplexType`). That's already surprising: `all_params` (`key_types + subtypes`) is `["String", "nil", "Enumerable<Integer>"]`, 3 elements, for what should be a 2-slot `Hash<K, V>`.

`generate_substring_from` (`lib/solargraph/complex_type/type_methods.rb:180-203`) renders correctly *only* while `hash_parameters?` is true:

```ruby
elsif hash_parameters?
  "{#{key_types_str} => #{subtypes_str}}"
...
elsif name == 'Hash'
  "<#{key_types_str}, #{subtypes_str}>"
```

But any code path that reconstructs this type's tag string without that flag set (anything that rebuilds it via `key_types`/`subtypes` rather than reusing the original parsed object) falls into the `name == 'Hash'` branch — which for this type necessarily produces `Hash<String, nil, Enumerable<Integer>>`, a 3-parameter string. Re-parsing that string is what raises "must have exactly two parameters", since the `:list`-syntax branch that handles `Hash<A, B>` (as opposed to `Hash{A => B}`) hard-requires exactly 2.

I wasn't able to pin the *exact* live call path that loses the flag (traced it toward `resolve_generics`'s handling of `all_params`, given this PR's diff to that method, but couldn't reproduce it via direct calls to `resolve_generics` in isolation — it degraded gracefully in my attempts rather than reproducing the 3-param string). What I can show directly, independent of any specific caller: **a `Hash{}` type with an unparenthesized union key has `key_types.length > 1`, and `generate_substring_from`'s `Hash<...>` fallback format is only valid when `key_types.length == 1`.** That mismatch is a real, demonstrable bug regardless of which of the several places that read `all_params`/rebuild tags actually triggers it in a given codebase.

Whether that's worth hardening (e.g. `generate_substring_from` explicitly handling `key_types.length != 1` for the `Hash` fallback branch, or the original `:hash` parse branch normalizing a multi-item key into a single unioned `ComplexType` instead of leaving it split) is up to you — happy to help narrow it further if useful.

apiology:
_Comment drafted by Claude (Anthropic), posted on behalf of @apiology while reviewing this branch._

## Separately: no error-handling boundary around per-call type inference in `TypeChecker`

Related to the crash in my previous comment, but a distinct issue worth flagging on its own: whatever eventually triggers `ComplexTypeError` for a single bad type, **the blast radius is the entire `solargraph typecheck` run for the whole workspace**, not just the one file/line with the problem.

The full trace from the crash:
```
Solargraph::ComplexTypeError: Bad hash type: ... (Solargraph::ComplexTypeError)
    from complex_type.rb:711:in `finish_atom'
    from complex_type.rb:635:in `parse_type_string'
    from complex_type.rb:509:in `block in parse'
    from api_map.rb:575:in `get_method_stack'
    from source/chain/call.rb:63:in `block in resolve'
    from complex_type/type_methods.rb:245:in `each_unique_type'
    from source/chain/call.rb:61:in `resolve'
    from source/chain.rb:133:in `define'
    from source/chain.rb:163:in `infer_uncached'
    from source/chain.rb:151:in `infer'
    from type_checker.rb:344:in `block in call_problems'
    from type_checker.rb:323:in `call_problems'
    from type_checker.rb:92:in `problems'
```

`TypeChecker#problems` ([`type_checker.rb:87-95`](https://github.com/castwide/solargraph/blob/c666672b4e6ebc8ff7b160533674f3d81c418fb1/lib/solargraph/type_checker.rb#L87-L95)) is the per-file aggregator that's supposed to collect an `Array<Problem>` — one entry per issue, so a typecheck run degrades gracefully line-by-line. `call_problems` ([`type_checker.rb:321`](https://github.com/castwide/solargraph/blob/c666672b4e6ebc8ff7b160533674f3d81c418fb1/lib/solargraph/type_checker.rb#L321) onward) iterates every call node in a file and calls `chain.infer(api_map, closure_pin, locals)` per call with no rescue around it. Any exception raised deep inside that inference — `ComplexTypeError` here, but this boundary is equally open to any other unexpected internal error — propagates all the way up through `problems` and aborts the CLI's `typecheck` command entirely.

Concretely: one malformed annotation anywhere in a large codebase currently means `solargraph typecheck` silently stops reporting *anything* for every file processed after that point in the scan, with no indication to the user that this happened (no error summary, no "N files skipped" — the process just exits non-zero after printing however many diagnostics it collected before the crash). On a real project this meant several files' worth of legitimate typecheck findings looked like "0 problems" simply because the scan never reached them.

Suggest wrapping the `chain.infer(...)` call in `call_problems` (and possibly the analogous per-tag walks in `method_tag_problems`/`variable_type_tag_problems`) in a `rescue Solargraph::ComplexTypeError` (or `StandardError` more broadly) that converts it into a `Problem` for that specific call site — something like "internal error inferring type for this expression" — rather than letting it kill the whole run. That way a single bad annotation degrades to one reported problem instead of taking down typechecking for the rest of the project.

Happy to put up a PR for this if it'd be useful, once the exact rescue scope is agreed on.

apiology:
**Claude:**

Tested this fix (58db0746, "Fixes #1254") against apiology/checkoff's real `@sg-ignore tool-limitation:type-narrowing` markers. It only covers an if-branch whose body is exactly one `raise`/`fail` statement. A branch with more than one statement -- very commonly, computing something for the error message before raising -- still doesn't narrow:

```ruby
# typed: true
# frozen_string_literal: true

class Example
  # @param s [String, nil]
  # @return [String]
  def single_statement_raise(s)
    if s.nil?
      raise 'not found'
    end
    s
  end

  # @param s [String, nil]
  # @return [String]
  def multi_statement_raise(s)
    if s.nil?
      valid = %w[a b c]
      raise "not found. Valid: #{valid.inspect}"
    end
    s
  end
end

# solargraph typecheck --level strong reports only one error, on
# multi_statement_raise -- single_statement_raise (identical shape,
# minus one extra statement before the raise) narrows correctly:
#
#   Declared return type ::String does not match inferred type
#   ::String, nil for Example#multi_statement_raise
```

Root cause: `always_leaves_compound_statement?` (`lib/solargraph/parser/flow_sensitive_typing.rb:706`) does:
```ruby
return false unless clause_node.type == :send
```
A multi-statement branch body parses as a `:begin` node, not `:send`, so this returns `false` immediately without ever checking whether the `:begin` node's last child is a raise/fail call.

apiology:
**Claude:**

Separately: `x || raise(...)` doesn't narrow at all, with or without this fix -- a different code path entirely.

```ruby
# typed: true
# frozen_string_literal: true

class Example
  # @param argv [Array<String>]
  # @return [String]
  def first_arg(argv)
    argv[0] || raise('missing first argument')
  end
end

# solargraph typecheck --level strong:
#   Example#first_arg return type could not be inferred
```

`always_leaves_compound_statement?` is only invoked from `if`-node handling (`then_clause`/`else_clause`, `flow_sensitive_typing.rb:135` and `:137`). A bare `x || raise(...)` is an `:or` node, not an `:if` node, so it never reaches that function regardless of the fix above. `||`-based nil-guards appear to have no flow-sensitive-typing support at all.

apiology:
_Comment drafted by Claude (Anthropic), posted on behalf of @apiology while reviewing this branch._

Pushed fixes for both gaps identified above:

- Multi-statement raise/fail branches: https://github.com/apiology/solargraph/commit/c41d1a33394469df6099f5f496f2d5559fc262ea moves `always_leaves_compound_statement?` into the shared `ParserGem::NodeMethods` module and recurses into a `:begin` clause's last child before the raise/fail shape check, so `if s.nil?; valid = ...; raise "..."; end` narrows the same as the single-statement case.
- `x || raise(...)` / `x ||= raise(...)`: https://github.com/apiology/solargraph/commit/58a780f22dcd4e5814174c46770b818f646820d6 fixes the root cause directly — `Chain::Or#resolve` was unioning the lhs type with the rhs's inferred type, and since `raise`/`fail` infer as `undefined`, `ComplexType` collapses any union containing an `undefined` item down to just `undefined` ([lib/solargraph/complex_type.rb:26](https://github.com/castwide/solargraph/blob/c666672b4e6ebc8ff7b160533674f3d81c418fb1/lib/solargraph/complex_type.rb#L26)), which is why the whole method's return type came back uninferrable. `Chain::Or` now takes a `rhs_never_returns:` flag (computed at chain-build time via the shared raise/fail-shape check) and, when set, returns just the lhs type minus nil instead of unioning in the never-returning rhs.

Both repro cases from the comments above now type-check clean, and the full suite is green (1633 examples, 0 failures, 65 pre-existing pending — matching this PR's original baseline).

lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
```diff
@@ -24,6 +24,10 @@ def process
               comments: comments,
               visibility: :public,
               gates: region.closure.gates.freeze,
+              # node: lets flow-sensitive typing find the end of this
+              # class/module body - see the same node: in
+              # NodeProcessor.process's root namespace pin.
```
apiology: Drop

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -107,9 +107,13 @@ def generate_links n
             # s(:or_asgn,
             #   s(:ivasgn, :@bar),
             #   s(:int, 123))
+            or_asgn_rhs_node = n.children[1] # s(:int, 123)
             lhs_chain = NodeChainer.chain n.children[0] # s(:ivasgn, :@bar)
-            rhs_chain = NodeChainer.chain n.children[1] # s(:int, 123)
-            or_link = Chain::Or.new([lhs_chain, rhs_chain])
+            # @sg-ignore Need to add nil check here
+            rhs_chain = NodeChainer.chain or_asgn_rhs_node
+            # @sg-ignore Need to add nil check here
```
apiology: Add nil checks, not tech debt

lib/solargraph/parser/parser_gem/node_methods.rb
```diff
@@ -181,6 +181,35 @@ def const_nodes_from node
           result
         end
 
+        # Determines whether a clause necessarily leaves its
+        # enclosing compound statement (method body, block, etc.) -
+        # e.g., via an explicit 'return', a loop-control keyword, or
+        # a raise/fail call - so code that depends on execution
+        # actually reaching the statements after it can treat those
+        # statements as unreachable from this clause.
```
apiology: You get 4 lines total between the method description and inline comments in method

lib/solargraph/parser/node_processor.rb
```diff
@@ -40,9 +40,16 @@ def deregister type, cls
       # @return [Array(Array<Pin::Base>, Array<Pin::LocalVariable>, Array<Pin::InstanceVariable>)]
       def self.process node, region = Region.new, pins = [], locals = [], ivars = []
         if pins.empty?
+          # node: is required so flow-sensitive typing can find the
+          # end of this compound statement (see
+          # FlowSensitiveTyping#process_if) - without it, an
+          # is_a?/nil? guard at the top level of a file never
+          # narrows, because enclosing_compound_statement_pin.node
+          # is always nil.
```
apiology: Drop changelogging comment

lib/solargraph/source/chain/or.rb
```diff
@@ -23,6 +37,14 @@ def resolve api_map, name_pin, locals
 
           [Solargraph::Pin::ProxyType.anonymous(combined_type, source: :chain)]
         end
+
+        protected
+
+        # @sg-ignore Fix "Not enough arguments to Module#protected"
```
apiology: Adhere to rules.rb standard

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -107,9 +107,14 @@ def generate_links n
             # s(:or_asgn,
             #   s(:ivasgn, :@bar),
             #   s(:int, 123))
+            or_asgn_rhs_node = n.children[1] # s(:int, 123)
+            raise "Or-assignment node missing right-hand side: #{n}" if or_asgn_rhs_node.nil?
+
             lhs_chain = NodeChainer.chain n.children[0] # s(:ivasgn, :@bar)
-            rhs_chain = NodeChainer.chain n.children[1] # s(:int, 123)
-            or_link = Chain::Or.new([lhs_chain, rhs_chain])
+            # @sg-ignore flow sensitive typing needs to handle 'raise if'
```
apiology: Replace with PR URL

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -118,8 +123,14 @@ def generate_links n
           elsif n.type == :and
             result.concat generate_links(n.children.last)
           elsif n.type == :or
-            result.push Chain::Or.new([NodeChainer.chain(n.children[0], @filename),
-                                       NodeChainer.chain(n.children[1], @filename, n)])
+            or_rhs_node = n.children[1]
+            # @sg-ignore Need to add nil check here
```
apiology: Don't add "Need to add nil check here" sg-ignores on new code - add the check or list the inference we're not doing yet that we should be, or the PR URL that implements it.

lib/solargraph/parser/node_processor.rb
```diff
@@ -40,9 +40,13 @@ def deregister type, cls
       # @return [Array(Array<Pin::Base>, Array<Pin::LocalVariable>, Array<Pin::InstanceVariable>)]
       def self.process node, region = Region.new, pins = [], locals = [], ivars = []
         if pins.empty?
+          # node: lets flow-sensitive typing find this statement's
+          # end (FlowSensitiveTyping#process_if); without it,
+          # top-level is_a?/nil? guards never narrow.
```
apiology: This comment belongs in the namespace initializer not here.

lib/solargraph/parser/parser_gem/node_chainer.rb
```diff
@@ -107,9 +107,14 @@ def generate_links n
             # s(:or_asgn,
             #   s(:ivasgn, :@bar),
             #   s(:int, 123))
+            or_asgn_rhs_node = n.children[1] # s(:int, 123)
+            raise "Or-assignment node missing right-hand side: #{n}" if or_asgn_rhs_node.nil?
```
apiology: Either this is possible or impossible.  If possible, what handles this raise?  If not possible, why not just call .fetch(1)?

lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
apiology: **Claude:**

Already dropped -- the current head passes `node: node,` with no comment above it.

lib/solargraph/parser/parser_gem/node_methods.rb

Cut to four in 05594a752: one description line plus three inline. Dropped
`(return/loop-control/raise/fail)`, which restated the keyword list two lines
below it.

lib/solargraph/source/chain/or.rb
apiology: **Claude:**

05594a752. Dropped `flow sensitive typing should be able to handle redefinition`
-- the typechecker already reported it unneeded. Replaced `return type could not
be inferred`, which is not a rules.rb string and only restated the error, with
the catalogued `literal arrays in this module turn into
::Solargraph::Source::Chain::Array`, which is the shared cause of both problems
on `equality_fields`. That string now appears on both lines of the method.

lib/solargraph/parser/node_processor.rb
apiology: **Claude:**

Gone from here in 05594a752, rewritten as a statement of what `node:` is for
with no reference to prior behaviour.

lib/solargraph/parser/node_processor.rb
apiology: **Claude:**

Moved in 05594a752 to `Pin::Namespace#initialize`. One thing found on the way:
`node:` is not declared there either -- it reaches `Namespace` via `**splat`
from `Pin::CompoundStatement#initialize`, which already documents it. The moved
comment cross-references that rather than repeating it.

lib/solargraph/parser/parser_gem/node_chainer.rb
apiology: **Claude:**

Impossible. The raise is gone and the line is `children.fetch(1)` as of
05594a752.

Checked by parsing 13 forms covering every left-hand kind (`lvasgn`, `ivasgn`,
`cvasgn`, `gvasgn`, `casgn`, index `send`, attr `send`) against every
interesting right-hand kind (`int`, `send`, `raise`, `return`, `nil`,
multi-statement `begin`, empty `kwbegin`). All 13 gave `:or_asgn` exactly two
real children -- `a ||= begin; end` yields `:kwbegin`, not `nil`. So nothing
would ever have handled that raise.

`.fetch(1)` does satisfy Solargraph here: the `Wrong argument type ... received
Parser::AST::Node, nil` on the line below is gone.

lib/solargraph/parser/parser_gem/node_chainer.rb
apiology: **Claude:**

No URL in the end. 05594a752 switched the line below to `children.fetch(1)`,
which resolves the argument-type error this marker was suppressing, so the
marker is deleted rather than relabelled. The gap stays tracked by its four
remaining occurrences elsewhere.

lib/solargraph/parser/parser_gem/node_chainer.rb
apiology: **Claude:**

All three gone in 05594a752. Two became real checks via `children.fetch`; the
third was already dead -- the typechecker reported it unneeded, because
`always_leaves_compound_statement?` declares
`@param clause_node [Parser::AST::Node, nil]`.

lib/solargraph/parser/parser_gem/node_chainer.rb
apiology: **Claude:**

Done in 05594a752. Both the `:or_asgn` and `:or` branches use `children.fetch`
now, and no `@sg-ignore` remains anywhere in the new code this PR adds to this
file.

PR #1260, opened 2026-08-04
Make RuboCop block CI again, and auto-prune .rubocop_todo.yml

apiology:
**Problem:** A new RuboCop offense could land on master without failing CI, because both steps of the `.rubocop_todo.yml` job were marked `continue-on-error`.

```yaml
    - name: Run RuboCop
      run: bundle exec rubocop -c .rubocop.yml
      # @todo Temporary, expect to revert in 0.60
      continue-on-error: true
```

That marker expected removal in 0.60; the repo is at 0.60.4. Separately, a todo entry whose offense someone has since fixed stays until a human regenerates the file, so the list only ever grows.

**Solution:** Drop both `continue-on-error` flags, and on pushes to master regenerate `.rubocop_todo.yml` and push it back when it shrinks. PRs keep the existing fail-with-a-diff behaviour, since `GITHUB_TOKEN` cannot push to the fork branches most PRs come from.

RuboCop also moves to 1.89.0: 1.80.0 ignored per-file `AllCops:Exclude`, so the deliberately-invalid encoding fixtures always failed `Lint/Syntax`, and RuboCop accepts no config for that cop at all.

**Test plan:**
- [ ] Auto-update step on a real push to master — the one path CI cannot exercise from a PR


.github/workflows/linting.yml
```diff
@@ -132,3 +141,26 @@ jobs:
            >&2 echo "Please address any new issues, then run '${cmd:?}' and push up any improvements"
            exit 1
         fi
+
+    # master only: the "Run RuboCop" step above already proved the tree is
+    # clean against the current .rubocop_todo.yml, so any diff here can only
+    # be a shrink (offenses that got fixed without their todo entry being
+    # removed) -- never a new offense. Commit and push that cleanup instead
+    # of asking someone to do it by hand.
```
apiology: Don't explain what is obvious (master only).  2 lines

PR #1262, opened 2026-08-05
Add specs for reverted fixes and uncovered branches

apiology:
**Problem:** Several fixes in this repo have been reverted or silently re-broken with the full suite green, because no spec asserted the fixed behavior — most recently an attempt to infer through `Class<generic<T>>#new`, which degraded arity checking for any `Class<X>` with a generic anywhere inside `X`:

```ruby
# k : Class<Array<generic<T>>>
k.new(1, 2, 3, 4)
# reported: Too many arguments to Class#initialize
# expected: Too many arguments to Array.new
```

That was caught by hand and reverted, and CI would not have caught it — nor the constant-alias resolution, the `cache_next_gemspec` re-entrancy guard, the class-wide `combined_pins_in_memory`, the `Chain::Call.new` signature external callers depend on, or the gemspec packaging a top-level `sig/`.

**Solution:** Adds eight examples asserting each of those behaviors directly, plus seven covering units and branches no spec reached (`YardMap::Cache#get_path_pins`, `DomainDirective.closure_at` and `#process_directive`, and `Chain::Call`'s non-macro directive fallthrough).

Spec-only; no `lib/` change.

**Test plan:** Full suite 1661 examples, 0 failures, 67 pending. Both generic-`Class` arity specs verified to fail when the reverted change is replayed on this base.

### DIFF
--- a/spec/gemspec_spec.rb
+++ b/spec/gemspec_spec.rb
@@ -1,13 +1,15 @@
 # frozen_string_literal: true
 
-describe 'solargraph.gemspec' do
-  let(:spec) { Gem::Specification.load(File.expand_path('../solargraph.gemspec', __dir__)) }
+describe Gem::Specification do
+  describe 'loaded from solargraph.gemspec' do
+    let(:spec) { described_class.load(File.expand_path('../solargraph.gemspec', __dir__)) }
 
-  it 'does not package a top-level sig/ directory that RBS auto-discovers in installed gems' do
-    # A prior sig/shims directory collided with consumers' own RBS
-    # collections (see https://github.com/castwide/solargraph/issues/1144).
-    # The shims now live under rbs/shims instead, which RBS does not scan
-    # automatically.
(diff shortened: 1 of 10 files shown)


spec/gemspec_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+describe Gem::Specification do
+  describe 'loaded from solargraph.gemspec' do
+    let(:spec) { described_class.load(File.expand_path('../solargraph.gemspec', __dir__)) }
+
+    it 'does not package a top-level sig/ directory that RBS auto-discovers in installed gems' do
+      # A prior sig/shims directory collided with consumers' own RBS
+      # collections (see https://github.com/castwide/solargraph/issues/1144).
+      # The shims now live under rbs/shims instead, which RBS does not scan
+      # automatically.
```
apiology: Drop the changelogging

PR #1263, opened 2026-08-05
Stop one unparseable type from aborting the whole typecheck run

apiology:
**Problem:** `solargraph typecheck` can die partway through with a backtrace, leaving the rest of the workspace unchecked.

```
$ solargraph typecheck --level strong
lib/a.rb:12: Unresolved call to bar
.../complex_type/unique_type.rb:54:in `parse': Bad hash type: name=Hash, substring=<String, nil, Enumerable<Integer>> - must have exactly two parameters (Solargraph::ComplexTypeError)
	from .../thor/command.rb:28:in `run'
	from .../thor/base.rb:585:in `start'
```

No summary line, no problem count, and every file after the failing one goes unvisited — one bad type annotation costs the entire run.

**Solution:** `TypeChecker#call_problems` now rescues each call site's inference and reports the failure as a `Problem` scoped to that call, and Hash types always render as `{K => V}`, which — unlike `<K, V>` — can represent ComplexTypes as keys and/or values.

https://claude.ai/code/session_01RALWyFBErKxFodRMRzLnBu


lib/solargraph/complex_type/type_methods.rb
```diff
@@ -190,7 +190,20 @@ def generate_substring_from &to_str
...
+          # that is itself a multi-item union) produces a string with too
+          # many top-level parameters to reparse. Fall back to the
+          # {K => V} notation, which can represent a comma-separated list
+          # on either side and still reparses correctly, whenever either
+          # side isn't exactly one single type.
+          if key_types.sum { |t| t.items.length } == 1 && subtypes.sum { |t| t.items.length } == 1
+            "<#{key_types_str}, #{subtypes_str}>"
+          else
+            "{#{key_types_str} => #{subtypes_str}}"
+          end
```
apiology: Why not always use => and save the reader a gigantic comment and complexity?

lib/solargraph/type_checker.rb
```diff
@@ -371,6 +371,15 @@ def call_problems
         end
         # @sg-ignore Need to add nil check here
         result.concat argument_problems_for(chain, api_map, closure_pin, locals, location)
+      rescue StandardError => e
+        # A single call site's type inference should never take down the
+        # rest of the typecheck run (or, when this is invoked per-file
+        # across a workspace, every file after it) -- degrade to one
+        # reported problem for this call site instead.
+        # @sg-ignore Need to add nil check here
```
apiology: 1 line

spec/complex_type_spec.rb
```diff
@@ -223,6 +223,50 @@
...
+    # A Hash{} type parsed with a comma-separated (unparenthesized union)
+    # key keeps key_types split across multiple ComplexType elements (see
+    # the 'parses multiple key/value types' example above). If that type's
+    # tag is later regenerated by code that no longer has hash_parameters?
+    # set (e.g. a UniqueType rebuilt with parameters_type: :list, as can
+    # happen when generics substitution recreates a Hash-named type from
+    # a context type's key_types/subtypes), the substring generator must
+    # not fall back to the 2-parameter Hash<K, V> notation, since that
+    # can't represent more than one key or value type and produces a
+    # string that fails to reparse.
```
apiology: Stop changelogging.  0-1 lines.

spec/complex_type_spec.rb
```diff
@@ -223,6 +223,50 @@
...
+        Solargraph::ComplexType.parse(mismatched.tag)
+      end.not_to raise_error
+    end
+
+    # A key_types/subtypes array can also have exactly one element that is
+    # itself a multi-item union ComplexType (as generics substitution can
+    # produce -- see UniqueType#resolve_generics, which replaces a single
+    # generic<Key> slot with ComplexType.new(context_type.key_types)).
+    # key_types.length == 1 alone isn't enough to guarantee the <K, V>
+    # notation is safe to use.
```
apiology: 0-1 lines, prefer 0

spec/type_checker_spec.rb
```diff
@@ -22,6 +22,31 @@
...
 
+  # Regression test for https://github.com/castwide/solargraph/pull/1259#issuecomment-5192216269
+  #
+  # call_problems iterates every call node in a file and infers each one
+  # with no rescue around it, so any exception raised deep inside a single
+  # call's type inference (a ComplexTypeError from a malformed type, or any
+  # other internal error) aborts TypeChecker#problems entirely instead of
+  # being reported as a problem for that one call site, silently losing
+  # every diagnostic for the rest of the file (and, when called per-file
+  # across a workspace, every file after it).
```
apiology: 0-1 lines.  Stop changelogging.

lib/solargraph/complex_type/type_methods.rb
```diff
@@ -190,7 +190,20 @@ def generate_substring_from &to_str
...
-          "<#{key_types_str}, #{subtypes_str}>"
+          # The <K, V> notation only has room for exactly one key type and
+          # one value type -- a single top-level comma splits it into K
+          # and V, so a second comma on either side (whether from more
+          # than one entry in key_types/subtypes, or from a single entry
+          # that is itself a multi-item union) produces a string with too
+          # many top-level parameters to reparse. Fall back to the
+          # {K => V} notation, which can represent a comma-separated list
+          # on either side and still reparses correctly, whenever either
+          # side isn't exactly one single type.
```
apiology: So why not prefer => notation?

spec/complex_type_spec.rb
```diff
@@ -223,6 +223,34 @@
...
+      original = Solargraph::ComplexType.parse('Hash{String, nil => Enumerable<Integer>}').items.first
+      expect(original.key_types.length).to eq(2)
+
+      mismatched = Solargraph::ComplexType::UniqueType.new(
+        'Hash', original.key_types, original.subtypes, rooted: original.rooted?, parameters_type: :list
+      )
+
+      expect do
+        Solargraph::ComplexType.parse(mismatched.tag)
+      end.not_to raise_error
```
apiology: Does this demonstrate a round-trip?

spec/complex_type_spec.rb
```diff
@@ -223,6 +223,34 @@
...
+      union_key = Solargraph::ComplexType.parse('String', 'nil')
+      value = Solargraph::ComplexType.parse('Enumerable<Integer>')
+
+      mismatched = Solargraph::ComplexType::UniqueType.new(
+        'Hash', [union_key], [value], rooted: false, parameters_type: :list
+      )
+
+      expect do
+        Solargraph::ComplexType.parse(mismatched.tag)
+      end.not_to raise_error
```
apiology: Does this demonstrate a round-trip?

lib/solargraph/complex_type/type_methods.rb
apiology: **Claude:** Done — the conditional and the comment are gone. `name == 'Hash'` now merges into the existing `hash_parameters?` branch so both render `{K => V}`, with `fixed_parameters?` moved ahead of it so `Hash(A, B)` still renders as a tuple. RuboCop's `Lint/DuplicateBranch` flagged the two identical bodies, which confirmed the merge was the right shape. `Hash<K, V>` remains valid input; only the generated tag changes, so two expectations that pinned the `<K, V>` rendering were updated.

lib/solargraph/complex_type/type_methods.rb
apiology: **Claude:** Now preferred unconditionally — see the reply on the `elsif` below. The `{K => V}` form is the only one that survives a union on either side, so there is no case left where `<K, V>` is worth choosing.

spec/complex_type_spec.rb
apiology: **Claude:** It did not — it asserted only that reparsing raised nothing. It now asserts the generated tag is `Hash{String, nil => Enumerable<Integer>}` and that reparsing that tag yields the same tag.

spec/complex_type_spec.rb
apiology: **Claude:** Same gap, same fix as the example above: the tag is asserted explicitly, then reparsed and compared against itself.

spec/complex_type_spec.rb
apiology: **Claude:** Trimmed in b4842103f, and the block is gone entirely as of b460ee249 — the example names carry it now.

spec/complex_type_spec.rb
apiology: **Claude:** Trimmed in b4842103f, and the block is gone entirely as of b460ee249 — the example names carry it now.

spec/type_checker_spec.rb
apiology: **Claude:** Trimmed in b4842103f, and the block is gone entirely as of b460ee249 — the example names carry it now.

PR #1266, opened 2026-08-06
Structurally verify RBS interface-typed expectations

### LINKED ISSUE #1232: Teach Solargraph to structurally verify RBS interface-typed parameters instead of blanket-allowing them

## Background

https://github.com/castwide/solargraph/pull/1228 fixed #1227 (`Hash#fetch` on `Hash{Symbol => Class<X>}` producing a spurious `generic<X>` typecheck error under rbs >= 4.1.0). The root cause was that `Hash#fetch`'s single-arg overload takes its key as the `::Hash::_Key` RBS interface instead of the generic `K`, and `Pin::Parameter#compatible_arg?` had no way to confirm a `Symbol` argument satisfies that interface without an explicit `include` declaration.

The fix passes `:allow_unmatched_interface` into the conformance check (`ComplexType::Conformance#ignore_interface?`), which treats *any* argument as compatible whenever an interface type is involved in the comparison.

## The gap

That's correct for `Hash::_Key` specifically — its required methods (`#hash`, `#eql?`) come from `Object`, so practically everything satisfies it. But `:allow_unmatched_interface` is a blanket suppression, not a real check. It now also silently accepts arguments for *any other* RBS interface (`_ToAry`, `_Each`, `_ToS`, third-party gem interfaces, etc.) even when the argument clearly doesn't implement it — e.g. passing an `Integer` where `_ToAry` is expected would now type-check without complaint, since `Integer` has no `#to_ary`.

## Proposed direction: real structural conformance

RBS `interface` declarations already become `Pin::Namespace` pins (`interface_decl_to_pin` in `lib/solargraph/rbs_map/conversions.rb`), with their required methods attached as regular method pins under that namespace. Confirmed this is queryable today:

```ruby
# The interface's own declared methods (filter out inherited Object noise via closure.path):
required = api_map.get_methods('Hash::_Key', scope: :instance)
                  .select { |p| p.closure && p.closure.path == 'Hash::_Key' }
required.map(&:name) # => ["eql?", "hash"]

# A candidate type either has each required method or it doesn't:
required.all? { |p| !api_map.get_method_stack('Symbol', p.name, scope: :instance).empty? } # => true
!api_map.get_method_stack('Integer', 'to_ary', scope: :instance).empty? # => false
```

So instead of (or in addition to, as a fallback) `:allow_unmatched_interface`, `ComplexType::Conformance` (or a new method on `ComplexType::UniqueType`) could do real duck-type verification: for an interface-typed `expected`, look up its directly-declared method pins and check that `inferred`'s method stack has a same-named method for each. If all required methods are present, treat it as conforming; if not, it's a genuine mismatch, not a "can't tell" case.

## Known complications for whoever picks this up

- **Generic interfaces** (e.g. `_Each[Elem]`, `_Pair[K, V]`) — presence-only checking of method names ignores the interface's own type parameters entirely; a v1 could ignore this (accept structurally by name only) and note it as a known limitation, or attempt matching each required method's parameter/return types too (more accurate, more complex, and recursive — checking a method's signature may itself involve interface types).
- **Performance** — this check would run on every method-call argument match against an interface-typed parameter; likely wants memoization keyed by `(inferred_type_name, interface_name)` since interface satisfaction is structural, not context-dependent, for most cases.
- **Where to hook it in** — likely `ComplexType::Conformance#ignore_interface?` (currently a blunt `rules.include?(:allow_unmatched_interface)` check) or a new method called before falling back to that rule, so `:allow_unmatched_interface` remains available as an explicit "give up and allow" escape hatch for cases where the structural check can't be done (e.g. the interface pin isn't found, or it's satisfied via `method_missing`/metaprogramming Solargraph can't see).
- **Existing `CoreFills::INCLUDES` hack** (`_ToAry`, `_Each`, `_ToS` nominally included into specific classes) could potentially be retired in favor of genuine structural checking, though that's a separate cleanup and not required to land this.
- **Test coverage** — should include both a positive case (interface-satisfying argument type-checks) and a negative case (an argument that clearly does *not* implement the interface should now be flagged, where before/currently it silently passes under `:allow_unmatched_interface`).

apiology:

## Problem

RBS 4.1 broke hash lookups in Solargraph.

```ruby
# @param registry [Hash{Symbol => Array<String>}]
# @return [Integer]
def count(registry)
  # #count return type could not be inferred
  registry.fetch(:x).length
end
```

4.1 loosened `Hash`'s key lookups from the class parameter `K` to the structural interface `Hash::_Key`. Solargraph has rough workarounds for a few existing interfaces and no general way to satisfy one, so `Symbol` fails to match the parameter, every `#[]` and `#fetch` overload is rejected as a non-match, and the caller is left with no usable type.

```rbs
class Hash[unchecked out K, unchecked out V]
  # rbs 4.0.3
  def fetch: (K arg0) -> V

  # rbs 4.1.3 — no K anywhere in the signature
  def fetch: (_Key key) -> V

  interface _Key
    def hash: () -> Integer
    def eql?: (untyped rhs) -> boolish
  end
end
```

## Solution

Match interfaces structurally: an inferred type conforms to an interface-typed expectation if its own method stack has every method the interface declares. `Symbol` has `#hash` and `#eql?`, so it satisfies `Hash::_Key` without anyone declaring an `include`.

`:allow_unmatched_interface` now applies only where the expected interface has no methods Solargraph can see, leaving nothing to check.

Limitation: generic interfaces (`_Each[Elem]`) are checked by method name only, not parameter or return types, since they get turned into YARD `#duck_types`.

Fixes https://github.com/castwide/solargraph/issues/1232

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -447,9 +447,9 @@ module Solargraph
     #
     # @param rooted_tag [String] The fully qualified namespace/interface to search for methods
     # @param scope [Symbol] :class or :instance
-    # @return [Array<Solargraph::Pin::Method>]
+    # @return [Enumerable<Solargraph::Pin::Method>]
     def get_own_methods rooted_tag, scope: :instance
-      get_methods(rooted_tag, scope: scope).select { |pin| pin.closure&.path == rooted_tag }
+      store.get_methods(ComplexType.try_parse(rooted_tag).namespace, scope: scope)
     end
 
     # Get an array of methods available in a particular context.
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -442,12 +442,9 @@ module Solargraph
       store.pins_by_class(Pin::Block)
     end
 
-    # Get the methods a namespace (e.g. an RBS interface) declares
-    # directly on itself, excluding ones inherited from Object,
-    # superclasses, or mixins. Useful for structural (duck-type)
-    # checks against an interface's own contract, e.g. whether some
-    # other type implements every method `Hash::_Key` itself declares
-    # (`#hash`, `#eql?`), independent of the interface's name.
+    # Methods a namespace declares directly on itself, excluding ones
+    # inherited from Object, superclasses, or mixins - e.g. the two
+    # methods `Hash::_Key` itself requires (`#hash`, `#eql?`).
     #
     # @param rooted_tag [String] The fully qualified namespace/interface to search for methods
     # @param scope [Symbol] :class or :instance
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -442,6 +442,20 @@ module Solargraph
       store.pins_by_class(Pin::Block)
     end
 
+    # Get the methods a namespace (e.g. an RBS interface) declares
+    # directly on itself, excluding ones inherited from Object,
+    # superclasses, or mixins. Useful for structural (duck-type)
+    # checks against an interface's own contract, e.g. whether some
+    # other type implements every method `Hash::_Key` itself declares
+    # (`#hash`, `#eql?`), independent of the interface's name.
+    #
+    # @param rooted_tag [String] The fully qualified namespace/interface to search for methods
+    # @param scope [Symbol] :class or :instance
+    # @return [Array<Solargraph::Pin::Method>]
+    def get_own_methods rooted_tag, scope: :instance
+      get_methods(rooted_tag, scope: scope).select { |pin| pin.closure&.path == rooted_tag }
+    end
+
     # Get an array of methods available in a particular context.
     #
     # @param rooted_tag [String] The fully qualified namespace to search for methods
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -443,8 +443,7 @@ module Solargraph
     end
 
     # Methods a namespace declares directly on itself, excluding ones
-    # inherited from Object, superclasses, or mixins - e.g. the two
-    # methods `Hash::_Key` itself requires (`#hash`, `#eql?`).
+    # inherited from Object, superclasses, or mixins.
     #
     # @param rooted_tag [String] The fully qualified namespace/interface to search for methods
     # @param scope [Symbol] :class or :instance
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -442,16 +442,6 @@ module Solargraph
       store.pins_by_class(Pin::Block)
     end
 
-    # Methods a namespace declares directly on itself, excluding ones
-    # inherited from Object, superclasses, or mixins.
-    #
-    # @param rooted_tag [String] The fully qualified namespace/interface to search for methods
-    # @param scope [Symbol] :class or :instance
-    # @return [Enumerable<Solargraph::Pin::Method>]
-    def get_own_methods rooted_tag, scope: :instance
-      store.get_methods(ComplexType.try_parse(rooted_tag).namespace, scope: scope)
-    end
-
     # Get an array of methods available in a particular context.
     #
     # @param rooted_tag [String] The fully qualified namespace to search for methods
--- a/lib/solargraph/complex_type/conformance.rb
+++ b/lib/solargraph/complex_type/conformance.rb
@@ -138,9 +138,6 @@ module Solargraph
         true
       end
 
-      # The methods `expected` declares directly on itself, excluding ones
-      # inherited from `Object` and other ancestors.
-      #
       # @return [Array<Pin::Method>]
       def required_interface_methods
         api_map.get_own_methods(expected.name)
--- a/lib/solargraph/complex_type/conformance.rb
+++ b/lib/solargraph/complex_type/conformance.rb
@@ -41,7 +41,7 @@ module Solargraph
           # :nocov:
         end
 
-        return true if ignore_interface?
+        return true if ignore_unmatchable_interface?
         return true if conforms_via_reverse_match?
 
         downcast_inferred = inferred.downcast_to_literal_if_possible
@@ -86,9 +86,14 @@ module Solargraph
         with_new_types(inferred, expected.erase_parameters).conforms_to_unique_type?
       end
 
-      def ignore_interface?
-        (expected.any?(&:interface?) && rules.include?(:allow_unmatched_interface)) ||
-          (inferred.interface? && rules.include?(:allow_unmatched_interface))
+      # An interface-typed expectation is verified structurally in
+      # #erased_type_conforms?. There's no equivalent structural check when
+      # the *inferred* type is itself an abstract interface (e.g., a method
+      # returns `_ToAry`), since Solargraph doesn't know which concrete type
+      # will show up at runtime, so :allow_unmatched_interface remains a
+      # blanket escape hatch for that direction only.
+      def ignore_unmatchable_interface?
+        inferred.interface? && rules.include?(:allow_unmatched_interface)
       end
 
       def can_strip_expected_parameters?
@@ -104,6 +109,8 @@ module Solargraph
       end
 
       def erased_type_conforms?
+        return true if expected.interface? && interface_conforms?
+
         case variance
         when :invariant
           return false unless inferred.name == expected.name
@@ -125,6 +132,43 @@ module Solargraph
           # :nocov:
         end
         true
+      end
+
+      # Whether `inferred` satisfies the `expected` RBS interface (e.g.
+      # `Hash::_Key`, `_ToAry`). Prefers real duck-type verification —
+      # `inferred` conforms if its method stack has every method the
+      # interface itself declares (methods it inherits from `Object` are
+      # ignored, since practically everything provides those) — and only
+      # falls back to the blanket :allow_unmatched_interface rule when no
+      # verdict could be reached, e.g. the interface pin isn't in the
+      # ApiMap.
+      #
+      # @return [Boolean]
+      def interface_conforms?
+        verdict = structural_interface_verdict
+        return verdict unless verdict.nil?
+
+        rules.include?(:allow_unmatched_interface)
+      end
+
+      # The methods `expected` declares directly on itself, excluding ones
+      # inherited from `Object` and other ancestors.
+      #
+      # @return [Array<Pin::Method>]
+      def required_interface_methods
+        api_map.get_methods(expected.name, scope: :instance)
+               .select { |pin| pin.closure&.path == expected.name }
+      end
+
+      # @return [Boolean, nil] true or false if `expected`'s directly
+      #   declared methods could be checked against `inferred`'s method
+      #   stack, or nil if no verdict could be reached (e.g., the interface
+      #   has no pin, or declares no methods of its own)
+      def structural_interface_verdict
+        required = required_interface_methods
+        return nil if required.empty?
+
+        required.all? { |pin| !api_map.get_method_stack(inferred.name, pin.name, scope: :instance).empty? }
       end
 
       def key_types_conform?
--- a/lib/solargraph/complex_type/conformance.rb
+++ b/lib/solargraph/complex_type/conformance.rb
@@ -41,8 +41,8 @@ module Solargraph
           # :nocov:
         end
 
-        conformance = interface_conformance
-        return conformance unless conformance.nil?
+        return conforms_via_interface? if interface_declares_methods?
+        return true if expected.interface? && rules.include?(:allow_unmatched_interface)
 
         return true if conforms_via_reverse_match?
 
@@ -88,18 +88,10 @@ module Solargraph
         with_new_types(inferred, expected.erase_parameters).conforms_to_unique_type?
       end
 
-      # Settles conformance when `expected` is an RBS interface:
+      # Whether `expected` is an RBS interface declaring methods of its own:
       # https://github.com/ruby/rbs/blob/master/docs/syntax.md#interface-declaration
-      # Does not verify interface type params: https://github.com/castwide/solargraph/issues/1267
-      # @return [Boolean, nil] nil when no interface, or nothing here decides
-      def interface_conformance
-        return nil unless expected.interface?
-
-        structural = structural_interface_conformance
-        return structural unless structural.nil?
-        return true if rules.include?(:allow_unmatched_interface)
-
-        nil
+      def interface_declares_methods?
+        expected.interface? && !required_interface_methods.empty?
       end
 
       def can_strip_expected_parameters?
@@ -140,16 +132,13 @@ module Solargraph
 
       # @return [Array<Pin::Method>]
       def required_interface_methods
-        api_map.get_own_methods(expected.name)
+        @required_interface_methods ||= api_map.get_own_methods(expected.name)
       end
 
-      # @return [Boolean, nil] whether `inferred` implements every method the
-      #   interface declares, or nil if it has no pin or declares none
-      def structural_interface_conformance
-        required = required_interface_methods
-        return nil if required.empty?
-
-        required.all? { |pin| !api_map.get_method_stack(inferred.name, pin.name, scope: :instance).empty? }
+      # Whether `inferred` implements every method the interface declares; their
+      # signatures go unchecked: https://github.com/castwide/solargraph/issues/1267
+      def conforms_via_interface?
+        required_interface_methods.all? { |pin| !api_map.get_method_stack(inferred.name, pin.name, scope: :instance).empty? }
       end
 
       def key_types_conform?
--- a/lib/solargraph/complex_type/conformance.rb
+++ b/lib/solargraph/complex_type/conformance.rb
@@ -88,25 +88,14 @@ module Solargraph
         with_new_types(inferred, expected.erase_parameters).conforms_to_unique_type?
       end
 
-      # Resolves interface-typed conformance before any parameter/subtype
-      # comparisons run, the same way the old blanket
-      # :allow_unmatched_interface bypass did. That's deliberate: RBS
-      # interfaces can have their own type parameters (e.g. `_Each[Elem]`),
-      # and this doesn't verify those (see
-      # https://github.com/castwide/solargraph/issues/1267), so once the
-      # interface question is settled, comparing `expected`'s subtypes
-      # against `inferred`'s would either be meaningless or wrong.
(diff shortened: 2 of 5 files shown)


lib/solargraph/complex_type/conformance.rb
```diff
@@ -86,9 +88,34 @@ def conforms_via_stripped_expected_parameters?
...
+      # There's no structural check when the *inferred* type is itself an
+      # abstract interface (e.g., a method returns `_ToAry`), since
+      # Solargraph doesn't know which concrete type will show up at
+      # runtime, so :allow_unmatched_interface remains a blanket escape
+      # hatch for that direction.
+      #
+      # @return [Boolean, nil] true/false if the interface question
+      #   settles conformance outright, nil if there's no interface
+      #   involved (or no verdict could be reached and no fallback rule
+      #   applies), meaning normal conformance checking should proceed
```
apiology: Why do you need 4 more lines after writing a novel above?  You get 4 lines total

lib/solargraph/complex_type/conformance.rb
```diff
@@ -127,6 +154,25 @@ def erased_type_conforms?
...
+      #
+      # @return [Array<Pin::Method>]
+      def required_interface_methods
+        api_map.get_own_methods(expected.name)
+      end
+
+      # @return [Boolean, nil] true or false if `expected`'s directly
+      #   declared methods could be checked against `inferred`'s method
+      #   stack, or nil if no verdict could be reached (e.g., the interface
+      #   has no pin, or declares no methods of its own)
```
apiology: 2 lines

lib/solargraph/api_map.rb
```diff
@@ -442,6 +442,20 @@ def get_block_pins
       store.pins_by_class(Pin::Block)
     end
 
+    # Get the methods a namespace (e.g. an RBS interface) declares
+    # directly on itself, excluding ones inherited from Object,
+    # superclasses, or mixins. Useful for structural (duck-type)
+    # checks against an interface's own contract, e.g. whether some
+    # other type implements every method `Hash::_Key` itself declares
+    # (`#hash`, `#eql?`), independent of the interface's name.
```
apiology: 2 lines only

spec/complex_type/conforms_to_spec.rb
```diff
@@ -81,7 +81,13 @@ class Sub < Sup; end
   end
 
   it 'handles singleton types compared against their literals' do
-    pending 'side of effect of inference changes'
+    # https://github.com/castwide/solargraph/issues/1196
+    #
+    # `nil` doesn't yet simplify to `NilClass` the way other literals
+    # simplify to their class name. Fixed by
+    # https://github.com/castwide/solargraph/pull/1223 (stacked:
+    # https://github.com/apiology/solargraph/pull/40).
```
apiology: No changelogging.  Put issue/PR URL in pending; add two pending lines if it depends on two

spec/complex_type/conforms_to_spec.rb
```diff
@@ -240,6 +246,88 @@ class Sub < Sup; end
...
+      inf = described_class.parse('Integer')
+      expect(inf.conforms_to?(api_map, exp, :method_call, [:allow_unmatched_interface])).to be(true)
+      expect(inf.conforms_to?(api_map, exp, :method_call)).to be(false)
+    end
+
+    it 'rejects a same-named method with the wrong return type' do
+      # https://github.com/castwide/solargraph/issues/1267
+      #
+      # The structural check only confirms a `to_ary` method exists; it
+      # doesn't verify it actually returns an Array.
```
apiology: Drop; put URL alone in pending

spec/complex_type/conforms_to_spec.rb
```diff
@@ -240,6 +246,88 @@ class Sub < Sup; end
...
+      inf = described_class.parse('BadToAry')
+      match = inf.conforms_to?(api_map, exp, :method_call)
+      expect(match).to be(false)
+    end
+
+    it 'rejects a same-named method with the wrong arity' do
+      # https://github.com/castwide/solargraph/issues/1267
+      #
+      # The structural check only confirms an `eql?` method exists; it
+      # doesn't verify it accepts the argument Hash::_Key#eql? requires.
```
apiology: Drop; put URL alone in pending

lib/solargraph/complex_type/conformance.rb
```diff
@@ -41,7 +41,9 @@ def conforms_to_unique_type?
           # :nocov:
         end
 
-        return true if ignore_interface?
+        interface_verdict = interface_bypass_verdict
+        return interface_verdict unless interface_verdict.nil?
```
apiology: 'interface verdict' is a vague term not explained anywhere.  try again at naming?

Also throw in a URL that explains RBS interfaces please

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Cut to 4 lines across 5078d153b and bdc250604. The rename in 549de1d39 pushed it back to 7, and 4fbbf11df cuts it to 4 again: one description line, the RBS URL, the 1267 URL, and the @return.

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Cut to a 2-line @return in 5078d153b, and still 2 after the rename to structural_interface_conformance in 549de1d39. Also trimmed the neighbouring required_interface_methods docstring to a bare @return in 0864eb510.

lib/solargraph/api_map.rb
apiology: **Claude:** Cut from 6 description lines to 3 in 5078d153b, then to 2 in bdc250604. The @param and @return tags below it are unchanged.

spec/complex_type/conforms_to_spec.rb
apiology: **Claude:** Done in bdc250604. The comment block is gone and there are two pending lines, one per dependency: issues/1196 and pull/1223.

spec/complex_type/conforms_to_spec.rb
apiology: **Claude:** Done in 5078d153b. Now pending https://github.com/castwide/solargraph/issues/1267 with no comment above it.

spec/complex_type/conforms_to_spec.rb
apiology: **Claude:** Done in 5078d153b, same as the wrong-return-type example above.

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Renamed in 549de1d39. verdict was a coined term: the two methods it named held the only two [Boolean, nil] returns in lib/, so there was nothing for a reader to match it against. They are now interface_conformance and structural_interface_conformance, matching conforms_to? and erased_type_conforms? already in this class, and what nil means moved into the @return tag. RBS URL added: https://github.com/ruby/rbs/blob/master/docs/syntax.md#interface-declaration - its _Hashing example declares the same hash and eql? pair that Hash::_Key requires in the specs.

lib/solargraph/complex_type/conformance.rb
```diff
@@ -86,9 +88,18 @@ def conforms_via_stripped_expected_parameters?
         with_new_types(inferred, expected.erase_parameters).conforms_to_unique_type?
       end
 
-      def ignore_interface?
-        (expected.any?(&:interface?) && rules.include?(:allow_unmatched_interface)) ||
-          (inferred.interface? && rules.include?(:allow_unmatched_interface))
+      # Settles conformance when `expected` is an RBS interface:
```
apiology: "Settles conformance"?  Not clear what that means.  Try again on wording.

Weird to have "interface conformance" be a boolean.  What would make this more clear?  conforms_via_interface?  

Something else?  Whatever it is, this is not it.

lib/solargraph/complex_type/conformance.rb
```diff
@@ -127,6 +138,20 @@ def erased_type_conforms?
...
       end
 
+      # @return [Array<Pin::Method>]
+      def required_interface_methods
+        api_map.get_own_methods(expected.name)
+      end
+
+      # @return [Boolean, nil] whether `inferred` implements every method the
+      #   interface declares, or nil if it has no pin or declares none
+      def structural_interface_conformance
```
apiology: Ditto above - and the method name gives no concept of what is different between this and the other one

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Took your name and dropped the tri-state, in da8b11323. There are now two plain predicates and no Boolean-or-nil left in the class: interface_declares_methods? asks whether `expected` is an interface with a contract to check, and conforms_via_interface? checks it, matching conforms_via_reverse_match? right below it. The allow_unmatched_interface fallback that nil used to encode is now its own line in conforms_to_unique_type?, so the caller shows the whole policy in two lines instead of hiding it behind a nil.

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Same commit, da8b11323. The difference is now carried by the names: interface_declares_methods? asks whether there is anything to check, conforms_via_interface? does the checking. required_interface_methods memoizes, since both predicates call it.

lib/solargraph/api_map.rb
```diff
@@ -442,6 +442,16 @@ def get_block_pins
...
 
+    # Methods a namespace declares directly on itself, excluding ones
+    # inherited from Object, superclasses, or mixins.
+    #
+    # @param rooted_tag [String] The fully qualified namespace/interface to search for methods
+    # @param scope [Symbol] :class or :instance
+    # @return [Array<Solargraph::Pin::Method>]
+    def get_own_methods rooted_tag, scope: :instance
+      get_methods(rooted_tag, scope: scope).select { |pin| pin.closure&.path == rooted_tag }
+    end
```
apiology: This seems strange to me.  Is this the way to do it because we create namespace pins for interfaces?

spec/complex_type/conforms_to_spec.rb
```diff
@@ -81,7 +81,8 @@ class Sub < Sup; end
   end
 
   it 'handles singleton types compared against their literals' do
-    pending 'side of effect of inference changes'
+    pending 'https://github.com/castwide/solargraph/issues/1196'
```
apiology: We have a PR for this now and it's 1223 - drop the above issue

lib/solargraph/api_map.rb
apiology: **Claude:** No, namespace pins for interfaces were not the reason - it was just wasteful, and 072538d7a drops it. get_methods does the full deep walk (inherited methods, mixins, conventions) and the select then threw almost all of it away. Store#get_methods already returns just the method pins a namespace declares itself, which is exactly what is wanted, so it now delegates there. I checked the two agree before changing it: Hash::_Key gives [eql?, hash], _ToAry gives [to_ary], _Each gives [each], and Symbol gives the same 29 names either way. Return type follows Store#get_methods to Enumerable, so the caller asks any? instead of empty?.

spec/complex_type/conforms_to_spec.rb
apiology: **Claude:** Dropped in f0e6d2a90; the 1223 pending is the only one left.

lib/solargraph/api_map.rb
apiology: Let me rephrase.

This seems strange to me.  Why does this method exist?  Why would we care whether a class implements a method or its superclass?

spec/complex_type/conforms_to_spec.rb
```diff
@@ -240,6 +240,80 @@ class Sub < Sup; end
     end
   end
 
+  context 'with RBS interface types' do
+    it 'structurally validates a type that satisfies the interface, without any rule' do
+      exp = described_class.parse('Hash::_Key')
+      inf = described_class.parse('Symbol')
+      match = inf.conforms_to?(api_map, exp, :method_call)
+      expect(match).to be(true)
+    end
```
apiology: There are versions of RBS where Hash::_Key is not defined - and I'm sure the same for _ToAry.   

Can we instead create a context where we define our own RBS interfaces?

lib/solargraph/complex_type/conformance.rb
```diff
@@ -41,7 +41,9 @@ def conforms_to_unique_type?
           # :nocov:
         end
 
-        return true if ignore_interface?
+        return conforms_via_interface? if interface_declares_methods?
```
apiology: We obscure more than we enlighten by creating interface_declares_methods - let's re-inline it.

lib/solargraph/api_map.rb
apiology: **Claude:** It should not exist, and bf94c112a deletes it. get_methods already takes deep: false, which returns the identical pins in every case I checked (Hash::_Key, _ToAry, _Each, Symbol), so this was a second public name for something that was already there. Conformance now calls get_methods(expected.name, scope: :instance, deep: false) directly.\n\nOn why own methods at all: the two sides are deliberately asymmetric. The interface side needs only what the interface itself declares, because that is the contract - a deep lookup would drag in every Object method Solargraph attributes to the namespace and demand those too. The inferred side stays deep, via get_method_stack, so a type may satisfy the contract by inheritance.

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Gone in bf94c112a; the interface? test and the allow_unmatched_interface fallback are back inline in conforms_to_unique_type?.

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Re-inlined in bf94c112a. conforms_to_unique_type? now reads:\n\n```ruby\nif expected.interface?\n  return conforms_via_interface? if required_interface_methods.any?\n  return true if rules.include?(:allow_unmatched_interface)\nend\n```

spec/complex_type/conforms_to_spec.rb
apiology: **Claude:** Done in a6ea6710a. The context now writes its own RBS to a tmpdir and indexes it, following the pattern in spec/rbs_map/conversions_spec.rb, so nothing depends on which interfaces a given RBS version ships. It declares _MyKey and _MyToAry plus MyKeyImpl, BadToAry and BadKey, and the examples use those instead of Hash::_Key and _ToAry. The core-fill example still checks Array against _MyToAry, since Array#to_ary comes from core and index keeps the core pins.

lib/solargraph/complex_type/conformance.rb
apiology: **Claude:** Correction - my previous reply on this thread belongs on the newer interface_declares_methods thread, not here. This thread was already settled by 549de1d39. Sorry for the noise.

lib/solargraph/complex_type/conformance.rb
```diff
@@ -41,7 +41,11 @@ def conforms_to_unique_type?
           # :nocov:
         end
 
-        return true if ignore_interface?
+        if expected.interface?
```
apiology: Do we have tests where the type to be conformed on is a union, one of whom is an interface?

How does ComplexType#interface? work?

Show me how RBS interfaces are processed when they are read in?

apiology: **Claude:** Taking the three in order.\n\n**Union tests:** there were none, and 92d3dd374 adds three - a union whose interface member is satisfied, the same with the interface second, and one where no member is satisfied. They pass.\n\n**Why they pass:** Conformance never sees a union. UniqueType#conforms_to? decomposes the expected side first, so each member arrives separately and the interface branch fires only for the interface member:\n\n```ruby\nexpected.any? do |expected_unique_type|\n  Conformance.new(api_map, self, expected_unique_type, situation, rules,\n                  variance: variance).conforms_to_unique_type?\nend\n```\n\n**interface?** is purely a name test, defined on UniqueType:\n\n```ruby\ndef interface?\n  name.start_with?('_') || name.include?('::_')\nend\n```\n\nTwo things about it worth knowing. It is defined twice, byte-identically, in TypeMethods and again in UniqueType, which includes TypeMethods - so the UniqueType copy shadows the mixin. The TypeMethods copy is not dead, though: ComplexType has no interface? of its own and reaches it through method_missing, which is gated on TypeMethods.public_method_defined?. And that delegation forwards to @items.first only, so ComplexType#interface? on a union answers for the first member alone - true for `_MyToAry, Integer` and false for `Integer, _MyToAry`. Nothing in the conformance path calls it, per the decomposition above, but it is a sharp edge.\n\n**Reading RBS in:** Conversions#interface_decl_to_pin turns `interface _Foo` into an ordinary Pin::Namespace with type: :module, closure ROOT_PIN, visibility: :hidden (a stated hack to keep interfaces out of autocompletion), and an @abstract (RBS interface) docstring tag. Its members then go through the same convert_member_to_pin path as any class, so they become plain method pins. That is why interface? can be a name test and why get_methods(name, deep: false) returns exactly the declared contract.

apiology: Is the '_' prefix anything other than a convention easily broken by RBS authors?

apiology: **Claude:** Enforced by the grammar, not a convention. The lexer gives underscore-prefixed names their own token class, and each declaration rule accepts only one of them, so the three name spaces cannot overlap. Ran RBS::Parser.parse_signature on each:\n\n```\ninterface _Foo   ACCEPTED\ninterface Foo    REJECTED  expected one of interface name, token=`Foo` (tUIDENT)\nclass _Foo       REJECTED  expected one of class/module/constant name, token=`_Foo` (tULIDENT)\ntype _foo = ...  REJECTED  expected one of alias name, token=`_foo` (tULLIDENT)\ntype foo = ...   ACCEPTED\n```\n\nAn RBS author cannot name an interface without the underscore, and cannot give a class, module or type alias a name that starts with one. So for any name that came in through RBS, the check is exact rather than heuristic.\n\nThe boundary is that interface? tests a Solargraph type-name string, not an RBS AST node, so it is only as strong as the source of the name. A YARD tag saying `@return [_Foo]` would be treated as an interface with nothing enforcing it. That degrades safely: with no interface pin behind the name, the declared-methods list is empty, so the branch falls through to allow_unmatched_interface or to the normal path, exactly as the _NoSuchInterface spec covers.\n\nNo code change from this one - flagging it as a question answered rather than a fix.

PR #1269, opened 2026-08-06
Fix false positive: Struct.new(keyword_init: true) members are optional

### LINKED ISSUE #1268: Struct.new(keyword_init: true) members treated as required, not optional/nil-defaulting
**Claude:** filed on behalf of the repo owner while auditing `@sg-ignore` suppressions in a downstream project.

## Summary

`solargraph typecheck --level strong` treats every member of a
`Struct.new(..., keyword_init: true)`-generated class as a *required*
keyword argument, and reports a false "Missing keyword argument" error
when a call site omits one. At runtime, Ruby's `Struct` with
`keyword_init: true` makes every member optional - an omitted keyword
simply defaults to `nil`, exactly like an ordinary `Struct.new(...)`
without `keyword_init`.

## Reproduction

```ruby
# typed: true
# frozen_string_literal: true

# @sg-ignore dynamic-metaprogramming
class Watch < Struct.new(:name, :time, keyword_init: true); end

class Caller
  # @return [Watch]
  def go
    Watch.new(name: 'foo')
  end
end
```

```
repro.rb:10: Missing keyword argument time to Watch.new
```

Confirmed this is a false positive - the identical call succeeds at
runtime with `time` defaulting to `nil`:

```
$ ruby -e '
Watch = Struct.new(:name, :time, keyword_init: true)
w = Watch.new(name: "foo")
p w
p w.time
'
#<struct Watch name="foo", time=nil>
nil
```

## Environment

- solargraph 0.60.3

## Notes

I searched existing issues/PRs first:

- #660 looked related by error message ("Missing keyword argument") but
  is a different bug - that repro supplies *both* keywords and Solargraph
  still misreports one as missing (an argument-accounting issue). Here
  the keyword genuinely is omitted at the call site, and the omission is
  valid Ruby.
- #939, #992, #962, #1031 (merged) add/improve general `Struct` support
  (attribute typing, inline definitions, a crash fix) but none address
  keyword-member optionality under `keyword_init: true`.

This affects real code: we're working around it today with 6
`@sg-ignore`s across one file, each suppressing a call site that
intentionally omits an optional Struct member.

apiology:
## Summary

`solargraph typecheck --level strong` treated every member of a `Struct.new(..., keyword_init: true)`-generated class as a *required* keyword argument, reporting a false "Missing keyword argument" error when a call site omitted one. At runtime, `keyword_init: true` Struct members default to `nil` when omitted, exactly like ordinary `Struct.new(...)` members.

`lib/solargraph/convention/struct_definition.rb` generated `Pin::Parameter` pins for keyword-init Struct members with `decl: :kwarg` (required). Changed to `decl: :kwoptarg` (optional), matching how `TypeChecker#parameterized_arity_problems_for` already distinguishes required (`:kwarg`) from optional (`:kwoptarg`) keyword parameters.

`Data.define` (`lib/solargraph/convention/data_definition.rb`) is untouched ��� its members genuinely are required at runtime, so `:kwarg` stays correct there.

```ruby
class Watch < Struct.new(:name, :time, keyword_init: true); end

class Caller
  # @return [Watch]
  def go
    Watch.new(name: 'foo') # no longer flagged as missing `time:`
  end
end
```

Fixes #1268

## Test plan
- [x] Added spec asserting keyword-init Struct parameters have `decl == :kwoptarg`
- [x] Added typecheck spec reproducing the issue's exact repro, asserting no "Missing keyword argument" problem
- [x] `bundle exec rspec` (full suite): 1626 examples, 0 failures
- [x] `bundle exec rubocop` on changed files: no offenses
- [x] Manually ran `solargraph typecheck --level strong` against the issue's repro: 0 problems found (previously reported the false positive)

���� Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TqwAY5yf6K1ZQdYoSZDkmy

### DIFF
--- a/spec/convention/struct_definition_spec.rb
+++ b/spec/convention/struct_definition_spec.rb
@@ -23,6 +23,19 @@ describe Solargraph::Convention::StructDefinition do
       expect(param_baz.return_type.tag).to eql('Integer')
     end
 
+    it 'treats keyword args as optional, since Ruby defaults omitted members to nil' do
+      source = Solargraph::SourceMap.load_string(%(
+        # @param bar [String]
+        # @param baz [Integer]
+        Foo = Struct.new(:bar, :baz, keyword_init: true)
+      ), 'test.rb')
+
+      # @type [Array<Solargraph::Pin::Parameter>]
+      params = source.pins.find { |p| p.path == 'Foo#initialize' }.parameters
+
+      expect(params.map(&:decl)).to eql(%i[kwoptarg kwoptarg])
+    end
+
     it 'sets closure to method on assignment operator parameters' do
       source = Solargraph::SourceMap.load_string(%(
         # @param bar [String]
@@ -147,6 +160,20 @@ describe Solargraph::Convention::StructDefinition do
         Foo = Struct.new(:bar, :baz)
       ))
       expect { checker.problems }.not_to raise_error
+    end
+
+    it 'does not report a missing keyword argument when a keyword_init member is omitted' do
+      checker = type_checker(%(
+        class Watch < Struct.new(:name, :time, keyword_init: true); end
+
+        class Caller
+          # @return [Watch]
+          def go
+            Watch.new(name: 'foo')
+          end
+        end
+      ))
+      expect(checker.problems.map(&:message)).not_to include(a_string_matching(/Missing keyword argument/))
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/convention/struct_definition_spec.rb
```diff
@@ -23,6 +23,19 @@
...
     end
 
+    it 'treats keyword args as optional, since Ruby defaults omitted members to nil' do
+      source = Solargraph::SourceMap.load_string(%(
+        # @param bar [String]
+        # @param baz [Integer]
+        Foo = Struct.new(:bar, :baz, keyword_init: true)
+      ), 'test.rb')
+
+      # @type [Array<Solargraph::Pin::Parameter>]
```
apiology: Drop @type; and add sg-ignore as needed, adding or finding something in rules.rb

PR #1273, opened 2026-08-06
Stop dropping union members that differ only by type parameter

### LINKED ISSUE #1272: Same-class generic resolution binds to the first union member instead of the actual member
**Claude:** drafted while investigating an unrelated PR; posting on the user's behalf via their GitHub credentials.

## Summary

When a value's declared type is a union of two instantiations of the *same* `@generic` class (e.g. `Box<Integer>, Box<String>`), calling a method whose return type is `generic<T>` always resolves `T` against whichever instantiation is listed **first** in the union - regardless of which one the value would actually be at runtime, and regardless of any argument that might otherwise disambiguate it. Swapping the declared order of the union members changes the inferred type, which shouldn't be possible for a real union.

## Reproduction

```ruby
# @generic T
class Box
  # @return [generic<T>]
  def get; end
end

class Repro
  # @param b [Box<Integer>, Box<String>]
  # @return [void]
  def process(b)
    # @type [Integer]
    x = b.get
  end
end
```

`solargraph typecheck --level strong` on this: clean, `b.get` infers `Integer`.

Now swap the union's declared order, changing nothing else:

```ruby
  # @param b [Box<String>, Box<Integer>]
  # @return [void]
  def process(b)
    # @type [Integer]
    x = b.get
  end
```

```
Declared type Integer does not match inferred type String for variable x
```

`b.get` now infers `String` - the *only* thing that changed was which member is listed first.

## Root cause

`Chain::Call#inferred_pins` computes `self_type` once for the whole call and uses it to resolve the class's generic parameters:

```ruby
self_type = if head?
              name_pin.context
            else
              name_pin.binder   # the *whole* union/intersection type
            end
...
type = with_params(new_return_type.self_to_type(self_type), self_type).qualify(api_map, *p.gates)
```

`self_type` here is `name_pin.binder` - the complete union type (`Box<Integer>, Box<String>`), not a specific member. Whatever generic-resolution logic binds `Box`'s `T` against that combined `self_type` ends up matching structurally against the first member it finds, so `generic<T>` always resolves to the first union member's argument, independent of which member is actually relevant to the call.

## Scope

This reproduces with a plain union (`,`), no intersection types (`&`) or any other new syntax involved - it's independent of [castwide/solargraph#1231](https://github.com/castwide/solargraph/pull/1231) (which adds intersection types) and [#1266](https://github.com/castwide/solargraph/pull/1266) (structural RBS interface checks). It also affects intersection-typed receivers the same way, since intersections and unions both flatten to the same per-member dispatch code in `Chain::Call`, but a plain union is sufficient to reproduce it on its own.

## Environment

Reproduced on `castwide/solargraph@8fda63384` (current `master`).

apiology:
**Problem:** A method called on a variable typed as two instantiations of the same generic class returns only one of them — the rest of the union is silently dropped.

```ruby
# @generic T
class Box
  # @return [generic<T>]
  def get; end
end

# @param b [Box<Integer>, Box<String>]
def one_order(b)
  b.get  # => Integer   (expected Integer, String)
end

# @param b [Box<String>, Box<Integer>]
def other_order(b)
  b.get  # => String    (expected Integer, String)
end
```

Every union of same-class generics is affected, no declaration order returns the full union, and the truncated result reads as a legitimate concrete type rather than an error.

**Solution:** Dedup the per-member method pins on both `path` and `return_type.rooted_tags` instead of `path` alone, using `rooted_tags` rather than `tag` because `ComplexType` defines no `#tag` and so falls through to its first union member.

Fixes castwide/solargraph#1272.

### DIFF
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -70,7 +70,7 @@ module Solargraph
           # been resolved to different return types for their
           # respective context. Dedup on both so we don't silently
           # drop every member but the first.
-          # @param p [Pin::Base]
+          # @sg-ignore uniq's block param isn't inferred from the receiver's element type (Array#first is)
           pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.tag] }
           return [] if pins.empty?
           inferred_pins(pins, api_map, name_pin, locals)
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -64,10 +64,10 @@ module Solargraph
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          # Union members resolve the same method path against different
-          # contexts, so pins can share a path yet return different types;
-          # keying on path alone would drop all but the first from the union.
-          # @sg-ignore uniq's block param isn't inferred from the receiver's element type
+          # Dedup on path and resolved return type. Union members resolve the
+          # same path against different contexts, so pins can share a path yet
+          # return different types; keying on path alone drops all but the first.
+          # @sg-ignore Array#flatten returns a bare Array, so uniq's block param has no element type to infer from
           pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.rooted_tags] }
           return [] if pins.empty?
           inferred_pins(pins, api_map, name_pin, locals)
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -64,7 +64,14 @@ module Solargraph
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Different union members can resolve to pins that share a
+          # path (e.g. the same generic method looked up against
+          # `Box<Integer>` and `Box<String>`) but that have already
+          # been resolved to different return types for their
+          # respective context. Dedup on both so we don't silently
+          # drop every member but the first.
+          # @param p [Pin::Base]
+          pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.tag] }
           return [] if pins.empty?
           inferred_pins(pins, api_map, name_pin, locals)
         end
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -64,14 +64,11 @@ module Solargraph
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          # Different union members can resolve to pins that share a
-          # path (e.g. the same generic method looked up against
-          # `Box<Integer>` and `Box<String>`) but that have already
-          # been resolved to different return types for their
-          # respective context. Dedup on both so we don't silently
-          # drop every member but the first.
-          # @sg-ignore uniq's block param isn't inferred from the receiver's element type (Array#first is)
-          pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.tag] }
+          # Union members resolve the same method path against different
+          # contexts, so pins can share a path yet return different types;
+          # keying on path alone would drop all but the first from the union.
+          # @sg-ignore uniq's block param isn't inferred from the receiver's element type
+          pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.rooted_tags] }
           return [] if pins.empty?
           inferred_pins(pins, api_map, name_pin, locals)
         end
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -65,7 +65,7 @@ module Solargraph
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
           # Several union members can end up with the same pin and return type.
-          # @sg-ignore Array#flatten returns a bare Array, so uniq's block param has no element type to infer from
+          # @sg-ignore Array#flatten returns a bare Array
           pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.rooted_tags] }
           return [] if pins.empty?
           inferred_pins(pins, api_map, name_pin, locals)
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -64,9 +64,7 @@ module Solargraph
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          # Dedup on path and resolved return type. Union members resolve the
-          # same path against different contexts, so pins can share a path yet
-          # return different types; keying on path alone drops all but the first.
+          # Several union members can end up with the same pin and return type.
           # @sg-ignore Array#flatten returns a bare Array, so uniq's block param has no element type to infer from
           pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.rooted_tags] }
           return [] if pins.empty?
--- a/spec/source/chain/call_spec.rb
+++ b/spec/source/chain/call_spec.rb
@@ -374,6 +374,30 @@ describe Solargraph::Source::Chain::Call do
     expect(type.tag).to eq('String')
   end
 
+  it 'resolves same-class generics from a union independently of declaration order' do
+    # https://github.com/castwide/solargraph/issues/1272
+    ['Box<Integer>, Box<String>', 'Box<String>, Box<Integer>'].each do |union_tag|
+      source = Solargraph::Source.load_string(%(
+        # @generic T
+        class Box
+          # @return [generic<T>]
+          def get; end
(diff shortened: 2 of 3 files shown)


spec/source/chain/call_spec.rb
```diff
@@ -401,6 +401,30 @@ def bar; end
     expect(type.tag).to eq('String')
   end
 
+  it 'resolves same-class generics from a union independently of declaration order' do
+    # https://github.com/castwide/solargraph/issues/1272
```
apiology: Drop issue reference unless you want to add a pending() referencing it

lib/solargraph/source/chain/call.rb
```diff
@@ -64,7 +64,14 @@ def resolve api_map, name_pin, locals
...
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Different union members can resolve to pins that share a
+          # path (e.g. the same generic method looked up against
+          # `Box<Integer>` and `Box<String>`) but that have already
+          # been resolved to different return types for their
+          # respective context. Dedup on both so we don't silently
+          # drop every member but the first.
+          # @param p [Pin::Base]
```
apiology: You shouldn't need a param here; add an sg-ignore if needed

spec/source/chain/call_spec.rb
apiology: **Claude:** Dropped. The `# https://github.com/castwide/solargraph/issues/1272` line is no longer in this example as of 104e2f22d, and no `pending()` was added in its place.

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Removed `@param p [Pin::Base]` and replaced it with a single-line `@sg-ignore`, as of 104e2f22d.

The suppression is still needed at head: the strong typecheck on 104e2f22d reports no `Unneeded @sg-ignore comment` against that line. The only such finding in this file is `call.rb:61`, which is unchanged from master.

lib/solargraph/source/chain/call.rb
```diff
@@ -64,7 +64,14 @@ def resolve api_map, name_pin, locals
...
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Different union members can resolve to pins that share a
+          # path (e.g. the same generic method looked up against
+          # `Box<Integer>` and `Box<String>`) but that have already
+          # been resolved to different return types for their
+          # respective context. Dedup on both so we don't silently
+          # drop every member but the first.
+          # @sg-ignore uniq's block param isn't inferred from the receiver's element type (Array#first is)
```
apiology: Drop parenthetical please

lib/solargraph/source/chain/call.rb
```diff
@@ -64,7 +64,14 @@ def resolve api_map, name_pin, locals
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Different union members can resolve to pins that share a
+          # path (e.g. the same generic method looked up against
+          # `Box<Integer>` and `Box<String>`) but that have already
+          # been resolved to different return types for their
+          # respective context. Dedup on both so we don't silently
+          # drop every member but the first.
```
apiology: Why is it safe to dedupe on .tags rather than .rooted_tags?

Why does not-deduping result in dropping every member (you mean unioned type?)

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Dropped in b5841aae7. The marker now reads:

```ruby
# @sg-ignore uniq's block param isn't inferred from the receiver's element type
```

lib/solargraph/source/chain/call.rb
apiology: **Claude:** It was not safe, and it was worse than `tags` — the code said `.tag`, singular. `ComplexType` defines no `#tag`, so it fell through `method_missing` to `@items.first.tag`:

```ruby
t = Solargraph::ComplexType.parse("Integer, String")
t.items.map(&:tag)  # => ["Integer", "String"]
t.tag               # => "Integer"        <- only the first member
t.tags              # => "Integer, String"
t.rooted_tags       # => "Integer, String"

Solargraph::ComplexType.parse("Integer, Float").tag  # => "Integer" — same key
```

So two pins sharing a path whose resolved return types genuinely differ past the first member keyed identically and one was dropped — the same class of bug this PR is fixing, one level down.

Switched to `rooted_tags` in b5841aae7: it maps every member, and `rooted_tag` keeps the `::` rooting so `Box` and `::Box` do not collide either.

On the second question — yes, the unioned type. `pin_groups` holds one pin per union member, each already resolved against its own context; `uniq` keeps only the first per key, and everything dropped there never reaches `inferred_pins`, so it is absent from the inferred union. Reworded to say that:

```ruby
# Union members resolve the same method path against different
# contexts, so pins can share a path yet return different types;
# keying on path alone would drop all but the first from the union.
```

Typecheck is unchanged: 591 problems before and after in the same worktree, zero added and zero removed.

lib/solargraph/source/chain/call.rb
```diff
@@ -64,7 +64,11 @@ def resolve api_map, name_pin, locals
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Union members resolve the same method path against different
+          # contexts, so pins can share a path yet return different types;
+          # keying on path alone would drop all but the first from the union.
+          # @sg-ignore uniq's block param isn't inferred from the receiver's element type
```
apiology: If we have a likely open PR that addresses this, list its URL instead here

lib/solargraph/source/chain/call.rb
```diff
@@ -64,7 +64,11 @@ def resolve api_map, name_pin, locals
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Union members resolve the same method path against different
```
apiology: Start with a terse statement of what the next line does before explaining why you don't do it a certain way.

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Searched; no open PR matches, so I left prose and corrected it instead — the old wording blamed the wrong method.

`Array#flatten` is declared in core RBS as returning a bare `Array`, so there is no element type left for `uniq`'s block param to bind to:

```ruby
# @param arr [Array<String>]
def check(arr)
  arr.compact   # => ::Array<::String>
  arr.flatten   # => ::Array          <- element type erased
end
```

Marker now reads:

```ruby
# @sg-ignore Array#flatten returns a bare Array, so uniq's block param has no element type to infer from
```

The two nearest open PRs are different shapes: [#1326](https://github.com/castwide/solargraph/pull/1326) is a method-level generic bound by an argument passed alongside a block (`each_with_object`), and [#1274](https://github.com/castwide/solargraph/pull/1274) is a generic return type lost when the declaring method takes a block. Neither covers a receiver that simply has no element type.

Separately, isolating this turned up an unrelated bug with no marker and no issue: a block call chained directly onto another block call loses its block param's type, and splitting the chain into two statements fixes it.

```ruby
# @param arr [Array<String>]
def chained(arr)
  arr.map { |s| s }.uniq { |s| s.upcase }   # Unresolved call to upcase
end

# @param arr [Array<String>]
def split(arr)
  mapped = arr.map { |s| s }
  mapped.uniq { |s| s.upcase }              # resolves
end
```

Not filed — flagging it here rather than acting on it.

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Reordered in 22f277c53 — the action first, the rationale after:

```ruby
# Dedup on path and resolved return type. Union members resolve the
# same path against different contexts, so pins can share a path yet
# return different types; keying on path alone drops all but the first.
```

lib/solargraph/source/chain/call.rb
```diff
@@ -64,7 +64,11 @@ def resolve api_map, name_pin, locals
             [stack.first].compact
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
+          # Dedup on path and resolved return type. Union members resolve the
```
apiology: Clarify why we have to dedupe.  

The rest is a changelog, which belongs in a commit message rather than a comment.  It certainly doesn't need 2.5 lines

apiology: **Claude:** Cut to one line in 99ab1aeb0:

```ruby
# Several union members can end up with the same pin and return type.
# @sg-ignore Array#flatten returns a bare Array, so uniq's block param has no element type to infer from
pins = pin_groups.flatten.uniq { |p| [p.path, p.return_type.rooted_tags] }
```

The removed clause — "keying on path alone drops all but the first" — described the old key rather than the current one, so it moved to the commit message.

PR #1274, opened 2026-08-07
Keep the generic return type when a method takes a block

### LINKED ISSUE #1265: @generic T return type is lost when the method also declares a block parameter
**Claude:** Written by Claude and posted using @apiology's GitHub credentials, while auditing `@sg-ignore` usage in a downstream project.

### Summary

A method that declares `@generic T` and returns `generic<T>` correctly propagates the bound type to callers - *unless* the method also declares a block parameter (`&block`), in which case the generic binding is lost entirely and the return is inferred as untyped `Object`, even if the block parameter is never referenced in the method body.

### Minimal reproduction

```ruby
# typed: true
# frozen_string_literal: true

class Foo
  def initialize; end
  def foo_method; 1; end
end

class Repro
  # @generic T
  # @param clazz [Class<generic<T>>]
  # @return [generic<T>]
  def create_object(clazz)
    clazz.new
  end

  # No block param - generic binds correctly.
  # @generic T
  # @param clazz [Class<generic<T>>]
  # @return [generic<T>]
  def wrap_no_block(clazz)
    create_object(clazz)
  end

  # Same signature plus one unused block param - generic binding is lost.
  # @generic T
  # @param clazz [Class<generic<T>>]
  # @return [generic<T>]
  def wrap_with_block(clazz, &unused)
    create_object(clazz)
  end

  def use_no_block
    obj = wrap_no_block(Foo)
    obj.foo_method # resolves fine
  end

  def use_with_block
    obj = wrap_with_block(Foo)
    obj.foo_method # Unresolved call to foo_method
  end
end
```

```
...
Unresolved call to foo_method
```

Only `use_with_block` fails. `use_no_block` - identical except `wrap_no_block` has no block parameter - resolves `obj` to `Foo` correctly and `obj.foo_method` typechecks clean.

### Expected behavior

`wrap_with_block`'s declared `@return [generic<T>]` should bind `T` to `Foo` from the `clazz` argument the same way `wrap_no_block` does; the presence of an unrelated, unused block parameter shouldn't affect return-type generic binding at all.

### Environment

- Found via `apiology/solargraph` fork, branch `2026-08-04`, commit `d240e5d49c77379dbf46d94b3bf9efef93290343` (based on castwide/solargraph)
- Ruby 3.3.12

### Real-world impact

Found in a test helper (`get_test_object(clazz, &twiddle_mocks)`) that's meant to return a properly-typed instance of an arbitrary class via a generic type parameter, while also accepting an optional block for post-construction mock setup. The block parameter's mere presence silently degrades every caller's return type to `Object`, producing "Unresolved call" errors on every method called on the constructed object - currently worked around with `@sg-ignore`.

apiology:
**Problem:** Adding an unused `&block` parameter to a generic method stops its generic from binding, so ordinary arguments are rejected and the call infers as `generic<T>` rather than the type passed in.

```ruby
class Repro
  # @generic T
  # @param obj [generic<T>]
  # @return [generic<T>]
  def with_retries(obj, &unused)
    obj
  end
end

Repro.new.with_retries('hi')
# Wrong argument type for Repro#with_retries: obj expected generic<T>, received String
```

Deleting `&unused` and changing nothing else makes it resolve, so any generic helper that takes a block is unusable with no indication why.

**Solution:** Signatures now record whether a block is mandatory -- only RBS can express that, as `{ ... }` versus `?{ ... }` -- and a blockless call is rejected only when it is, so a bare `&block` no longer discards the signature while `Array#each`-style Enumerator overloads still disambiguate.

Method-scoped generics are then left unresolved against the receiver, because they bind per call site rather than to the namespace's type parameters.

Fixes #1265

Generated with Claude Code

https://claude.ai/code/session_01TNnFsUN4Uryo6Xqh7xv2sr

### DIFF
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -443,12 +443,8 @@ module Solargraph
       # @param context_type [ComplexType] The receiver type
       # @return [UniqueType, ComplexType]
       def resolve_generics definitions, context_type
-        # Method/signature-scoped generics (e.g. a `@generic T` on the
-        # method itself) aren't resolved against a receiver's type
-        # params -- they're bound per call site via
-        # resolve_generics_from_context. Leave them unresolved here
-        # rather than mistaking the method's own generics for the
-        # namespace's and erasing them to undefined.
+        # Method-scoped generics bind per call site via
+        # resolve_generics_from_context, not against a receiver's type params.
         return self if definitions.is_a?(Pin::Callable)
         return self if definitions.nil? || definitions.generics.empty?
 
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -443,6 +443,13 @@ module Solargraph
       # @param context_type [ComplexType] The receiver type
       # @return [UniqueType, ComplexType]
       def resolve_generics definitions, context_type
+        # Method/signature-scoped generics (e.g. a `@generic T` on the
+        # method itself) aren't resolved against a receiver's type
+        # params -- they're bound per call site via
+        # resolve_generics_from_context. Leave them unresolved here
+        # rather than mistaking the method's own generics for the
+        # namespace's and erasing them to undefined.
+        return self if definitions.is_a?(Pin::Callable)
         return self if definitions.nil? || definitions.generics.empty?
 
         transform(name) do |t|
--- a/lib/solargraph/pin/callable.rb
+++ b/lib/solargraph/pin/callable.rb
@@ -14,12 +14,17 @@ module Solargraph
       # @param block [Signature, nil]
       # @param return_type [ComplexType, nil]
       # @param parameters [::Array<Pin::Parameter>]
+      # @param block_required [Boolean] Whether callers must pass a block for
+      #   this signature to apply. Only ever true for RBS-sourced signatures
+      #   with a non-optional block (`{ ... }` rather than `?{ ... }`); a bare
+      #   `&block` parameter or YARD @yield tag never makes a block mandatory.
       # @param [Hash{Symbol => Object}] splat
-      def initialize block: nil, return_type: nil, parameters: [], **splat
+      def initialize block: nil, return_type: nil, parameters: [], block_required: false, **splat
         super(**splat)
         @block = block
         @return_type = return_type
         @parameters = parameters
+        @block_required = block_required
       end
 
       def reset_generated!
@@ -55,6 +60,7 @@ module Solargraph
       def combine_with other, attrs = {}
         new_attrs = {
           block: combine_blocks(other),
+          block_required: block_required? || other.block_required?,
           return_type: combine_return_type(other)
         }.merge(attrs)
         new_attrs[:parameters] = choose_parameters(other).clone.freeze unless new_attrs.key?(:parameters)
@@ -241,7 +247,7 @@ module Solargraph
         argcount = arguments.length
         parcount = mandatory_positional_param_count
         parcount -= 1 if !parameters.empty? && parameters.last.block?
-        return false if block? && !with_block
+        return false if block? && block_required? && !with_block
         # @todo this and its caller should be changed so that this can
         #   look at the kwargs provided and check names against what
         #   we acccept
@@ -265,6 +271,11 @@ module Solargraph
 
       def block?
         !!@block
+      end
+
+      # @return [Boolean]
+      def block_required?
+        !!@block_required
       end
 
       protected
--- a/lib/solargraph/pin/callable.rb
+++ b/lib/solargraph/pin/callable.rb
@@ -14,10 +14,8 @@ module Solargraph
       # @param block [Signature, nil]
       # @param return_type [ComplexType, nil]
       # @param parameters [::Array<Pin::Parameter>]
-      # @param block_required [Boolean] Whether callers must pass a block for
-      #   this signature to apply. Only ever true for RBS-sourced signatures
-      #   with a non-optional block (`{ ... }` rather than `?{ ... }`); a bare
-      #   `&block` parameter or YARD @yield tag never makes a block mandatory.
+      # @param block_required [Boolean] Whether callers must pass a block, per
+      #   an RBS non-optional block (`{ ... }` rather than `?{ ... }`).
       # @param [Hash{Symbol => Object}] splat
       def initialize block: nil, return_type: nil, parameters: [], block_required: false, **splat
         super(**splat)
--- a/spec/type_checker/levels/strong_spec.rb
+++ b/spec/type_checker/levels/strong_spec.rb
@@ -128,6 +128,58 @@ describe Solargraph::TypeChecker do
       expect(checker.problems.map(&:message)).to be_empty
     end
 
+    it 'does not leak nil from an earlier &. into an unrelated later call in the same chain' do
+      checker = type_checker(%(
+        class Repro
+          # @param x [String, nil]
+          # @return [Boolean]
+          def process(x)
+            x&.to_s == '1'
+          end
+        end
+      ))
+      expect(checker.problems.map(&:message)).to be_empty
+    end
+
+    it 'still flags a chain ending in a safe navigation call as nullable' do
+      checker = type_checker(%(
+        class Repro
+          # @param x [String, nil]
+          # @return [String]
+          def process(x)
+            x&.to_s
+          end
+        end
+      ))
+      expect(checker.problems.map(&:message)).not_to be_empty
+    end
+
+    it 'still flags nil leaking through a self-returning call after an earlier &.' do
+      checker = type_checker(%(
+        class Repro
+          # @param x [String, nil]
+          # @return [String]
+          def process(x)
+            x&.to_s.itself
+          end
+        end
+      ))
+      expect(checker.problems.map(&:message)).not_to be_empty
+    end
+
+    it 'does not flag a call after &. whose result on NilClass is a fixed non-nil type' do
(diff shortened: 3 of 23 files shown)


apiology:
🤖 Posted by Claude, not the account owner — acting on their behalf via their GitHub credentials.

Re-tested this fix against `apiology/checkoff` (pulling the `apiology/solargraph` fork's `2026-08-04` branch, which has this PR's commits merged in as of `0cb1a0e`). The original #1265 symptom is genuinely gone — 71 `@sg-ignore` workarounds for it now correctly report `Unneeded @sg-ignore comment` and were removed. But the same update introduces a new regression: a generic type parameter now leaks into inferred types for code that has nothing to do with any `@generic` declaration at all — 50 new failures appeared across 13 files in that project alone.

### Minimal reproduction

```ruby
# typed: false
# frozen_string_literal: true

class SingleFetch
  # @param data [Hash{String => String}]
  # @return [String]
  def value(data)
    data.fetch('key')
  end
end
```

```
$ bundle exec solargraph typecheck --level strong single_fetch.rb
single_fetch.rb:7: Declared return type ::String does not match inferred type ::String, generic<X> for SingleFetch#value
```

No block param, no `@generic` tag anywhere in the file or the rest of the loaded workspace — just a single `Hash#fetch` call on a `Hash{String => String}`-typed parameter.

### Bisection

Same file, same `.solargraph.yml`, only the `solargraph` gem revision changed:

- `apiology/solargraph@10c5a5502259088903e41010e295a16c2bd7cdaf` (before this PR's commits): **0 problems**.
- `apiology/solargraph@0cb1a0e24cf07b1aa2af66aa17295ff79509c69e` (`2026-08-04` branch with this PR merged in): **1 problem**, as shown above.

### Observed pattern

Across the 50 new failures in `checkoff`, the shape is consistently a *union* — the correctly-resolved type survives, with an extra unresolved generic placeholder appended alongside it (`String, generic<X>`, `Hash{String => String}, generic<X>`, `String, generic<T>`, etc.) — never a total loss of type info the way the original #1265 bug was. The placeholder letter varies (`X` in `Hash#fetch`-shaped failures, `T` in others) and doesn't correspond to any generic name declared in the affected project, which points at internal generic-parameter names (e.g. from `Hash`'s own RBS generic signature) leaking through rather than anything project-specific.

Guess, not confirmed by reading the diff: the fix for #1265 changed the return-type substitution/binding step so it stops *discarding* a generic binding when a block param is present. `Hash`, `Array`, etc. are themselves RBS-generic classes, so every `#fetch`/`#[]` call already exercises that same binding step regardless of whether the calling code ever writes `@generic`. If the fix removed or weakened whatever previously collapsed the placeholder away after substitution, that would explain a placeholder now surviving *alongside* the resolved type on these totally unrelated core-class calls, instead of being fully replaced.

Happy to share the full `checkoff` diff (50 new `@sg-ignore tool-limitation:pr-1274-follow-on` markers, one per failure site) if useful for cross-referencing failure shapes.

apiology:
🤖 Posted by Claude, not the account owner — acting on their behalf via their GitHub credentials.

This reproduces, but it's a stale on-disk pin cache, not a regression in this PR's diff.

`PinCache.work_dir` keys the cache directory only on `Solargraph::VERSION`, a static string this PR doesn't bump. `Marshal.load` silently tolerates deserializing pins written by the older `Pin::Callable` schema (a missing `@block_required` ivar just reads back as `nil`/`false`). So a `bundle update` that swaps the gem revision without also clearing `~/.cache/solargraph` reuses pins from before this fix — `block_required?` comes back `false` for an RBS signature whose block is genuinely required, the block-only generic overload spuriously matches a no-block call site, and `generic<X>` leaks into the inferred return type. Matches the reported shape exactly.

Confirmed directly: populated the cache on the commit before this PR, switched to the commit with this PR merged under the same `Solargraph::VERSION`, without clearing the cache — reproduced `single_fetch.rb`'s exact error. A fresh cache (or clearing `~/.cache/solargraph`) makes it go away; no code change needed here.

Opened a follow-up so this class of bug can't recur silently on future PRs that don't bump VERSION: folds a digest of solargraph's own `lib/` into the cache key so any code change busts it. apiology/solargraph#54: https://github.com/apiology/solargraph/pull/54 (draft/WIP)

apiology:
🤖 Posted by Claude, not the account owner — acting on their behalf via their GitHub credentials.

Found a related gap while auditing `@sg-ignore` suppressions in a downstream project. This PR's title says the loss happens "when the method also declares a block parameter," but the minimal repro below shows neither an explicit `&block` parameter nor `.yield`-via-block-object is required — bare `yield` alone triggers the identical failure, as long as `@generic T` is combined with `@yieldreturn [generic<T>]`.

### Reproduction

```ruby
class Repro
  # @generic T
  # @yieldreturn [generic<T>]
  # @return [generic<T>]
  def call
    yield
  end
end
```

```
repro.rb:5: Repro#call return type could not be inferred
```

Reproduced on `apiology/solargraph@71f16f832d6ee1790c3db0f868ff0335502438da` (`2026-08-04` branch, includes this PR's commits).

apiology:
Ready for review

apiology:

Confirming this fixes the reported symptom: with this branch in our baseline, both arms of #1265's repro (`wrap_no_block` and `wrap_with_block`) now typecheck clean, so the block parameter no longer destroys the binding.

The helper underneath them still doesn't resolve, though — `clazz.new` on a `Class<generic<T>>` parameter isn't inferred as `generic<T>` inside the method body:

```ruby
class Foo
  # @return [Integer]
  def foo_method
    1
  end
end

class Repro
  # @generic T
  # @param clazz [Class<generic<T>>]
  # @return [generic<T>]
  def create_object(clazz)
    clazz.new
  end

  # @param clazz [Class<Foo>]
  # @return [Foo]
  def create_concrete(clazz)
    clazz.new
  end

  # @return [Integer]
  def use_it
    create_object(Foo).foo_method
  end
end
```

```
repro.rb:15: Repro#create_object return type could not be inferred
1 problem found.
```

`create_concrete` is the control: `Class<Foo>#new` resolves to `Foo` fine, so it's specific to the generic form. `use_it` also passes — callers bind `T` correctly from the argument; it's only the method's own declared return that can't be inferred from `clazz.new`.

That accounts for 7 suppressions in our codebase, separate from the 14 tied to #1265 itself.

apiology:

```ruby
clazz.new(:foo, **args)   # Too many arguments to Class#initialize
```

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -443,6 +443,13 @@ def resolve_param_generics_from_context generics_to_resolve, context_type, resol
       # @param context_type [ComplexType] The receiver type
       # @return [UniqueType, ComplexType]
       def resolve_generics definitions, context_type
+        # Method/signature-scoped generics (e.g. a `@generic T` on the
+        # method itself) aren't resolved against a receiver's type
+        # params -- they're bound per call site via
+        # resolve_generics_from_context. Leave them unresolved here
+        # rather than mistaking the method's own generics for the
+        # namespace's and erasing them to undefined.
```
apiology: Please limit to 2 lines

lib/solargraph/pin/callable.rb
```diff
@@ -14,12 +14,17 @@ class Callable < Closure
       # @param block [Signature, nil]
       # @param return_type [ComplexType, nil]
       # @param parameters [::Array<Pin::Parameter>]
+      # @param block_required [Boolean] Whether callers must pass a block for
+      #   this signature to apply. Only ever true for RBS-sourced signatures
+      #   with a non-optional block (`{ ... }` rather than `?{ ... }`); a bare
+      #   `&block` parameter or YARD @yield tag never makes a block mandatory.
```
apiology: limit to 2 lines

spec/type_checker/levels/strong_spec.rb
```diff
@@ -661,6 +661,20 @@ def block_pins
...
+        class Repro
+          # @generic T
+          # @yieldreturn [generic<T>]
+          # @return [generic<T>]
+          def call
+            yield
+          end
+        end
+      ))
+      expect(checker.problems.map(&:message)).to be_empty
```
apiology: Your assertion says we should not give an error for the above.   Let's add a spec documenting what we should derive from:

x = Repro.new.call { 123 }
x

lib/solargraph/complex_type/unique_type.rb
apiology: **Claude:**

Trimmed to two lines in 865926a67:

```ruby
# Method-scoped generics bind per call site via
# resolve_generics_from_context, not against a receiver's type params.
```

lib/solargraph/pin/callable.rb
apiology: **Claude:**

Trimmed to two lines in 865926a67:

```ruby
# @param block_required [Boolean] Whether callers must pass a block, per
#   an RBS non-optional block (`{ ... }` rather than `?{ ... }`).
```

spec/type_checker/levels/strong_spec.rb

Added in 865926a67, directly below the existing one. The declared `@type [String]` is deliberately wrong so the reported problem names the inferred type:

```ruby
it 'derives a method-scoped generic return type from a block result' do
  checker = type_checker(%(
    class Repro
      # @generic T
      # @yieldreturn [generic<T>]
      # @return [generic<T>]
      def call
        yield
      end
    end
    # @type [String]
    x = Repro.new.call { 123 }
    x
  ))
  expect(checker.problems.map(&:message))
    .to eq(['Declared type String does not match inferred type Integer for variable x'])
end
```

So `x` derives as `Integer` — the block result flows through `@yieldreturn [generic<T>]` into the `@return [generic<T>]`.

PR #1277, opened 2026-08-08
Stop flagging unreachable code as type errors

### LINKED ISSUE #1276: raise and abort confuse typechecker
## Problem

Two typecheck failures share the same shape: an expression that can only ever raise/abort — never actually produce a value — isn't recognized as automatically satisfying whatever type is expected of it.

**1. A method body that's only `raise` fails typecheck despite an explicit `@return` tag:**

```ruby
# @return [Boolean]
def matches?
  raise 'Override me!'
end
```
```
file.rb:LINE: ...#matches? return type could not be inferred
```

The declared `@return [Boolean]` is never even compared against — the checker reports "could not be inferred" instead. This is a common "must be overridden" abstract-stub shape, and the same failure appears on every raise-only method body regardless of its declared return type.

**2. `Array#fetch`'s block-form generic doesn't resolve when the block ends in `abort`:**

```ruby
issues_path = T.cast(ARGV.fetch(0) { abort("usage: #{$PROGRAM_NAME} FILE") }, String)
```
```
file.rb:185: Wrong argument type for T.cast: value expected Object, received String, generic<T>
```

An unresolved `generic<T>` leaks out of the `fetch` call and fails the very next line that consumes the value — independent of the explicit `T.cast`, and independent of what the block's abort message says.

## Proposed solution

Both failures trace to the same root cause: RBS declares `Kernel#raise`/`#fail` and `Kernel.abort` as returning `bot` (RBS's bottom type — `RBS::Types::Bases::Bottom`), meaning "this expression never actually produces a value, so it's compatible with any type expected of it." But `RbsTranslator.to_complex_type` (`lib/solargraph/rbs_translator.rb:253`) collapses `Bottom` into the same `'undefined'` tag used for `Any`:

```ruby
when RBS::Types::Bases::Any, RBS::Types::Bases::Bottom
  # `Bottom`` is used in contexts where nothing will ever return
  # - e.g., it could be the return type of 'exit()' or 'raise'
  # @todo define a specific bottom type and use it to
  #   determine dead code
  'undefined'
```

`undefined` already means something else downstream — "the type couldn't be determined" — and every consumer of that tag treats it as a failure mode rather than a trivial pass:

- `type_checker.rb:172` treats any `undefined` inferred return type as "return type could not be inferred," reported *before* the declared `@return` tag is even consulted. That's failure 1.
- `conformance.rb:217-219` only treats `undefined` as auto-compatible under the `:allow_undefined` rule, which `--level strong` doesn't enable. `Array#fetch`'s block-form signature is generic — `[I, T] (I index) { (I index) -> T } -> (E | T)` — and when the block is `{ abort(...) }`, `T` should resolve to `bot` and `E | bot` should simplify to plain `E` (bot is the identity element of a union). Instead `T` resolves to `undefined`, indistinguishable from "couldn't be resolved at all," so the union never collapses. That's failure 2.

Fix, in three small, contained steps:

**1. Give `Bottom` its own tag instead of collapsing it into `Any`'s.** `ComplexType::BOT = ComplexType.parse('bot')` (`complex_type.rb:757`) already exists as a distinct, parseable type — it's just never produced. Change `rbs_translator.rb:253`:

```ruby
when RBS::Types::Bases::Any
  'undefined'
when RBS::Types::Bases::Bottom
  'bot'
```

**2. Make `type_checker.rb`'s return-type check skip the "could not be inferred" branch for a `bot`-inferred type**, so it falls through to the normal conformance comparison (or is treated as an automatic pass, equivalent to `void`) instead of short-circuiting before the declared `@return` tag is even consulted (`type_checker.rb:172`, alongside the existing `inferred.undefined?` check). This resolves failure 1.

**3. Make `conformance.rb` treat `bot` as unconditionally compatible with any expected type**, not gated behind `:allow_undefined` the way plain `undefined` is — this isn't a leniency knob, it's a type-theoretic fact (bot is a subtype of everything). Add this alongside the existing `undefined?` carve-out at `conformance.rb:217-219`. Once `T` correctly resolves to `bot` instead of `undefined` (step 1) and `bot` is recognized as absorbable into a union (`E | bot => E`), failure 2 should resolve via the same existing union-simplification logic that already drops duplicate/subtype members — worth confirming after steps 1-3 land before writing any bespoke handling for the generic case specifically.

apiology:

Fixes https://github.com/castwide/solargraph/issues/1276

**Problem:** Code that can never execute — a method that only ever raises, or a branch a type guard has already ruled out — still gets flagged as a type error.

```ruby
# @return [String]
def broken
  raise 'always fails'
end
# solargraph typecheck: broken return type could not be inferred
```

Both shapes force a suppression on code that was never actually wrong, defeating the point of typechecking it in the first place.

**Solution:** RBS already declares `raise`/`abort` as returning `bot` (RBS's bottom type — a subtype of everything); this teaches Solargraph's own type-exclusion logic to produce that same `bot` tag when narrowing rules out every possibility, and makes every consumer that used to treat an unresolved type as a failure — return-type checking, method-call resolution — treat `bot` as trivially valid instead.

**Test plan:**

- [x] Verified locally: this PR's self-typecheck and `.rubocop_todo.yml` regeneration both introduce zero delta beyond what already exists on `master`.

### DIFF
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -253,6 +253,11 @@ module Solargraph
       # @param variance [:invariant, :covariant, :contravariant]
       def conforms_to? api_map, expected, situation, rules = [],
                        variance: erased_variance(situation)
+        # bot is a subtype of every type - not a leniency knob like
+        # :allow_undefined, but a type-theoretic fact (e.g., the return
+        # type of `raise` or `abort`)
+        return true if bot?
+
         return true if undefined? && rules.include?(:allow_undefined)
 
         # @todo teach this to validate duck types as inferred type
@@ -559,7 +564,7 @@ module Solargraph
       def qualify api_map, *gates
         transform do |t|
           next t if t.name == GENERIC_TAG_NAME
-          next t if t.duck_type? || t.void? || t.undefined? || t.literal?
+          next t if t.duck_type? || t.void? || t.undefined? || t.literal? || t.bot?
           open = t.rooted? ? [''] : gates
           fqns = api_map.qualify(t.non_literal_name, *open)
           if fqns.nil?
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -626,6 +626,13 @@ module Solargraph
       end
 
       UNDEFINED = UniqueType.new('undefined', rooted: false)
+      # @note Equal (`==`) to ComplexType::BOT.first - rooted: true
+      #   matches how ComplexType.parse('bot') constructs it, unlike
+      #   UNDEFINED's rooted: false. Getting this wrong keeps #bot?/
+      #   #tag/#to_s working (they don't read @rooted) but silently
+      #   breaks #== and anything relying on it (Array#uniq, Array#-,
+      #   Set membership, pin dedup).
+      BOT = UniqueType.new('bot', rooted: true)
       BOOLEAN = UniqueType.new('Boolean', rooted: true)
       TRUE = UniqueType.new('true', rooted: true)
       FALSE = UniqueType.new('false', rooted: true)
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -626,12 +626,10 @@ module Solargraph
       end
 
       UNDEFINED = UniqueType.new('undefined', rooted: false)
-      # @note Equal (`==`) to ComplexType::BOT.first - rooted: true
-      #   matches how ComplexType.parse('bot') constructs it, unlike
-      #   UNDEFINED's rooted: false. Getting this wrong keeps #bot?/
-      #   #tag/#to_s working (they don't read @rooted) but silently
-      #   breaks #== and anything relying on it (Array#uniq, Array#-,
-      #   Set membership, pin dedup).
+      # #eql? and #hash read the raw @rooted ivar, while #rooted?
+      # reports true for any lowercase name regardless of it. rooted:
+      # true is therefore required for BOT to compare equal to - and
+      # hash alongside - ComplexType::BOT.first.
       BOT = UniqueType.new('bot', rooted: true)
       BOOLEAN = UniqueType.new('Boolean', rooted: true)
       TRUE = UniqueType.new('true', rooted: true)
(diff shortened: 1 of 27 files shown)


apiology:
Issue was accidentally closed.  Did the unit tests in this PR help explain the motivation?

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -621,6 +626,13 @@ def self.can_root_name? name
       end
 
       UNDEFINED = UniqueType.new('undefined', rooted: false)
+      # @note Equal (`==`) to ComplexType::BOT.first - rooted: true
+      #   matches how ComplexType.parse('bot') constructs it, unlike
+      #   UNDEFINED's rooted: false. Getting this wrong keeps #bot?/
+      #   #tag/#to_s working (they don't read @rooted) but silently
+      #   breaks #== and anything relying on it (Array#uniq, Array#-,
+      #   Set membership, pin dedup).
```
apiology: Explain what this is meant to prevent or please drop

apiology: Rewritten rather than dropped. There is a constraint here that survives, but the old text stated it as a bug report instead of as a property of the code.

The comment now reads:

```ruby
UNDEFINED = UniqueType.new('undefined', rooted: false)
# #eql? and #hash read the raw @rooted ivar, while #rooted?
# reports true for any lowercase name regardless of it. rooted:
# true is therefore required for BOT to compare equal to - and
# hash alongside - ComplexType::BOT.first.
BOT = UniqueType.new('bot', rooted: true)
```

`UniqueType#eql?` compares the raw `@rooted` ivar against `other.rooted?`, and `#hash` includes `@rooted` directly. `#rooted?` is `!can_root_name? || @rooted`, and `can_root_name?('bot')` is false because `'bot'` is all lowercase, so `#rooted?` reports `true` for a `bot` type however it was constructed. Only `rooted: true` makes the constant compare equal to, and hash alongside, `ComplexType::BOT.first`:

```ruby
UniqueType.new('bot', rooted: true)  == ComplexType::BOT.first  # => true
UniqueType.new('bot', rooted: false) == ComplexType::BOT.first  # => false
UniqueType.new('bot', rooted: false).tag                        # => "bot"
```

The equality example in `spec/complex_type/unique_type_spec.rb` pins it.

Two statements in the old text were wrong and are also fixed:

- it credited `#==` to "Equality's `equality_fields`"; no such mechanism is involved, `UniqueType#eql?` is hand-written and `#==` delegates to it;
- the spec was titled `is rooted, unlike ::UNDEFINED`, but `UniqueType::UNDEFINED.rooted?` is `true` as well, for the same lowercase-name reason.

PR #1280, opened 2026-08-11
Stop reporting a duck type's own method as unresolved

### LINKED ISSUE #1257: Chain::Call#resolve can't resolve calls to a YARD duck-type param's own declared method

### Background

A method whose param is documented with YARD duck typing (`@param thing [#read_body]`) can't resolve a call to the exact method the duck type names, even directly and with no block:

```ruby
# typed: true
# frozen_string_literal: true

class Repro
  # @param thing [#read_body]
  # @return [void]
  def call(thing)
    thing.read_body
  end
end
```

```
$ bundle exec solargraph typecheck --level strict repro.rb
repro.rb:8: Unresolved call to read_body on #read_body
```

Only surfaces at `--level strict` and above (`strict`, `strong`, `alpha`) — `normal` and `typed` don't report it, since `report_undefined?` (`type_checker/rules.rb:36-38`) gates the "Unresolved call" diagnostic behind `rank >= LEVELS[:strict]`. Confirmed empirically across all four levels:

```
$ for lvl in normal typed strict strong; do echo "=== $lvl ==="; bundle exec solargraph typecheck --level $lvl repro.rb; done
=== normal ===
0 problems found.
=== typed ===
0 problems found.
=== strict ===
repro.rb:8: Unresolved call to read_body on #read_body
1 problem found.
=== strong ===
repro.rb:8: Unresolved call to read_body on #read_body
1 problem found.
```

Reproduces on plain `solargraph 0.60.2`, no project config, default `Gemfile` (`solargraph`, `sorbet-runtime`), with or without a block on the call.

### Root cause

Two different code paths handle duck types inconsistently:

- `ApiMap#get_complex_type_methods` (`lib/solargraph/api_map.rb:544-549`) handles it correctly — for a duck type it explicitly synthesizes a `Pin::DuckMethod` named after the declared method, in addition to merging in `Object`'s methods:

  ```ruby
  if type.duck_type?
    result.add Pin::DuckMethod.new(name: type.to_s[1..], source: :api_map)
    result.merge get_methods('Object')
  ```

- `Chain::Call#resolve` (`lib/solargraph/source/chain/call.rb:53-63`), the path used to resolve an actual method call in source (and therefore what strict typecheck's "Unresolved call" diagnostic depends on), does not use `get_complex_type_methods` at all. It computes a plain namespace tag and looks up the method stack directly on that:

  ```ruby
  binder = binder.without_nil if nullable?
  # @sg-ignore Need to handle duck-typed method calls on union types
  pin_groups = binder.each_unique_type.map do |context|
    ns_tag = context.namespace == '' ? '' : context.namespace_type.tag
    stack = api_map.get_method_stack(ns_tag, word, scope: context.scope)
    [stack.first].compact
  end
  ```

  For a duck type, `context.namespace_type` (`complex_type/type_methods.rb:152-153`) returns `ComplexType.parse('::Object')`, so `ns_tag` becomes `"Object"`. `api_map.get_method_stack('Object', 'read_body', ...)` then looks for `read_body` on `Object`'s real method stack — which doesn't have it, since the whole point of the duck-type tag was to name a method the receiver isn't statically known to have via inheritance. The stack comes back empty, `pins` ends up `[]`, and the call is reported unresolved.

The `@sg-ignore Need to handle duck-typed method calls on union types` comment sitting directly on that code path already acknowledges a version of this gap, though the repro above shows it's not specific to union types — a single, non-union duck-type param hits the same code and fails the same way.

### Proposed direction

`Chain::Call#resolve` should special-case `context.duck_type?` the same way `get_complex_type_methods` does: when the receiver's type is a duck type, treat a call matching the duck type's own declared name as resolved (via a `Pin::DuckMethod`) instead of only checking `Object`'s real method stack. This would likely also fix the adjacent union case the existing `@sg-ignore` comments already flag.

apiology:
**Problem:** A call to a method that a duck-typed parameter itself declares is reported as unresolved.

```ruby
# @param thing [#read_body]
def fetch(thing)
  thing.read_body
end

# solargraph typecheck --level strict
# => Unresolved call to read_body on #read_body
```

It fires at `strict` and above on every duck-typed parameter, even though `ApiMap#get_complex_type_methods` already resolves the same call correctly for completion.

**Solution:** `Chain::Call#resolve` synthesizes a `Pin::DuckMethod` when the receiver is a duck type naming the called method, mirroring `get_complex_type_methods`; the pin is `explicit: false` so arity checking is skipped, since a duck type declares that the method exists, not what signature it has.

Fixes https://github.com/castwide/solargraph/issues/1257

### DIFF
--- a/lib/solargraph/pin/duck_method.rb
+++ b/lib/solargraph/pin/duck_method.rb
@@ -6,11 +6,31 @@ module Solargraph
     # use duck typing, e.g., `@param file [#read]`.
     #
     class DuckMethod < Pin::Method
-      # @param location [Solargraph::Location]
-      # @param name [String]
-      # def initialize location, name
-      #   # super(location, '', name, nil, :instance, :public, [])
-      # end
+      # A duck-type tag has no syntax for arguments, so the default empty
+      # #parameters would synthesize a zero-arg signature no real call matches.
+      #
+      # @param splat [Hash{Symbol => Object}]
+      def initialize **splat
+        super(parameters: accepts_any_arguments, **splat)
+      end
+
+      # @return [::Array<Pin::Parameter>]
+      def accepts_any_arguments
+        [
+          Pin::Parameter.new(decl: :restarg, name: 'args', closure: self, source: :api_map),
+          Pin::Parameter.new(decl: :kwrestarg, name: 'kwargs', closure: self, source: :api_map)
+        ]
+      end
+
+      # A synthetic pin sits in no ancestor chain and has no closure. Letting
+      # the inherited walk find a type sends #typify on to `closure.gates`,
+      # which raises.
+      #
+      # @param _api_map [ApiMap]
+      # @return [Array<Pin::Method>]
+      def rest_of_stack _api_map
+        []
+      end
     end
   end
 end
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -59,9 +59,15 @@ module Solargraph
           binder = binder.without_nil if nullable?
           # @sg-ignore Need to handle duck-typed method calls on union types
           pin_groups = binder.each_unique_type.map do |context|
-            ns_tag = context.namespace == '' ? '' : context.namespace_type.tag
-            stack = api_map.get_method_stack(ns_tag, word, scope: context.scope)
-            [stack.first].compact
+            if context.duck_type? && context.name[1..] == word
+              # explicit: false skips arity checking; the duck type
+              # only tells us the method exists, not its signature
+              [Pin::DuckMethod.new(name: word, source: :chain, explicit: false)]
+            else
+              ns_tag = context.namespace == '' ? '' : context.namespace_type.tag
+              stack = api_map.get_method_stack(ns_tag, word, scope: context.scope)
+              [stack.first].compact
+            end
           end
           pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
           pins = pin_groups.flatten.uniq(&:path)
--- a/spec/pin/duck_method_spec.rb
+++ b/spec/pin/duck_method_spec.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+describe Solargraph::Pin::DuckMethod do
+  let(:api_map) do
+    Solargraph::ApiMap.new.tap do |map|
+      map.map Solargraph::Source.load_string(%(
+        class ClassTest
+          # @param clazz [#new]
+          def create_object(clazz)
+            clazz.new
+          end
+        end
+      ))
+    end
+  end
+
+  # The one place these are built: a duck-type tag reaches ApiMap via its
+  # parsed ComplexType, not via the source that declared it.
+  let(:duck_new) do
+    api_map.get_complex_type_methods(Solargraph::ComplexType.parse('#new'))
+           .grep(described_class).first
+  end
+
+  # ApiMap leaves #closure nil; the call site is what gives an ancestor
+  # walk somewhere to go, since ClassTest inherits its own Class#new.
+  let(:duck_new_at_call_site) do
+    described_class.new(name: 'new', source: :api_map,
+                        closure: api_map.get_path_pins('ClassTest#create_object').first)
+  end
+
+  it 'synthesizes a signature that accepts any arguments' do
+    expect(duck_new.signatures.first.parameters.map(&:decl)).to eq(%i[restarg kwrestarg])
+  end
+
+  it 'has no ancestor chain of its own to walk' do
+    expect(duck_new_at_call_site.rest_of_stack(api_map)).to be_empty
+  end
+
+  it "infers nothing rather than the call site's own inherited #new" do
+    expect(duck_new_at_call_site.typify(api_map)).to be_undefined
+  end
+end
--- a/spec/type_checker/levels/alpha_spec.rb
+++ b/spec/type_checker/levels/alpha_spec.rb
@@ -193,6 +193,19 @@ describe Solargraph::TypeChecker do
       expect(checker.problems.map(&:message)).to eq([])
     end
 
+    it 'resolves calls to a duck type param\'s own declared method' do
+      checker = type_checker(%(
+        class Foo
+          # @param baz [#read_body]
+          # @return [void]
+          def bar(baz)
+            baz.read_body
+          end
+        end
+      ))
+      expect(checker.problems).to be_empty
+    end
+
     it 'resolves self correctly in arguments (second case)' do
       checker = type_checker(%(
         class Blah


apiology:
This PR introduced a `NoMethodError: undefined method 'gates' for nil` crash (not a normal typecheck problem — a hard exception) for us in CI. `Pin::DuckMethod.new` (`source/chain/call.rb:233`) never sets `closure:`, so `Pin::Method#typify`'s unconditional `closure.gates` (`pin/method.rb:302`) crashes; using the pin's own `gates` accessor instead fixes that (verified identical to `closure.gates` across 35 sampled real pins in a full project typecheck). There's also a second, deeper issue in the same method: `typify_from_super` (`pin/method.rb:625`) returns an ancestor pin's raw, *unqualified* `return_type` instead of calling `pin.typify(api_map)` on it (the way `resolve_reference`, a few lines away, already correctly does) — so the borrowed type gets qualified against the duck pin's own (fabricated, empty) gates instead of the ancestor's real ones. This stayed invisible in our crash only because the ancestor we hit (`String#to_s`) is RBS/core and already `::`-rooted, and `ComplexType::UniqueType#qualify` (`unique_type.rb:666`) ignores passed-in gates entirely for rooted types — for any duck-typed call landing on a YARD-documented ancestor with a relative return type, this would silently resolve against the wrong namespace with no crash and no warning.

Both changes together still resolve the original crash (`solargraph typecheck --level strong`: 0 problems) with zero regressions on a full 106-file project typecheck; I don't have a standalone repro for the raw-vs-qualified divergence itself, since duck-dispatch resolution seems to need full-workspace scale to trigger at all.

```ruby
# pin/method.rb:302
qualified = type.qualify(api_map, *gates)

# pin/method.rb:625-631
def typify_from_super api_map
  stack = rest_of_stack api_map
  return nil if stack.empty?
  stack.each do |pin|
    next if pin.return_type.undefined?
    return pin.typify(api_map)
  end
  nil
end
```

🤖 Posted by [Claude Code](https://claude.com/claude-code) on behalf of @apiology.

apiology:
**Claude:** Fixed. Commit eee8ff14f ("Fix nil closure crash and unqualified super return type in Pin::Method#typify") landed the same day as this comment and covers both issues raised: `pin/method.rb` now uses the pin's own `gates` (not `closure.gates`) at line 297, and `typify_from_super` (line 620) now calls `pin.typify(api_map)` instead of returning the raw unqualified `return_type`. Confirmed at current PR head (bfae6ebb2).

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -621,6 +626,13 @@ def self.can_root_name? name
       end
 
       UNDEFINED = UniqueType.new('undefined', rooted: false)
+      # @note Equal (`==`) to ComplexType::BOT.first - rooted: true
+      #   matches how ComplexType.parse('bot') constructs it, unlike
+      #   UNDEFINED's rooted: false. Getting this wrong keeps #bot?/
+      #   #tag/#to_s working (they don't read @rooted) but silently
+      #   breaks #== and anything relying on it (Array#uniq, Array#-,
+      #   Set membership, pin dedup).
```
apiology: Is this true?

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -253,6 +253,11 @@ def erased_version_of? other
       # @param variance [:invariant, :covariant, :contravariant]
       def conforms_to? api_map, expected, situation, rules = [],
                        variance: erased_variance(situation)
+        # bot is a subtype of every type - not a leniency knob like
+        # :allow_undefined, but a type-theoretic fact (e.g., the return
+        # type of `raise` or `abort`)
+        return true if bot?
```
apiology: How does this PR relate to the bottom type PR?  Should it build on top?

lib/solargraph/complex_type/unique_type.rb
apiology: **Claude:** Partly, and the wrong part is now fixed in 5fc683213.

`#hash` does read the raw `@rooted` ivar. `#eql?`/`#==` do not — the source
is `@rooted == other.rooted?`: the raw ivar on the left, the *method* on the
right. For a lowercase name that makes equality asymmetric:

```ruby
UniqueType::BOT == UniqueType.new('bot', rooted: false)  # => true
UniqueType.new('bot', rooted: false) == UniqueType::BOT  # => false
```

So the note's conclusion held but its mechanism did not. `#hash` is what
actually makes a wrong value fail here: `Set[BOT].include?(unrooted)` is
false in both directions and `[BOT, unrooted].uniq.size == 2` in both
orders, while `#==` still answers true one way round.

The rest checked out. `can_root_name?('bot')` is `false`, so `#rooted?`
returns true for every variant regardless of the ivar, and
`bot?`/`tag`/`to_s`/`rooted_tag` are identical either way.
`ComplexType.parse('bot')` sets `rooted = true` via the
`!can_root_name?(name)` branch, so `rooted: true` is required for `BOT` to
hash alongside `ComplexType::BOT.first`.

Two pre-existing things surfaced while checking, both on master and not
introduced by this branch: the asymmetric `#eql?` at `unique_type.rb:193`,
and `#hash` at `:276` reading `@sub_types` when the ivar is `@subtypes` —
so `#hash` never actually incorporates subtypes, masked by `@all_params`
sitting in the same tuple.

lib/solargraph/complex_type/unique_type.rb
apiology: **Claude:** The bottom type PR is #1277. This branch already contains all of
it — by content, not by ancestry.

`git merge-base --is-ancestor 5c7de037a HEAD` exits 1, so #1277's head is not
in this branch's history. But diffing the two heads leaves five of #1277's
eight files byte-identical: `complex_type.rb`, `rbs_translator.rb`,
`type_checker.rb`, `spec/complex_type_spec.rb` and
`spec/type_checker/levels/typed_spec.rb`. What is left between them is only
this branch's own work:

```
 lib/solargraph/complex_type/type_methods.rb | 19 -------------------
 lib/solargraph/complex_type/unique_type.rb  |  7 +++----
 lib/solargraph/pin/method.rb                |  8 ++++----
 lib/solargraph/source/chain/call.rb         | 12 +++++++++---
 spec/complex_type/unique_type_spec.rb       |  6 +++++-
 spec/type_checker/levels/alpha_spec.rb      | 13 +++++++++++++
```

So: yes, it should build on top, and effectively it does. The cost of the
duplication is that this PR's diff against `master` shows #1277's change too,
so both get reviewed twice.

Rebasing `fix-1257-duck-type-own-method` onto `fix-raise-abort-bot-type`
would make the ancestry match the content and reduce this PR to the six files
above. It would not change what GitHub renders yet: the base repo is
`castwide` and `fix-raise-abort-bot-type` exists only in the `apiology` fork,
so the PR base cannot be pointed at it. The combined diff stays until #1277
merges, then collapses on its own.

One thing that surfaced while checking: `ComplexType#bottom?` on master
(`complex_type.rb:430`) calls `@items.all?(&:bot?)`, but no `bot?` is
defined anywhere in master's `lib/` — so it would raise `NoMethodError` if
called. It has no callers in `lib/` or `spec/` today, which is why nothing
catches it. `#1277` adds the missing `TypeMethods#bot?`, and this branch
carries that same addition, so either one landing closes the gap.

lib/solargraph/complex_type/type_methods.rb
```diff
@@ -209,25 +217,6 @@ def == other
...
-          next t if t.duck_type? || t.void? || t.undefined?
-          recon = (t.rooted? ? '' : context)
-          fqns = api_map.qualify(t.name, recon)
-          if fqns.nil?
-            next UniqueType::BOOLEAN if t.tag == 'Boolean'
-            next UniqueType::UNDEFINED
-          end
-          t.recreate(new_name: fqns, make_rooted: true)
-        end
-      end
```
apiology: Please note in PR description why this method is being removed

apiology: **Claude:** Added to the description:

> `TypeMethods#qualify` is deleted as unreachable: the module is included only
> by `UniqueType`, and both it and `ComplexType` define their own `qualify`.

Checked at runtime rather than by reading. `UniqueType.instance_method(:qualify).owner`
is `UniqueType` and `ComplexType.instance_method(:qualify).owner` is
`ComplexType`, so a method defined directly in the class won over the module
copy in both cases. `ComplexType.ancestors.include?(TypeMethods)` is also
false — it reaches `TypeMethods` only through the `method_missing` forward to
`@items.first`, and that never fires for a name the class defines itself.

Nothing dispatched to the module copy, so removing it is not a behaviour
change.

PR #1281, opened 2026-08-11
Improved RBS type alias support

### LINKED ISSUE #1255: Typecheck rejects arguments that satisfy an RBS type alias, because aliases aren't expanded before comparison

### Background

`FileUtils.ln_sf` (and every other `FileUtils` method taking a path) is declared in the stdlib RBS as:

```
def self?.ln_sf: (src ::FileUtils::pathlist, dest ::FileUtils::path, ?noop: boolish, ?verbose: boolish) -> void
```

where `FileUtils::path`/`FileUtils::pathlist` are RBS `type` aliases:

```
type path = string | _ToPath
type pathlist = path | Array[path]
```

(`stdlib/fileutils/0/fileutils.rbs`, `core/builtin.rbs`). `string` is `String`, so a plain `String` argument should satisfy `path` structurally.

`solargraph typecheck --level strong` rejects it anyway:

```
Wrong argument type for FileUtils.ln_sf: src expected FileUtils::path, Array<FileUtils::path>, received String
```

### Minimal repro

```ruby
# typed: true
# frozen_string_literal: true

require 'fileutils'

class Repro
  # @param source [String]
  # @param dest [String]
  # @return [void]
  def link(source, dest)
    FileUtils.ln_sf(source, dest)
  end
end
```

```
repro.rb:10: Wrong argument type for FileUtils.ln_sf: src expected FileUtils::path, Array<FileUtils::path>, received String
```

Reproduces on plain `solargraph 0.60.2`, no project config, no gems beyond the default `Gemfile` (`solargraph`, `sorbet-runtime`).

### Root cause

`RbsTranslator.type_to_tag` (`lib/solargraph/rbs_translator.rb:158-166`) converts an `RBS::Types::Alias` to a type tag using only the alias's own name, never its underlying definition:

```ruby
when RBS::Types::ClassInstance, RBS::Types::Alias, RBS::Types::Interface
  # `Alias` is a top-level type alias, e.g., 'bool' in "type bool = true | false"
  # @todo ensure these get resolved after processing all aliases
  # @todo handle recursive aliases
  #
  # `Interface represents a mix-in module which can be considered a
  # subtype of a consumer of it
  #
  type_tag(type.name, type.args)
```

So `FileUtils::path` becomes the *nominal* type tag `"FileUtils::path"`, not the union `"String, FileUtils::_ToPath"` it's actually defined as. `ComplexType::Conformance` (`lib/solargraph/complex_type/conformance.rb`) has no handling for aliases at all — it never expands one before checking whether an argument's inferred type conforms, so the comparison degenerates to "is `String` the literal type named `FileUtils::path`?", which is always false.

Separately, `Pin::Reference::TypeAlias` pins do get created and indexed (`rbs_map/conversions.rb:71-90`) — so the alias *is* known to Solargraph, just not consulted during argument compatibility checking.

### Related but distinct

- #630 ("Support for TypeAliases") — maintainer-confirmed real bug, but for YARD constant-style aliases (`TypeAlias = Foo::Bar`), not RBS `type X = ...` syntax. Same *shape* of gap (inference sees through it, typecheck doesn't), different mechanism, no linked fix.
- #1232 — RBS *interfaces* being blanket-allowed rather than structurally verified. Different bug (leniency vs. this one's rejection), but adjacent: fixing structural interface conformance per #1232 wouldn't by itself fix this, since the top-level problem here is the alias never being expanded into its member types (interface or otherwise) before any conformance check runs.

### Proposed direction

In `RbsTranslator.type_to_tag`, when `type` is an `RBS::Types::Alias`, resolve it to its underlying `RBS::AST::Declarations::TypeAlias#type` and recurse into `type_to_tag` on that, instead of emitting a nominal tag from the alias's name. The existing `@todo` comments at that exact call site already flag this as known/intended future work — this issue is a repro plus root-cause pointer for picking it up. Recursive aliases (also flagged in the `@todo`) would need cycle detection if tackled the same way.

apiology:
**Problem:** `solargraph typecheck --level strong` rejects valid arguments when a parameter is typed by an RBS `type` alias.

```rbs
# core/builtin.rbs
type path = string | _ToPath

# stdlib/fileutils/0/fileutils.rbs
type pathlist = ::path | Array[::path]
def self?.ln_sf: (pathlist src, ::path dest, ?noop: boolish, ?verbose: boolish) -> void
```

```ruby
require 'fileutils'

class Repro
  # @param source [String]
  # @param dest [String]
  # @return [void]
  def link(source, dest)
    # Wrong argument type for FileUtils.ln_sf: src expected
    #   FileUtils::path, Array<FileUtils::path>, received String
    FileUtils.ln_sf(source, dest)
  end
end
```

`String` satisfies `string`, so `pathlist` accepts it; every stdlib and gem method with an alias-typed parameter is affected.

**Solution:** `type_to_tag` expands an `RBS::Types::Alias` to its underlying type, falling back to the nominal tag for recursive aliases and for generic ones like `type box[T] = Array[T] | nil`, whose args are not substituted into the expansion.

Alias names resolve against a core-aware copy of the environment, because `RbsMap` loads each library with `core_root: nil` and `resolve_type_names` rebinds a reference it cannot find rather than raising. That copy reparses RBS core, so it is built only for a library whose aliases show the rebinding: one of 41 here.

**Test plan:** `FileUtils.ln_sf('a', 'b')` reports `0 problems found` against real stdlib RBS; the spec suite's self-contained RBS does not cover that path.

Cost: on a cold cache build of this bundle, not distinguishable from master — 113.3s against 113.7s over four interleaved runs on Ruby 3.2.6 / rbs 4.1.2.

Fixes #1255.

### DIFF
--- a/lib/solargraph/rbs_map/conversions.rb
+++ b/lib/solargraph/rbs_map/conversions.rb
@@ -445,7 +445,6 @@ module Solargraph
       # @param decl [RBS::AST::Declarations::Constant]
       # @return [void]
       def constant_decl_to_pin decl
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls)
         pins.push create_constant(decl.name.relative!.to_s, tag, decl.comment&.string, decl)
       end
@@ -462,7 +461,6 @@ module Solargraph
           type_location: location_decl_to_pin_location(decl.location),
           source: :rbs
         )
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         rooted_tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted.rooted_tags
         pin.docstring.add_tag(YARD::Tags::Tag.new(:type, '', rooted_tag))
         pins.push pin
@@ -650,7 +648,6 @@ module Solargraph
           visibility: visibility,
           source: :rbs
         )
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         rooted_tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted.rooted_tags
         pin.docstring.add_tag(YARD::Tags::Tag.new(:return, '', rooted_tag))
         logger.debug do
@@ -682,13 +679,11 @@ module Solargraph
         pin.parameters <<
           Solargraph::Pin::Parameter.new(
             name: 'value',
-            # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
             return_type: RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted,
             source: :rbs,
             closure: pin,
             type_location: type_location
           )
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         rooted_tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted.rooted_tags
         pin.docstring.add_tag(YARD::Tags::Tag.new(:return, '', rooted_tag))
         pins.push pin
@@ -714,7 +709,6 @@ module Solargraph
           comments: decl.comment&.string,
           source: :rbs
         )
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         rooted_tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted.rooted_tags
         pin.docstring.add_tag(YARD::Tags::Tag.new(:type, '', rooted_tag))
         pins.push pin
@@ -732,7 +726,6 @@ module Solargraph
           type_location: location_decl_to_pin_location(decl.location),
           source: :rbs
         )
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         rooted_tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted.rooted_tags
         pin.docstring.add_tag(YARD::Tags::Tag.new(:type, '', rooted_tag))
         pins.push pin
@@ -750,7 +743,6 @@ module Solargraph
           type_location: location_decl_to_pin_location(decl.location),
           source: :rbs
         )
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         rooted_tag = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted.rooted_tags
         pin.docstring.add_tag(YARD::Tags::Tag.new(:type, '', rooted_tag))
         pins.push pin
@@ -831,7 +823,6 @@ module Solargraph
       # @return [ComplexType]
       # @param [Object] implicit_nil
       def extract_method_type_return_type type, implicit_nil
-        # @sg-ignore Wrong argument type for to_complex_type: type expected RBS::Types::Bases::Base, received union of concrete subtypes
         tag = RbsTranslator.to_complex_type(type.type.return_type, type_alias_decls: type_alias_decls)
         return ComplexType.parse("#{tag}, nil") if tag && implicit_nil
         tag
--- a/lib/solargraph/rbs_map/conversions.rb
+++ b/lib/solargraph/rbs_map/conversions.rb
@@ -95,7 +95,11 @@ module Solargraph
         # references into core won't expand in this case, but we
         # still get a nominal tag rather than failing to load the
         # library's pins at all.
-        logger.debug { "Could not build a core-aware environment for #{loader.libs}: #{e.message}" }
+        logger.warn do
+          "Could not build a core-aware environment for #{loader.libs}: #{e.message}. " \
+            'Type aliases referencing names declared in RBS core will not be expanded ' \
+            'for these libraries, and will be typed by their alias name instead.'
+        end
         fallback
       end
 
@@ -330,7 +334,6 @@ module Solargraph
         if decl.super_class
           type = build_type(decl.super_class.name, decl.super_class.args)
           generic_values = type.all_params.map(&:to_s)
-          decl.super_class.name.to_s
           pins.push Solargraph::Pin::Reference::Superclass.new(
             type_location: location_decl_to_pin_location(decl.super_class.location),
             closure: class_pin,
@@ -624,7 +627,6 @@ module Solargraph
       # @return [Array(Array<Pin::Parameter>, ComplexType)]
       def parts_of_function type, pin, implicit_nil
         [
-          # @sg-ignore Wrong argument type for to_parameter_pins: method_type expected RBS::MethodType, received RBS::MethodType, RBS::Types::Block
           RbsTranslator.to_parameter_pins(type, pin, pin.parameter_names, type_alias_decls: type_alias_decls),
           extract_method_type_return_type(type, implicit_nil).force_rooted
         ]
--- a/lib/solargraph/rbs_map/conversions.rb
+++ b/lib/solargraph/rbs_map/conversions.rb
@@ -127,12 +127,11 @@ module Solargraph
                                      # @sg-ignore flow sensitive typing should support case/when
                                      "Ignoring closure #{closure.inspect} on alias type name #{decl.name}")
           end
-          # @sg-ignore Unresolved calls to name, type, type_location
+          # @sg-ignore flow sensitive typing should support case/when
           alias_return_type = RbsTranslator.to_complex_type(decl.type, type_alias_decls: type_alias_decls).force_rooted
           pins.push(
-            # @sg-ignore Wrong argument type for Solargraph::Pin::Reference::TypeAlias.new: return_type expected Solargraph::ComplexType, received Solargraph::ComplexType::UniqueType, Solargraph::ComplexType
             Solargraph::Pin::Reference::TypeAlias.new(
-              # @sg-ignore Unresolved calls to name, type, type_location; return_type type mismatch
+              # @sg-ignore flow sensitive typing should support case/when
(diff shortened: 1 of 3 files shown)


apiology:

This fixes the single-level case (`dest: FileUtils::path`), but a nested case survives: `FileUtils::pathlist = path | Array[path]` is a union alias whose own member `path` is *itself* an alias. Expanding `pathlist` one level still leaves the inner `path` reference and the `Array` type argument unresolved.

Minimal repro, same shape as #1255's:

```ruby
require 'fileutils'

FileUtils.mkdir_p('a')
```

```
repro.rb:3: Wrong argument type for FileUtils.mkdir_p: list expected FileUtils::path, Array, received String
```

Verified against this PR's branch (commit `8b97bb1c`) merged into an integration branch, not just pre-fix `main` — the `dest`-style single alias case is fixed, this nested case still fails with a partially-expanded, partially-nominal message (`FileUtils::path, Array` — the alias name for `path` and no type argument for `Array`, versus #1255's original pre-fix message which at least showed `Array<FileUtils::path>` in full).

Given `type_to_tag`'s recursion is keyed on `RBS::Types::Alias` nodes specifically, I'd guess the union member `path` inside `pathlist`'s definition isn't being recursed into the same way the top-level `dest: path` reference is — worth confirming whether this PR's expansion walks into `RBS::Types::Union` members before recursing, or only handles the alias's immediate `type` definition. Happy to dig further if useful.

apiology:

Confirmed, and it goes further than the `pathlist` case above.

Root cause isn't a `Union`-member recursion gap in `type_to_tag` -- it's `RbsMap#loader` building every stdlib/gem environment with `core_root: nil` (to avoid re-declaring core, already loaded once via `RbsMap::CoreMap`). `FileUtils::path` is itself `type path = ::path`, referencing a type alias declared only in RBS core's `builtin.rbs`. With core absent from the environment, `RBS::Environment#resolve_type_names` doesn't fail to resolve `::path` -- it silently rebinds the reference back onto `FileUtils::path` itself. That's a genuine self-reference, so the recursion guard here correctly catches it and falls back to a nominal tag. But that means the single-alias case from #1255 was never actually fixed against the real stdlib RBS either -- only against this PR's own specs, which all define their alias entirely within one temp RBS file and never reference a name declared elsewhere:

```
$ bundle exec solargraph typecheck --level strong repro.rb   # FileUtils.ln_sf('a', 'b')
repro.rb:3: Wrong argument type for FileUtils.ln_sf: src expected FileUtils::path, Array, received String
```

Fixed in `fb7581bad` by resolving type alias names against a copy of the environment that includes RBS core, while pin generation still uses the core-less environment (avoids duplicating core's pins, and falls back to the core-less resolution on `RBS::DuplicatedDeclarationError` for the rare case where a loaded RBS shim redeclares a core name).

A/B on the real stdlib RBS (`StdlibMap.new('fileutils', rebuild: true)`, before → after `fb7581bad`):

| | before | after |
|---|---|---|
| `ln_sf` `src` | `FileUtils::path` | `String, _ToStr, _ToPath, Array<String, _ToStr, _ToPath>` |
| `ln_sf` `dest` | `FileUtils::path` | `String, _ToStr, _ToPath` |
| `mkdir_p` `list` | `FileUtils::path` | `String, _ToStr, _ToPath, Array<String, _ToStr, _ToPath>` |

`FileUtils.mkdir_p('a')` and `FileUtils.ln_sf('a', 'b')` both now typecheck clean. Added a spec in `spec/rbs_map/conversions_spec.rb` covering an alias that references a name declared only in RBS core.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -280,7 +330,7 @@ def class_decl_to_pin decl
         if decl.super_class
           type = build_type(decl.super_class.name, decl.super_class.args)
           generic_values = type.all_params.map(&:to_s)
-          superclass_name = decl.super_class.name.to_s
+          decl.super_class.name.to_s
```
apiology: What's the point of keeping this line?

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -573,7 +624,8 @@ def location_decl_to_pin_location(location)
       # @return [Array(Array<Pin::Parameter>, ComplexType)]
       def parts_of_function type, pin, implicit_nil
         [
-          RbsTranslator.to_parameter_pins(type, pin, pin.parameter_names),
+          # @sg-ignore Wrong argument type for to_parameter_pins: method_type expected RBS::MethodType, received RBS::MethodType, RBS::Types::Block
+          RbsTranslator.to_parameter_pins(type, pin, pin.parameter_names, type_alias_decls: type_alias_decls),
```
apiology, line 629: Shouldn't we fix the annotation instead of adding sg-ignores?

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -48,9 +53,52 @@ def load_environment_to_pins loader
...
+        RBS::Environment.from_loader(core_loader).resolve_type_names
+      rescue RBS::DuplicatedDeclarationError => e
+        # A declaration in `loader` (e.g. a local RBS shim) collides
+        # with something in RBS core itself. Fall back to the
+        # core-less resolution for type aliases; cross-namespace
+        # references into core won't expand in this case, but we
+        # still get a nominal tag rather than failing to load the
+        # library's pins at all.
+        logger.debug { "Could not build a core-aware environment for #{loader.libs}: #{e.message}" }
+        fallback
```
apiology, line 87: This seems like a bad situation to degrade our typechecking without revealing the situation to the user.

lib/solargraph/rbs_map/conversions.rb
apiology: Nothing ��� it was dead. The assignment it came from was removed earlier in this branch and the leftover expression's value was discarded. Deleted in 35a439d28.

lib/solargraph/rbs_map/conversions.rb
apiology: Fixed the annotation. `to_parameter_pins` only reads `method_type.type`, which both `RBS::MethodType` and `RBS::Types::Block` provide, so the declared `@param method_type [RBS::MethodType]` was too narrow. Widened to `[RBS::MethodType, RBS::Types::Block]`, which removes both `@sg-ignore` comments this branch added ��� the one here and the identical one at `rbs_translator.rb:106`; neither is on master.

One suppression remains at the block call site, for a different error the widening exposed: `received RBS::Types::Block, nil`, because the `if method_type.block` guard does not narrow nil out of the second `method_type.block` call. That is castwide/solargraph#1249 rather than an annotation problem, so it is tagged with that link instead of being worked around. Local `typecheck --level strong` is unchanged at 494 problems, and CI's is unchanged at 490.

lib/solargraph/rbs_map/conversions.rb
apiology: Agreed. The fallback now logs at `warn` rather than `debug`, naming the libraries involved and what the degradation costs: type aliases referencing names declared in RBS core will not be expanded for them and will be typed by their alias name instead.

Not `Solargraph.assert_or_log`: a local RBS shim redeclaring a core name is a legitimate configuration ��� this repo ships `rbs/shims` entries itself ��� so raising under assertions would break valid setups rather than reveal a problem.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -75,11 +127,13 @@ def convert_decl_to_pin decl, closure
                                      # @sg-ignore flow sensitive typing should support case/when
                                      "Ignoring closure #{closure.inspect} on alias type name #{decl.name}")
           end
+          # @sg-ignore Unresolved calls to name, type, type_location
```
apiology: I think 'flow sensitive typing should support case/when' would cover this case

lib/solargraph/rbs_map/conversions.rb
apiology: Please apply this fix throughout for similar issues.  If solargraph limitations result in a bad typecheck, name the limitation in the sg-ignore; known limitations get documented in rules.rb with counts.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -37,6 +37,11 @@ def initialize loader:
```
apiology, line 37: Is this PR's work redundant with the typedef PR?  I should be more familiar with it than I am.

lib/solargraph/rbs_map/conversions.rb
apiology: **Claude:** Confirmed - retagged in 41049803a. `decl` is still typed as the case discriminant `RBS::AST::Declarations::Base` inside the `TypeAlias` branch, same as the sibling branches above it; stripping the old ignores showed the literal error is "Unresolved call to type/name/location on RBS::AST::Declarations::Base", not the generic wording that was there before.

lib/solargraph/rbs_map/conversions.rb
apiology: **Claude:** Applied throughout in 05da6c6c2 and 41049803a. `to_complex_type`/`type_to_tag`'s `type` param was declared `RBS::Types::Bases::Base`, which only covers the Bases::* variants (Bool, Void, Any, etc.) - every dispatch on a concrete node (Optional, Union, Tuple...) needed an `@sg-ignore` to work around the mismatch. Widened both to `RBS::Types::t`, RBS's own alias for the full type-node union, which cleared 13 ignores across `rbs_translator.rb` and `conversions.rb`. Two call sites inside `type_to_tag`'s own `case/when` needed a new "flow sensitive typing should support case/when" ignore, matching the existing convention - Solargraph does not narrow types across case/when branches. Full-project strong typecheck: 494 -> 491 problems, no regressions.

lib/solargraph/rbs_map/conversions.rb
apiology: **Claude:** Not redundant - castwide/solargraph#1244 ("Introduce Type base class for ComplexType/UniqueType hierarchy") is an internal refactor of the ComplexType/UniqueType class hierarchy (replacing the TypeMethods mixin with a shared Type base class). It touches no files this PR touches; this PR is about expanding RBS type aliases before conformance checks. Different, non-overlapping concerns.

lib/solargraph/rbs_map/conversions.rb
```diff
@@ -730,7 +783,7 @@ def prepend_to_pin decl, closure
       # @return [void]
       def extend_to_pin decl, closure
         type = build_type(decl.name, decl.args)
-        generic_values = type.all_params.map(&:rooted_tags)
+        type.all_params.map(&:rooted_tags)
```
apiology: This is a bug; we ought to be sending this into Extend.new (similar with Prepend above.

If this is triggering rubocop and was auto-corrected, I'd prefer to leave it erroring; fixing it is out of scope for this PR, but doing this gets rid of the todo file entry that tracks this as an issue to reproduce and fix.

apiology: **Claude:** Restored in a91e98ab4. Both `prepend_to_pin` and `extend_to_pin` are back to master's form, so `Lint/UselessAssignment` keeps firing on them and the todo entry survives. Passing the values into `Prepend.new` / `Extend.new` stays out of scope here.

PR #1285, opened 2026-08-11
ensure clause breaks return type inference

### LINKED ISSUE #1284: A method with any ensure clause has its return type never inferred, regardless of the method body
🤖 Filed by Claude, not the account owner — acting on their behalf via their GitHub credentials.

Any method that has an `ensure` clause — even an empty one — fails `solargraph typecheck --level strong` with `return type could not be inferred`, regardless of the declared `@return` tag or what the method body actually does. This is unrelated to generics, `yield`, or block parameters.

### Minimal reproduction

```ruby
class Repro
  # @return [String]
  def call
    'hello'
  ensure
    nil
  end
end
```

```
repro.rb:3: Repro#call return type could not be inferred
```

Removing the `ensure` clause (identical method otherwise) — **0 problems**.

### Isolation

- Declared return type doesn't matter — tried `String`, `Object`, `void`; all fail identically with `ensure` present.
- Body content doesn't matter — a literal (`'hello'`), a local variable, a method call, all fail identically.
- Content of the `ensure` block doesn't matter — `nil`, an assignment, an empty block, all trigger it.

### Environment

Reproduced on plain upstream `solargraph 0.60.3` (installed via `gem install solargraph`, no project config, no Gemfile in scope — confirmed with `BUNDLE_GEMFILE`/`GEM_HOME`/`GEM_PATH` explicitly unset to rule out any local fork/patch contamination).

apiology:
**Problem:** Any method with an `ensure` clause fails `solargraph typecheck --level strong` with `return type could not be inferred`, whatever the body returns and whatever `@return` declares.

```ruby
class Repro
  # @return [String]
  def call
    'hello'
  ensure
    nil
  end
end
```

```
repro.rb:3: Repro#call return type could not be inferred
```

An empty `ensure` is enough to trigger it, so every method using one for cleanup is unusable at `strict` or `strong`. Fixes #1284.

**Solution:** Unwrap an `:ensure` node to its body's value, scanning the ensure clause separately for explicit `return`s, since only those reach the caller.

`reduce_to_value_nodes` also gains the `:rescue` case it never had, which is what a rescue nested inside an ensure body needs in order to resolve.

https://claude.ai/code/session_01XGDWQqVqW2ZdP2rbfr5DDs

### DIFF
--- a/lib/solargraph/parser/parser_gem/node_methods.rb
+++ b/lib/solargraph/parser/parser_gem/node_methods.rb
@@ -442,6 +442,11 @@ module Solargraph
             CONDITIONAL_ALL_BUT_FIRST = %i[if unless].freeze
             ONLY_ONE_CHILD = [:return].freeze
             FIRST_TWO_CHILDREN = [:rescue].freeze
+            # :ensure's value is always its body's value (first
+            # child); the ensure clause itself (second child) never
+            # contributes to the method's return value unless it
+            # explicitly returns.
+            ENSURE = [:ensure].freeze
             COMPOUND_STATEMENTS = %i[begin kwbegin].freeze
             SKIPPABLE = %i[def defs class sclass module].freeze
             FUNCTION_VALUE = [:block].freeze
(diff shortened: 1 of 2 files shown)


lib/solargraph/parser/parser_gem/node_methods.rb
```diff
@@ -488,6 +493,12 @@ def from_value_position_statement node, include_explicit_returns: true
                 result.concat reduce_to_value_nodes([node.children[0]])
               elsif FIRST_TWO_CHILDREN.include?(node.type)
                 result.concat reduce_to_value_nodes([node.children[0], node.children[1]])
+              elsif ENSURE.include?(node.type)
+                result.concat reduce_to_value_nodes([node.children[0]])
+                if include_explicit_returns
+                  # @sg-ignore Need to add nil check here
```
apiology: So add it

Done at add70853 — the branch now binds `ensure_body = node.children[1]` and
guards with `ensure_body.is_a?(::Parser::AST::Node)` before recursing, so the
nil case no longer reaches `explicit_return_values_from_compound_statement`.

The `@sg-ignore` still on that line covers a different gap: flow-sensitive
typing does not narrow the type from the `is_a?` guard.

PR #1288, opened 2026-08-12
Fix generic binding and block signature merging through a @!parse stub

### LINKED ISSUE #1286: Generic T doesn't bind through a cross-file @!parse stub
🤖 Filed by Claude, not the account owner — acting on their behalf via their GitHub credentials.

### Summary

When a class's real implementation lives in one file (e.g. a gem) and a `@!parse`-based YARD annotation override in a separate file (e.g. `config/annotations_*.rb`) declares that class as `@generic T`, calling a generic instance method through a stub that returns a parameterized instantiation of that class (`Collection<Widget>`) never binds `T` to the parameter — a subsequent call to a method that exists only on `Widget` (not `Object`) fails to resolve, and the enclosing method's return type inference fails.

### Reproduction

**`lib/widgetbox.rb`** (a real gem, no annotations of its own):

```ruby
module Widgetbox
  class Collection
    def self.make
      new
    end

    def last
      nil
    end
  end

  class Widget
    # @return [String, nil]
    def resource_subtype; end
  end
end
```

**`config/annotations.rb`** (application-level `@!parse` override):

```ruby
# @!parse
#   module Widgetbox
#     # @generic T
#     class Collection
#       class << self
#         # @return [Widgetbox::Collection<Widgetbox::Widget>]
#         def make; end
#       end
#       # @return [generic<T>]
#       def last; end
#     end
#   end
```

**`app.rb`**:

```ruby
# @return [String, nil]
def check
  Widgetbox::Collection.make.last.resource_subtype
end
```

```
$ solargraph typecheck --level strong app.rb
app.rb:2: #check return type could not be inferred
```

`Widgetbox::Collection.make` is declared to return `Widgetbox::Collection<Widgetbox::Widget>`. `#last`'s declared return is `generic<T>`, which should resolve to `Widget` — but `.resource_subtype` (a method that exists only on `Widget`, not `Object`) can't be resolved on the result, and the whole chain's return type is never inferred.

`bundle exec solargraph pin 'Widgetbox::Collection#last'` shows the merged pin, with `T` still unbound:

```
def last: () -> (T | nil), context=::Widgetbox::Collection ... via :combined
```

### Environment

Reproduced against `apiology/solargraph@b6eea57c94999ec3b22fa659e6791836a8cc4357` (`2026-08-04` branch). Not yet checked against plain upstream `castwide/solargraph`, since the `@generic`/`@!parse` combination this depends on may itself only exist on the fork's more recent generics work.

apiology:
Fixes https://github.com/castwide/solargraph/issues/1286

**Problem:**

A `@generic` tag that a `@!parse` stub adds to a class already defined elsewhere doesn't apply, so the type variable never binds and every call downstream of it stops resolving.

```ruby
# widgetbox.rb - the gem's own class, no generics
module Widgetbox
  class Collection
    def self.make; new; end
    def last; nil; end
  end

  class Widget
    # @return [String, nil]
    def resource_subtype; end
  end
end

# annotations.rb - a @!parse stub adds @generic T and the return types
# @!parse
#   module Widgetbox
#     # @generic T
#     class Collection
#       class << self
#         # @return [Widgetbox::Collection<Widgetbox::Widget>]
#         def make; end
#       end
#       # @return [generic<T>]
#       def last; end
#     end
#   end

# app.rb
# @return [String, nil]
def check
  Widgetbox::Collection.make.last.resource_subtype
end
```

```
$ solargraph typecheck --level strong app.rb
app.rb:2: #check return type could not be inferred
```

It happens whenever the plain definition is mapped first — a gem loading ahead of the workspace, or simply appearing earlier in the same file — so an annotation written this way is inert with nothing reported to say so.

**Solution:**

Choose the namespace pin that actually declares generics rather than whichever loaded first, and combine same-path method pins so a stub's `@return` tags survive alongside the gem's own definition.

---

**Problem:**

A workspace `@!parse` stub's block parameter type has no effect when the gem's own doc already declares a return type:

```ruby
module Widgetbox
  class << self
    # @return [String]
    def build(&block) = 'x'
  end
end

# @!parse
#   module Widgetbox
#     class << self
#       # @yieldparam config [String]
#       # @return [String]
#       def build(&block); end
#     end
#   end

Widgetbox.build { |config| config.upcase } # Unresolved call to upcase
```

The method pin itself carries only one of the two signatures, so hover and completion lose the same information inference does.

**Solution:**

`combine_signatures_by_type_arity` buckets signatures by `type_arity` before merging, and a block's declared-parameter count changes that bucket, so this adds a pass that merges block-informativeness-only differences first and makes `Callable#combine_blocks` prefer the more informative block instead of choosing arbitrarily; two related gaps stay open (a stub missing `@return` is discarded wholesale by `combine_signatures`, and `def self.build` paired with `class << self` never combines pins at all).

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -814,7 +814,7 @@ module Solargraph
 
     # The namespace pin for fqns that actually declares its generics,
     # picked from any duplicate pins for the same namespace.
-    #
+    # @todo Consider trade-offs of merging namespace pins instead of trying to choose the best one for this use
     # @param fqns [String]
     # @return [Pin::Namespace, nil]
     def namespace_pin_for_generics fqns
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -818,8 +818,8 @@ module Solargraph
     # @param fqns [String]
     # @return [Pin::Namespace, nil]
     def namespace_pin_for_generics fqns
-      # @type [Array<Pin::Namespace>]
       candidates = store.get_path_pins(fqns).select { |p| p.is_a?(Pin::Namespace) }
+      # @sg-ignore select with an is_a? block does not narrow the element type
       candidates.find { |p| !p.generics.empty? } || candidates.first
     end
 
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -812,12 +812,8 @@ module Solargraph
       @store ||= Store.new
     end
 
-    # Get the namespace pin that should be used to resolve generic type
-    # parameters for a fully qualified namespace. Multiple pins can
-    # exist for the same namespace (e.g., a gem's own class definition
-    # plus a `@!parse` stub in a different file that adds `@generic`
-    # tags); the one that actually declares the generics must be used,
-    # regardless of load order.
+    # The namespace pin for fqns that actually declares its generics,
+    # picked from any duplicate pins for the same namespace.
     #
     # @param fqns [String]
     # @return [Pin::Namespace, nil]
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -452,7 +452,7 @@ module Solargraph
     def get_methods rooted_tag, scope: :instance, visibility: [:public], deep: true
       rooted_type = ComplexType.try_parse(rooted_tag)
       fqns = rooted_type.namespace
-      namespace_pin = store.get_path_pins(fqns).select { |p| p.is_a?(Pin::Namespace) }.first
+      namespace_pin = namespace_pin_for_generics(fqns)
       cached = cache.get_methods(rooted_tag, scope, visibility, deep)
       return cached.clone unless cached.nil?
       # @type [Array<Solargraph::Pin::Method>]
@@ -783,9 +783,10 @@ module Solargraph
       # @todo Can inner_get_methods be cached?  Lots of lookups of base types going on.
       methods = inner_get_methods(resolved_reference_type.tag, scope, visibility, deep, skip, no_core)
       if namespace_pin && !resolved_reference_type.all_params.empty?
-        reference_pin = store.get_path_pins(resolved_reference_type.name).select { |p| p.is_a?(Pin::Namespace) }.first
+        reference_pin = namespace_pin_for_generics(resolved_reference_type.name)
         # logger.debug { "ApiMap#add_methods_from_reference(type=#{type}) - resolving generics with #{reference_pin.generics}, #{resolved_reference_type.rooted_tags}" }
         methods = methods.map do |method_pin|
+          # @sg-ignore Need to add nil check here
           method_pin.resolve_generics(reference_pin, resolved_reference_type)
         end
       end
@@ -811,6 +812,21 @@ module Solargraph
       @store ||= Store.new
     end
 
+    # Get the namespace pin that should be used to resolve generic type
+    # parameters for a fully qualified namespace. Multiple pins can
+    # exist for the same namespace (e.g., a gem's own class definition
+    # plus a `@!parse` stub in a different file that adds `@generic`
+    # tags); the one that actually declares the generics must be used,
+    # regardless of load order.
+    #
+    # @param fqns [String]
+    # @return [Pin::Namespace, nil]
+    def namespace_pin_for_generics fqns
+      # @type [Array<Pin::Namespace>]
+      candidates = store.get_path_pins(fqns).select { |p| p.is_a?(Pin::Namespace) }
+      candidates.find { |p| !p.generics.empty? } || candidates.first
+    end
+
     # @return [Solargraph::ApiMap::Cache]
     attr_reader :cache
 
@@ -826,7 +842,7 @@ module Solargraph
       rooted_type = ComplexType.parse(rooted_tag).force_rooted
       fqns = rooted_type.namespace
       rooted_type.all_params
-      namespace_pin = store.get_path_pins(fqns).select { |p| p.is_a?(Pin::Namespace) }.first
+      namespace_pin = namespace_pin_for_generics(fqns)
       return [] if no_core && fqns =~ /^(Object|BasicObject|Class|Module)$/
       reqstr = "#{fqns}|#{scope}|#{visibility.sort}|#{deep}"
       return [] if skip.include?(reqstr)
@@ -867,6 +883,7 @@ module Solargraph
           end
           rooted_sc_tag = qualify_superclass(rooted_tag)
           unless rooted_sc_tag.nil?
+            # @sg-ignore Need to add nil check here
             result.concat inner_get_methods_from_reference(rooted_sc_tag, namespace_pin, rooted_type, scope,
                                                            visibility, true, skip, no_core)
           end
@@ -880,6 +897,7 @@ module Solargraph
           end
           rooted_sc_tag = qualify_superclass(rooted_tag)
           unless rooted_sc_tag.nil?
+            # @sg-ignore Need to add nil check here
             result.concat inner_get_methods_from_reference(rooted_sc_tag, namespace_pin, rooted_type, scope,
                                                            visibility, true, skip, true)
           end
--- a/lib/solargraph/api_map/store.rb
+++ b/lib/solargraph/api_map/store.rb
@@ -297,16 +297,12 @@ module Solargraph
         @index ||= Index.new
       end
 
-      # A method can be defined by more than one pin with the same
-      # path - e.g., a gem's own implementation plus a `@!parse` stub
-      # in a separate file that overrides its documentation. Combine
-      # them into a single pin so callers see one consistent signature
-      # instead of an arbitrary pick among duplicates.
-      #
-      # Aliases are skipped: combining a MethodAlias pin with a
-      # non-alias pin at the same path produces a `:combined` pin that
-      # #resolve_method_alias can't trace back to its original target,
-      # which raises under SOLARGRAPH_ASSERTS=on.
+      # Combines pins that share a method path (e.g. a gem's own
+      # implementation plus a `@!parse` stub overriding its docs)
+      # into one pin, so callers see a single consistent signature.
+      # Aliases are skipped: combining a MethodAlias with a
+      # non-alias pin breaks #resolve_method_alias under
+      # SOLARGRAPH_ASSERTS=on.
       #
       # @param pins [Array<Pin::Method>]
       # @return [Array<Pin::Method>]
--- a/lib/solargraph/api_map/store.rb
+++ b/lib/solargraph/api_map/store.rb
@@ -297,10 +297,9 @@ module Solargraph
         @index ||= Index.new
       end
 
-      # Combines same-path pins into one. Skips aliases (breaks
-      # #resolve_method_alias under SOLARGRAPH_ASSERTS=on) and
-      # DelegatedMethod (initialize needs exactly one of
-      # :method/:receiver; a merged pin can't hold both).
+      # Combines same-path pins into one. Skips aliases (a merged pin
+      # breaks #resolve_method_alias under SOLARGRAPH_ASSERTS=on) and
+      # DelegatedMethod (initialize needs exactly one of :method/:receiver).
       #
       # @param pins [Array<Pin::Method>]
       # @return [Array<Pin::Method>]
--- a/lib/solargraph/api_map/store.rb
+++ b/lib/solargraph/api_map/store.rb
@@ -304,12 +304,20 @@ module Solargraph
       # non-alias pin breaks #resolve_method_alias under
       # SOLARGRAPH_ASSERTS=on.
       #
+      # DelegatedMethod pins are skipped for the same reason, but they
+      # raise outright rather than degrade: Pin::Base#combine_with rebuilds
+      # the merged pin with self.class.new(**new_attrs) carrying only
+      # generic pin attributes, and DelegatedMethod#initialize requires
+      # exactly one of :method / :receiver. Each pin's delegation target is
+      # per-pin state that a merged pin has no way to represent, so both
+      # are kept.
+      #
       # @param pins [Array<Pin::Method>]
       # @return [Array<Pin::Method>]
       def combine_duplicate_method_pins pins
         result = []
         pins.group_by(&:path).each_value do |group|
-          if group.length == 1 || group.any? { |pin| pin.is_a?(Pin::MethodAlias) }
+          if group.length == 1 || group.any? { |pin| pin.is_a?(Pin::MethodAlias) || pin.is_a?(Pin::DelegatedMethod) }
             result.concat(group)
           else
             # @sg-ignore group is never empty here (group_by never yields an empty group)
--- a/lib/solargraph/api_map/store.rb
+++ b/lib/solargraph/api_map/store.rb
@@ -297,11 +297,8 @@ module Solargraph
         @index ||= Index.new
       end
 
-      # Combines pins that share a method path (e.g. a gem's own
-      # implementation plus a `@!parse` stub overriding its docs)
-      # into one pin, so callers see a single consistent signature.
-      # Aliases are skipped: combining a MethodAlias with a
-      # non-alias pin breaks #resolve_method_alias under
+      # Combines same-path pins into one pin. Skips aliases - merging
+      # a MethodAlias breaks #resolve_method_alias under
       # SOLARGRAPH_ASSERTS=on.
       #
       # @param pins [Array<Pin::Method>]
--- a/lib/solargraph/api_map/store.rb
+++ b/lib/solargraph/api_map/store.rb
@@ -72,11 +72,12 @@ module Solargraph
       # @param fqns [String]
       # @param scope [Symbol]
       # @param visibility [Array<Symbol>]
-      # @return [Enumerable<Solargraph::Pin::Method>]
+      # @return [Array<Solargraph::Pin::Method>]
       def get_methods fqns, scope: :instance, visibility: [:public]
-        namespace_children(fqns).select do |pin|
+        pins = namespace_children(fqns).select do |pin|
           pin.is_a?(Pin::Method) && pin.scope == scope && visibility.include?(pin.visibility)
         end
+        combine_duplicate_method_pins(pins)
       end
 
       BOOLEAN_SUPERCLASS_PIN = Pin::Reference::Superclass.new(name: 'Boolean', closure: Pin::ROOT_PIN,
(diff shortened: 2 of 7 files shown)


apiology:
🤖 Filed by Claude, not the account owner — acting on their behalf via their GitHub credentials.

## Why this fix combines pins

**Why combine at all:** a class defined in a gem and reopened via `@!parse` to add `@generic` tags produces two pins for the same path. Without combining, `ApiMap` picks whichever loaded first - a gem's own pins load before workspace pins, so the `@generic` tag and any overridden return type were silently ignored. `Store#get_methods` now merges same-path method pins via `Pin::Base#combine_with`, skipping `Pin::MethodAlias` pins (combining one with a regular pin at the same path produced an unresolvable `:combined` alias, raising under `SOLARGRAPH_ASSERTS=on`).

**Why this way and not more broadly:** I first tried combining at `ApiMap#catalog` time, across `@doc_map.pins + iced_pins`, via a generalized `GemPins.combine_method_pins`. That regressed `Integer#+`'s signature count - `Pin::Method#==` never compares `signatures`, so `combine_method_pins`'s "skip if already identical" shortcut silently dropped a pin instead of merging it. This is exactly why https://github.com/castwide/solargraph/pull/1195 scoped combination down to doc-maps only. `Store#get_methods` sidesteps the bug structurally: it calls `combine_with` directly and never goes through `==`.

apiology:
**Claude:** This introduces a crash (`NoMethodError: undefined method 'gates' for nil`) on a plain YARD duck-type parameter — no `@!parse` stub or generics involved.

**Minimal repro** (4 lines):

```ruby
# @param x [#to_s]
# @return [void]
def go(x)
  x.to_s
end
```

`bundle exec solargraph typecheck --level strong` against this file crashes:

```
[WARN] Error checking call to to_s at #<Solargraph::Range:...>: [NoMethodError] undefined method `gates' for nil
lib/solargraph/pin/method.rb:298:in `typify': undefined method `gates' for nil (NoMethodError)

          qualified = type.qualify(api_map, *closure.gates)
                                                    ^^^^^^
	from lib/solargraph/source/chain/call.rb:131:in `match_overload_type'
	from lib/solargraph/source/chain/call.rb:362:in `block (2 levels) in inferred_pins'
	...
	from lib/solargraph/pin/method.rb:691:in `block in infer_from_return_nodes'
	from lib/solargraph/type_checker.rb:173:in `method_return_type_problems_for'
```

apiology:
**Claude:** Fixed in 2341ccc9f — `Pin::DuckMethod` (built for `#method_name` duck-type tags) is constructed with no `closure`, and `Method#typify` called `closure.gates` unconditionally once it resolved a type via `see_reference`/`typify_from_super`. Guarded it with `closure&.gates || ['']` (same pattern used elsewhere in the file) and added a regression spec reproducing the crash directly on a closure-less `DuckMethod` pin.

apiology:
**Claude:** Second, separate false positive from this same PR — a class that reopens a core type (e.g. `class String; def foo; ...; end; end`) to add a new method, then calls a genuinely core method (`gsub`) later, gets a false "Unresolved call" if the reopened-method result feeds into an earlier statement and at least one more statement follows.

Bisected the same way as the `to_s`/`puts` crash: clean on `71f16f832d6ee1790c3db0f868ff0335502438da` (pre-#1288), broken on both `cd567a0ad8c77934a6be5b8451d9b16ae7e5bfb8` (post-#1288) and the current `2026-08-04` tip (`a6942fc53231d8fe4f1e16599774a27f86af8574`, which already includes 2341ccc9 fixing the first issue) — so this is a second symptom of #1288 that the `closure&.gates` fix didn't cover.

**Minimal repro** (13 lines):

```ruby
class String
  # @return [String]
  def depunctuate
    self
  end
end

# @param str [String]
# @param other [String]
# @return [String]
def go(str, other)
  str = str.gsub(other.depunctuate, other)
  str = str.gsub(other, other)
  str.gsub(other, other)
end
```

`bundle exec solargraph typecheck --level strong`:

```
repro.rb:13: Unresolved call to gsub on String
```

Line 13 is the *second* `gsub` call — not the one using the reopened method (line 11), and not the last one (line 14). Reduced by cutting one axis at a time:
- Dropping to 2 statements (either order) clears the error.
- Moving the `.depunctuate` call to the *last* statement clears the error — it only reproduces when it's in a non-final position.
- Replacing `.depunctuate` with a call to a real, unmodified `String` method removes the reopened-class angle and the error disappears — the reopened method's own presence in the pinset for `String` is load-bearing, not just "any 3-statement gsub chain."

Not yet traced into which pin gets combined/miscombined here, but the "reopen a core class, call a real core method afterward" shape feels adjacent to `Store#combine_duplicate_method_pins`' same-path pin merging from this PR, same as the first issue.

apiology:
**Claude:** Root cause for this one: `Pin::Base#==` compares `#location` but not `#presence`. `combine_with` results choose the earliest assignment's `#location`, so two combined pins covering a different number of assignments to the same variable can share `#location` while covering different `#presence` ranges. `Chain`'s inference recursion guard keys off of `#==` (via `Array#include?`), so it treated two distinct combined pins as the same one and dropped the second — leaving its receiver's type undefined for the middle statement.

Not actually reachable on this branch or on `2026-08-04` alone in isolation — it needs the combination of two other in-flight branches, #1258 and #1282, both of which add fields (`narrowed_return_type`/`definite`) to the same code path. Verified neither reproduces it alone; only the two merged together (as on `2026-08-04`) does.

Fix + regression specs (including this 13-line repro, kept as a standing guard since the failure mode depends on merge order, not on either branch individually): https://github.com/castwide/solargraph/pull/1293

lib/solargraph/api_map/store.rb
```diff
@@ -296,6 +297,33 @@ def index
...
+      # A method can be defined by more than one pin with the same
+      # path - e.g., a gem's own implementation plus a `@!parse` stub
+      # in a separate file that overrides its documentation. Combine
+      # them into a single pin so callers see one consistent signature
+      # instead of an arbitrary pick among duplicates.
+      #
+      # Aliases are skipped: combining a MethodAlias pin with a
+      # non-alias pin at the same path produces a `:combined` pin that
+      # #resolve_method_alias can't trace back to its original target,
+      # which raises under SOLARGRAPH_ASSERTS=on.
```
apiology: 6 lines at most

lib/solargraph/api_map.rb
```diff
@@ -811,6 +812,21 @@ def store
       @store ||= Store.new
     end
 
+    # Get the namespace pin that should be used to resolve generic type
+    # parameters for a fully qualified namespace. Multiple pins can
+    # exist for the same namespace (e.g., a gem's own class definition
+    # plus a `@!parse` stub in a different file that adds `@generic`
+    # tags); the one that actually declares the generics must be used,
+    # regardless of load order.
```
apiology: 2 lines at most

spec/api_map/store_spec.rb
```diff
@@ -49,6 +51,96 @@
...
+      pins = store.get_methods('Foo', scope: :instance).select { |p| p.name == 'bar' }
+      expect(pins.length).to eq(1)
+      expect(pins.first.return_type.tag).to eq('String')
+    end
+
+    it 'does not combine a method alias with a regular method sharing its path' do
+      # Combining an alias pin with a non-alias pin at the same path
+      # previously produced a `:combined` pin that couldn't be
+      # resolved back to its original target, raising under
+      # SOLARGRAPH_ASSERTS=on.
```
apiology: 2 lines at most

spec/api_map/store_spec.rb
```diff
@@ -49,6 +51,96 @@
...
+      # gap - it groups by union member *count*, not the actual
+      # types), so they hit the ">10" bail-out before ever reaching
+      # the code path #1238 fixed. #1238's own regression spec
+      # (spec/pin/method_spec.rb "combines many non-mergeable
+      # same-type-arity signatures without exponential blowup") is
+      # the precise guard for that bug, exercising
+      # combine_same_type_arity_signatures directly with signatures
+      # engineered to never merge. This spec instead just confirms
+      # Store#get_methods' combination of many real same-path pins,
+      # as this PR adds, completes quickly rather than hanging.
```
apiology: WHOA.  Holy changelogging batman.  2 lines at most.

spec/source_map/clip_spec.rb
```diff
@@ -1917,6 +1917,54 @@ def bad_passthrough; yield; end
     expect(type.to_s).to eq('undefined')
   end
 
+  it 'binds generics through a cross-file @!parse stub that adds @generic to an existing class' do
+    # The plain implementation (e.g., as it would be defined in a gem) is
+    # unaware of any generics. It's mapped first, simulating a gem's pins
+    # being loaded before the workspace's.
```
apiology: be more specific than "plain"

lib/solargraph/api_map/store.rb
apiology: **Claude:**

Trimmed in 538cd79fd. It still ran to seven lines counting the tags, so
88f5693b9 folds the fourth prose line into the third:

```ruby
      # Combines same-path pins into one. Skips aliases (a merged pin
      # breaks #resolve_method_alias under SOLARGRAPH_ASSERTS=on) and
      # DelegatedMethod (initialize needs exactly one of :method/:receiver).
      #
      # @param pins [Array<Pin::Method>]
      # @return [Array<Pin::Method>]
```

lib/solargraph/api_map.rb

Trimmed to two lines in 538cd79fd:

```ruby
    # The namespace pin for fqns that actually declares its generics,
    # picked from any duplicate pins for the same namespace.
```

spec/api_map/store_spec.rb
apiology: **Claude:**

Trimmed to two lines in 538cd79fd:

```ruby
      # A combined alias pin can't be traced back to its original
      # target, which #resolve_method_alias needs to work.
```

spec/api_map/store_spec.rb
apiology: **Claude:**

The whole 22-line block went in 538cd79fd; the history it recounted lives
in that commit message instead. What remains states only the constraint
still in force:

```ruby
      # The regression is combinatorial blowup, not a specific merge
      # outcome, so bound the size rather than assert exact merges.
```

Reworded again in 88f5693b9 - the intervening version used "pin" as a
verb, which collides with Solargraph::Pin.

spec/source_map/clip_spec.rb
apiology: **Claude:**

538cd79fd rewrote the prose to spell out what plain meant, but the word
survived in the variable name. 88f5693b9 renames it so the name carries
that and the comment shrinks:

```ruby
    # gem_source has no `@generic` tag or `@!parse` stub, and is mapped
    # before the workspace's stub.
    gem_source = Solargraph::SourceMap.load_string(%(
```

lib/solargraph/api_map/store.rb
```diff
@@ -296,6 +297,26 @@ def index
         @index ||= Index.new
       end
 
+      # Combines same-path pins into one. Skips aliases (a merged pin
+      # breaks #resolve_method_alias under SOLARGRAPH_ASSERTS=on) and
+      # DelegatedMethod (initialize needs exactly one of :method/:receiver).
```
apiology: Please update comment to list situations where we need to do this.

What classes currently call into .combine_with from outside the Pin type hierarchy?

spec/api_map/store_spec.rb
```diff
@@ -49,6 +51,87 @@
...
+            # @return [Type#{i}]
+            def bar(other); end
+          end
+        ), "source#{i}.rb")
+      end
+      store = nil
+      Timeout.timeout(5) { store = described_class.new(maps.flat_map(&:pins)) }
+
+      result = nil
+      Timeout.timeout(5) { result = store.get_methods('Foo', scope: :instance) }
```
apiology: Were the timeouts here due to a particular issue found?

spec/pin/method_spec.rb
```diff
@@ -786,4 +786,10 @@ def foo(**bar); end
       expect { pin.signatures }.not_to raise_error
     end
   end
+
+  it 'typifies a DuckMethod pin with no closure without raising' do
+    api_map = Solargraph::ApiMap.new
+    pin = Solargraph::Pin::DuckMethod.new(name: 'to_s', source: :api_map)
+    expect { pin.typify(api_map) }.not_to raise_error
+  end
```
apiology: Can we add additional better expectations?  This seems a low bar.

lib/solargraph/api_map.rb
```diff
@@ -811,6 +812,17 @@ def store
       @store ||= Store.new
     end
 
+    # The namespace pin for fqns that actually declares its generics,
+    # picked from any duplicate pins for the same namespace.
+    #
+    # @param fqns [String]
+    # @return [Pin::Namespace, nil]
+    def namespace_pin_for_generics fqns
+      # @type [Array<Pin::Namespace>]
```
apiology: Don't add a `@type` here - may be hiding an inference defect we want to know about

lib/solargraph/api_map/store.rb
apiology: **Claude:**

Named in d1b325d54:

```ruby
      # Combines same-path pins into one. They arise when a method is
      # documented in more than one file - a `@!parse` stub re-documenting
      # a method the gem already defines, or a reopened class. Aliases and
      # DelegatedMethod are skipped; neither survives a merge.
```

The alias and DelegatedMethod reasoning that used to sit here is stated by
the two specs covering those cases, so it is not repeated.

Three callers of `#combine_with` sit outside the Pin hierarchy:

- [`ApiMap#var_at_location`](https://github.com/apiology/solargraph/blob/d1b325d54/lib/solargraph/api_map.rb#L419) -
  `vars_at_location.inject(&:combine_with)`, merging the variable pins
  visible at one location.
- [`GemPins.combine_method_pins`](https://github.com/apiology/solargraph/blob/d1b325d54/lib/solargraph/gem_pins.rb#L29) -
  merging a gem method's RBS and YARD pins.
- [`ApiMap::Store#combine_duplicate_method_pins`](https://github.com/apiology/solargraph/blob/d1b325d54/lib/solargraph/api_map/store.rb#L313) -
  this PR.

Inside the hierarchy: `Pin::Method#combine_same_type_arity_signatures`,
`Pin::LocalVariable#combine_with`, `Pin::Callable#combine_with`.

Only the first of those three situations is what this method handles - the
RBS/YARD merge is GemPins' job, so it is not listed in the comment.

spec/api_map/store_spec.rb
apiology: **Claude:**

Yes - the exponential blowup in
`Pin::Method#combine_same_type_arity_signatures`, fixed in #1238. It first
presented as an apparent infinite loop and was stubbed out in #1186 before
being diagnosed.

This spec does not reach that code path, though. Its 30 signatures each have
one parameter of one type, so they land in a single type-arity bucket and
[method.rb:513](https://github.com/apiology/solargraph/blob/d1b325d54/lib/solargraph/pin/method.rb#L513) bails out before
the O(n^2) merge:

```ruby
        # This is an O(n^2) operation, so bail out if n is not small
        return same_type_arity_signatures if same_type_arity_signatures.length > 10
```

#1238's own regression spec in `spec/pin/method_spec.rb` is the precise guard
for that bug. What this one covers is the narrower property the PR adds:
`Store#get_methods` combining many real same-path pins without hanging.

Left the reference out of the comment rather than adding it - citing #1238
there would claim a guard the bail-out defeats.

spec/pin/method_spec.rb
apiology: **Claude:**

It proved only the absence of a raise. d1b325d54:

```ruby
  it 'resolves a closureless DuckMethod pin to the core method return type' do
    pin = Solargraph::Pin::DuckMethod.new(name: 'to_s', source: :api_map)
    expect(pin.closure).to be_nil
    expect(pin.return_type).to be_undefined
    expect(pin.typify(api_map).rooted_tags).to eq('::String')
  end
```

The pin has no closure and its own return type is undefined, yet typify still
reaches `Object#to_s` and yields `::String`. That resolution is the property
that would actually regress; a raise fails the example on the way to it.

lib/solargraph/api_map.rb

It was hiding one. With the `@type` dropped:

```
lib/solargraph/api_map.rb:822: Unresolved call to generics on Solargraph::Pin::Base
```

`select` with an `is_a?` block does not narrow the returned element type, so
the array stays `Array<Pin::Base>`. Distinct from the flow-sensitive-local
narrowing tracked in #1241, #1251, #1254 and #1296 - there is no local being
narrowed here; the gap is the block's effect on the collection's element type.
Searched issues and PRs for select / is_a? / element-type narrowing and found
nothing covering it. Not filed.

d1b325d54 replaces the cast with a marker naming the gap:

```ruby
      candidates = store.get_path_pins(fqns).select { |p| p.is_a?(Pin::Namespace) }
      # @sg-ignore select with an is_a? block does not narrow the element type
      candidates.find { |p| !p.generics.empty? } || candidates.first
```

Strong typecheck on the file reports 11 problems with the `@type` gone and no
marker, 10 with the marker, and does not report the marker unneeded. Prose
rather than a URL because `rules.rb` carries no catalogued reason list in this
tree and no upstream issue exists to point at.

spec/pin/method_spec.rb
```diff
@@ -786,4 +786,12 @@ def foo(**bar); end
       expect { pin.signatures }.not_to raise_error
     end
   end
+
+  it 'resolves a closureless DuckMethod pin to the core method return type' do
```
apiology: "core method return type" - what's that?

lib/solargraph/api_map.rb
```diff
@@ -811,6 +812,17 @@ def store
       @store ||= Store.new
     end
 
+    # The namespace pin for fqns that actually declares its generics,
+    # picked from any duplicate pins for the same namespace.
+    #
```
apiology:
```suggestion
    # @todo Consider trade-offs of merging namespace pins instead of trying to choose the best one for this use
```

spec/api_map/store_spec.rb
apiology: 1238 is merged, so let's drop the timeout.

spec/pin/method_spec.rb
apiology: **Claude:**

Nothing - it was vague. The method being resolved to is `Object#to_s`, and the
type is `String`. f6c7defe0:

```ruby
  it 'typifies a closureless DuckMethod pin as String via Object#to_s' do
```

lib/solargraph/api_map.rb

Applied as written in f6c7defe0, replacing the separator line the suggestion
targeted:

```ruby
    # The namespace pin for fqns that actually declares its generics,
    # picked from any duplicate pins for the same namespace.
    # @todo Consider trade-offs of merging namespace pins instead of trying to choose the best one for this use
    # @param fqns [String]
    # @return [Pin::Namespace, nil]
```

Strong typecheck on the file still reports 10 problems and nothing new against
this method, so YARD still reads the tags without the blank separator.

spec/api_map/store_spec.rb
apiology: **Claude:**

Dropped in f6c7defe0 - both wrappers, plus the `require 'timeout'` that came in
with them, and the example renamed for what it actually asserts:

```ruby
    it 'combines many same-path pins into a single pin' do
      maps = (1..30).map do |i|
        Solargraph::SourceMap.load_string(%(
          class Foo
            # @param other [Type#{i}]
            # @return [Type#{i}]
            def bar(other); end
          end
        ), "source#{i}.rb")
      end
      store = described_class.new(maps.flat_map(&:pins))

      bar_pins = store.get_methods('Foo', scope: :instance).select { |p| p.name == 'bar' }
      expect(bar_pins.length).to eq(1)
      # The regression is combinatorial blowup, not a specific merge
      # outcome, so bound the size rather than assert exact merges.
      expect(bar_pins.first.signatures.length).to be <= maps.length
    end
```

The signature-count bound stays: it is the part that would still show a blowup,
and it does not depend on the wall clock.

PR #1290, opened 2026-08-12
Fix @yieldparam type lost on multi-overload block-form methods

### LINKED ISSUE #1289: @yieldparam type on a block-form @overload isn't applied to the block-local variable at the call site
## Body

**Claude:** Filed on behalf of the repo owner while auditing `@sg-ignore` usage in a downstream project.

### Summary

When a method declares two `@overload` entries — one plain, one that adds a `@yieldparam` for its block-form call — Solargraph fails to apply the declared `@yieldparam` type to the block-local variable at the call site. The block param resolves as untyped, and any method called on it inside the block produces `Unresolved call to <method>`.

A single (non-overloaded) method with the same `@yieldparam` type-checks correctly. The bug only appears once the `@yieldparam`-bearing signature is one of *two or more* `@overload` entries on the same method. No `@generic` tag is required to trigger it — this is not a generics bug.

### Minimal reproduction

```ruby
# typed: true
# frozen_string_literal: true

# @overload build
#   @return [String]
# @overload build
#   @yieldparam widget [String]
#   @return [void]
def build
  return 'hi' unless block_given?

  yield 'hi'
end

build do |w|
  w.upcase
end
```

```
repro.rb:12: Unresolved call to upcase
Typecheck finished in ...
1 problem found.
```

A single-overload version of the same method (dropping the plain `@overload` entirely) type-checks clean:

```ruby
# @yieldparam widget [String]
# @return [void]
def build
  yield 'hi'
end

build do |w|
  w.upcase
end
```
→ 0 problems found.

Overload order doesn't matter (tried both orders, same result). `@yieldreturn` and `@generic` are not required — removing them doesn't change the failure.

### Expected behavior

`w` inside the block should resolve to `String` (per the `@yieldparam` tag on the block-form `@overload`) regardless of whether that overload is declared alongside a second, plain overload.

### Workaround

An explicit `# @param w [String]` comment placed **immediately above** the call (not inside the block) clears the error:

```ruby
# @param w [String]
build do |w|
  w.upcase
end
```

An in-block `# @type [String] w` cast on the block-local variable does **not** work — the type appears to already be committed to untyped before that point.

### Environment

- Reproduced against plain, released `solargraph` 0.60.3 (Ruby 3.3.12) — not fork-specific.
- Also reproduced against `apiology/solargraph` fork, branch `2026-08-04`, commit `1933086868c782b9224be19964bb0cbd129858af` (integration branch for #1231/#1258/#1259).

### Real-world impact

Found in `download_uri` in a downstream project, calling `Net::HTTP.start(...) do |http| ... end`. `Net::HTTP.start`'s RBS signature has exactly this shape — a plain overload returning `Net::HTTP`, and a generic block-form overload yielding a `Net::HTTP`. `bin/solargraph pin 'Net::HTTP.start'` confirms the block-form overload's yieldparam is declared correctly as `Net::HTTP`, but at the call site `http.request(...)` fails to resolve, currently worked around with `@sg-ignore`.

### Possibly related

castwide/solargraph#1265 (also filed from this project) is a different mechanism — `@generic T` return type lost when a method has an *explicit* `&block` parameter, no `@overload`/`@yieldparam` involved — but shares the general theme of block declarations interfering with otherwise-correct type inference. Not believed to be a duplicate.

apiology:
## Summary

Fixes https://github.com/castwide/solargraph/issues/1289.

`Pin::Method#generate_signature` always read `@yieldparam`/`@yieldreturn` tags from the method's own docstring, even when it was building the `Signature` for an `@overload` tag. Per YARD, `@overload` docstrings are self-contained, so when a method declared a block-form `@overload` alongside a second, plain `@overload`, the block-form overload's `@yieldparam` was silently dropped — the block-local variable resolved as untyped at the call site.

`generate_signature` now takes the docstring to read those tags from, and `#signatures` passes each overload tag's own docstring instead of the method's.

```ruby
# @overload build
#   @return [String]
# @overload build
#   @yieldparam widget [String]
#   @return [void]
def build
  return 'hi' unless block_given?
  yield 'hi'
end

build do |w|
  w.upcase # was: Unresolved call to upcase
end
```

## Test plan

- [x] `bundle exec rspec` (1443 examples, 0 failures)
- [x] `bundle exec rubocop` on changed files (clean)
- [x] Manually confirmed `solargraph typecheck --level strong` on the issue's repro now reports 0 problems

### DIFF
--- a/lib/solargraph/pin/method.rb
+++ b/lib/solargraph/pin/method.rb
@@ -148,12 +148,13 @@ module Solargraph
 
       # @param parameters [::Array<Parameter>]
       # @param return_type [ComplexType, nil]
+      # @param tag_docstring [YARD::Docstring] source of @yieldparam/@yieldreturn tags; pass an @overload tag's own docstring here
       # @return [Signature]
-      def generate_signature parameters, return_type
+      def generate_signature parameters, return_type, tag_docstring = docstring
         # @type [Pin::Signature, nil]
         block = nil
-        yieldparam_tags = docstring.tags(:yieldparam)
-        yieldreturn_tags = docstring.tags(:yieldreturn)
+        yieldparam_tags = tag_docstring.tags(:yieldparam)
+        yieldreturn_tags = tag_docstring.tags(:yieldreturn)
         generics = docstring.tags(:generic).map(&:name)
         needs_block_param_signature =
           parameters.last&.block? || !yieldreturn_tags.empty? || !yieldparam_tags.empty?
@@ -747,7 +748,9 @@ module Solargraph
         top_type = generate_complex_type
         result = []
         result.push generate_signature(parameters, top_type) if top_type.defined?
-        result.concat(overloads.map { |meth| generate_signature(meth.parameters, meth.return_type) }) unless overloads.empty?
+        # @param meth [Pin::Signature]
+        # @param tag [YARD::Tags::OverloadTag]
+        result.concat(overloads.zip(docstring.tags(:overload).select(&:parameters)).map { |meth, tag| generate_signature(meth.parameters, meth.return_type, tag.docstring) }) unless overloads.empty?
         result.push generate_signature(parameters, @return_type || ComplexType::UNDEFINED) if result.empty?
         result
       end
--- a/spec/pin/method_spec.rb
+++ b/spec/pin/method_spec.rb
@@ -315,6 +315,32 @@ describe Solargraph::Pin::Method do
     expect(kwrestarg_overload.parameters.first.decl).to eq(:kwrestarg)
   end
 
+  it 'applies yieldparam tags from the matching overload only' do
+    pin = described_class.new(name: 'build', comments: %(
+@overload build
+  @return [String]
+@overload build
+  @yieldparam widget [String]
+  @return [void]
+    ))
+    expect(pin.signatures.length).to eq(2)
+    plain, block_form = pin.signatures
+    expect(plain.block).to be_nil
+    expect(block_form.block).not_to be_nil
+    expect(block_form.block.parameters.first.return_type.tag).to eq('String')
+  end
+
+  it 'does not leak a method-level yieldparam into an overload that declares no block' do
+    pin = described_class.new(name: 'foo', comments: %(
+@yieldparam bing [Integer]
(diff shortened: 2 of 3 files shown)


apiology:
*🤖 Filed by Claude, not Vince — acting on his credentials.*

This fixes `@yieldparam` loss for YARD `@overload`-declared block-form methods, but our real-world case (`Net::HTTP.start`, an RBS-declared generic block-form overload) still fails on this branch — a different code path from the one this PR touches (`Pin::Method#generate_signature`/`#signatures`, YARD-only). Bisecting further: the failure is conditional on a keyword argument being present in the call. `Net::HTTP.start('example.com') { |http| ... }` (no kwargs) resolves `http` correctly; `Net::HTTP.start('example.com', use_ssl: true) { |http| ... }` does not. An equivalent plain YARD `@generic`/`@yieldparam` method (no RBS involved) is unaffected by adding a keyword arg to its call — so this looks specific to RBS generic-method-type translation, not a general kwarg-plus-block gap.

```ruby
require 'net/http'

Net::HTTP.start('example.com', use_ssl: true) do |http|
  http.request(Net::HTTP::Get.new('/'))
end
```
```
repro.rb:4: Unresolved call to request
```
Dropping `use_ssl: true` (call with no keyword arg at all) makes this `0 problems found` on the same branch.

apiology:
*🤖 Filed by Claude, not Vince — acting on his credentials.*

Root cause isn't RBS-specific. `Call#inferred_pins` (`lib/solargraph/source/chain/call.rb`) matches call-site arguments to a signature's parameters purely by array index. A trailing keyword-arguments hash (`use_ssl: true`) is just another element in that array, so it gets checked against whatever positional parameter sits at that index — `port` here — instead of the kwrest parameter (`**opt`). The mismatch rejects the whole overload, so the block-form generic overload never matches and `http` falls back to untyped.

Confirmed this predates and is independent of this PR's `0b60c02e4` commit: it reproduces on `master` before that commit, and with a plain YARD `**kwrest` method that has no RBS involvement at all.

Fix and reproduction (matching the case above) are up at https://github.com/castwide/solargraph/pull/1292.

castwide:
:+1: :shipit:

PR #1291, opened 2026-08-12
Fix solargraph-rspec plugin specs

lekemula:
Specs should be fixed from the new release:
https://github.com/lekemula/solargraph-rspec/releases/tag/v0.6.0


castwide:
Thanks!

PR #1293, opened 2026-08-12
Fix Pin::Base#== missing presence, add regression coverage

apiology:
## Summary

`Pin::Base#==` compares `#location` but not `#presence`. `combine_with` results choose the earliest assignment's `#location`, so two combined pins covering a different number of assignments to the same variable can share `#location` while covering different `#presence` ranges — e.g. one pin combined through a variable's first reassignment, another combined through its second. Any caller keying off of `#==` (e.g. `Array#include?`) treated these as the same pin.

`BaseVariable#==` now also compares `presence`, `intersection_return_type`, and `exclude_return_type`.

## Regression coverage

- `spec/pin/base_variable_spec.rb`: directly exercises the equality gap above — fails without the fix, passes with it.
- `spec/type_checker/levels/strong_spec.rb`: a 13-line repro (https://github.com/castwide/solargraph/pull/1288#issuecomment-5273022388) where this equality gap, combined with in-flight flow-sensitive-typing work (#1258, #1282), produces a false "Unresolved call" via `Chain`'s inference recursion guard. Not currently reachable on master alone — verified neither #1258 nor #1282 reproduces it in isolation, only the two combined. Kept as a standing regression guard so whatever future combination reintroduces the failure mode gets caught, regardless of merge order.

## Test plan

- [x] `bundle exec rspec` — 1626 examples, 0 failures, 60 pending (pre-existing)
- [x] `bundle exec rubocop` on touched files — clean
- [x] `bundle exec solargraph typecheck` on touched files — clean (one pre-existing warning on an unmodified line)

### DIFF
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -372,7 +372,19 @@ module Solargraph
       return self if exclude_types.nil?
 
       types = items - exclude_types.items
-      types = [ComplexType::UniqueType::UNDEFINED] if types.empty?
+      if types.empty?
+        # An exclusion built from more than one excluded type can
+        # result from combining independently-derived flow-sensitive
+        # facts (e.g., a `x.nil? || x.is_a?(Foo)` guard) that together
+        # happen to cover every member of the declared type. That's a
+        # sign of unreachable/defensive code, not a real type error -
+        # treat it as a no-op instead of collapsing to undefined, which
+        # would otherwise make legitimate calls in that dead code look
+        # unresolved.
+        return self if exclude_types.items.length > 1
+
+        types = [ComplexType::UniqueType::UNDEFINED]
+      end
       ComplexType.new(types)
     end
 
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -36,7 +36,7 @@ module Solargraph
     def qualify api_map, *gates
       red = reduce_object
       types = red.items.map do |t|
-        next t if %w[nil void undefined].include?(t.name)
+        next t if %w[nil void undefined bot].include?(t.name)
         next t if ['::Boolean'].include?(t.rooted_name)
         api_map.unalias(t.name) || t.qualify(api_map, *gates)
       end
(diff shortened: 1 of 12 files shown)


apiology:
**Claude:** This regresses Solargraph's own self-hosted strong typecheck. Reproduces standalone on this branch (`97127a99d`):

```
git checkout 97127a99d
bundle install
bundle exec rbs collection update
bundle exec solargraph typecheck --level strong lib/solargraph/pin/base.rb
```

gives 11 problems, including two not present on the parent commit (`8fda63384`, 9 problems):
```
lib/solargraph/pin/base.rb:360: Unresolved call to inspect
lib/solargraph/pin/base.rb:362: Unresolved call to inspect
```

Those lines are in `Pin::Base#choose_pin_attr_with_same_name`:

```ruby
# @type [Pin::Base, nil]
val1 = send(attr)
# @type [Pin::Base, nil]
val2 = other.send(attr)
raise "Expected pin for #{attr} on\n#{inspect},\ngot #{val1.inspect}" unless val1.nil? || val1.is_a?(Pin::Base)
```

`val1`/`val2` are explicitly typed `Pin::Base, nil` via the `@type` comment, so `.inspect` (universal on `Object`) should always resolve. Post-merge, `val1`/`val2` infer as the `undefined` pseudo-type at that call site instead.

The causal chain past that observation is not fully traced. Best guess, unconfirmed: `Pin::Base#==` is used for deduplication somewhere in the pin-combining/caching path (e.g. `Array#include?` over collected pins), and making `==` stricter (comparing `presence` in addition to `assignment`) changes which pins Solargraph's self-analysis treats as identical, which in turn changes what `send(attr)`'s type resolution sees at that call site. That's a guess based on the shape of the regression, not something verified by stepping through the combine logic.

Note: `lib/solargraph/pin/base.rb` (where the symptom shows up) isn't touched by this commit at all - the actual change is in `lib/solargraph/pin/base_variable.rb`. Tried reducing this to a standalone synthetic example (a small class mimicking the same `send(attr)` + `@type` + `.inspect` shape) and got no reproduction even inside the full project - this only manifests through Solargraph's whole-project self-analysis, not in isolation. The command above is the smallest reliable reproduction available.

apiology:
Closing in favor of https://github.com/apiology/solargraph/pull/56, which is based on `fix-raise-abort-bot-type` ��� this work now depends on https://github.com/castwide/solargraph/pull/1277, and a base branch must live in the base repo. Will reopen here once #1277 merges to master.

PR #1297, opened 2026-08-13
Narrow via literal-equality, kind_of?, instance_of? and instance_variable_defined? in flow-sensitive typing

### LINKED ISSUE #1296: Flow-sensitive typing doesn't narrow a literal-equality (==/!=) guard against a literal union member
**Claude:** Filed on behalf of the repo owner while auditing `@sg-ignore` markers in an application codebase.

## Summary

When a parameter's declared type is a union including a literal singleton type (e.g. `Array<String>, :not_specified`, via a YARD `@param [Array<String>, :not_specified]` tag), guarding on `!= :not_specified` does not narrow the variable's type inside the guarded branch — the checker still reports the full original union and flags any call to a method that exists on only some of its members. This is despite `is_a?`-based narrowing working correctly for the exact same declared union in the exact same position, which shows the general narrowing mechanism is present but literal-equality comparisons against a literal type aren't feeding into it. Unlike a narrowing check against generic `Symbol`, this isn't a case of "other values could still be there" — `:not_specified` is declared as a literal singleton type, so excluding it by identity should leave nothing but `Array<String>`.

Found in real application code guarding a `sections` parameter declared `@param sections [Array<String>, :not_specified]` before iterating it as an array — `sections != :not_specified` left `sections.each` unresolved at `strong`, requiring an `@sg-ignore`, while switching the same guard to `sections.is_a?(Array)` resolved cleanly with no ignore needed. This looks like a distinct gap from #1254 (raise/return-based nil guards) and #1251 (`is_a?` combined with `&&`/`elsif`) — no return/raise statement or boolean combination is involved here, just a bare literal-equality `if`.

## Repro

```ruby
# typecheck --level strong

# @param sections [Array<String>, :not_specified]
# @return [void]
def not_equal_guard(sections)
  if sections != :not_specified
    # Unresolved call to each on Array<String>, :not_specified -- despite
    # the guard above already excluding the only non-Array member of the
    # declared union (:not_specified is a literal singleton type)
    sections.each { |s| puts s }
  end
end

# @param sections [Array<String>, :not_specified]
# @return [void]
def is_a_guard(sections)
  if sections.is_a?(Array)
    # No error here -- is_a? narrows the same union correctly
    sections.each { |s| puts s }
  end
end
```

Typechecking this reports exactly one problem, on the `sections.each` line inside `not_equal_guard`; `is_a_guard` typechecks clean.

## Expected

Inside the `if sections != :not_specified` branch, `sections` should narrow to `Array<String>`, the same way `is_a?(Array)` already narrows it.

apiology:
**Problem:** Four common guard styles narrow nothing, so a call the guard has already proven safe still fails at `strong`.

```ruby
# @param sections [Array<String>, :not_specified]
def literal_equality(sections)
  return if sections == :not_specified
  # Unresolved call to each on Array<String>, Symbol
  sections.each { |s| puts s }
end

# @param arg [String, Integer]
def kind_of_guard(arg)
  # Unresolved call to upcase on String, Integer
  arg.upcase if arg.kind_of?(String)
end

# @param arg [Object]
def instance_of_guard(arg)
  # Unresolved call to keys on Object
  arg.keys if arg.instance_of?(Hash)
end

# mixed into classes that set @value in their own initialize
module Inspectable
  # @return [String]
  def to_s
    return '(empty)' unless instance_variable_defined?(:@value)
    # Unresolved call to @value
    @value.inspect
  end
end
```

`is_a?` narrows these same unions correctly, so the machinery exists — these four never feed into it.

**Solution:** Adds `process_eq`/`process_neq` (symbol, string, integer and boolean literals, either side of the operator), `process_instance_of` and `process_instance_variable_defined`, and extends `is_a?` narrowing to its `kind_of?` alias — all through the existing fact plumbing, so they compose with `&&`/`||`.

Two deliberate limits: `instance_of?` narrows only its true path, since a subclass instance fails `instance_of?(T)` while still being a `T`; and literal types widen to their class in `qualify` (#1201), which is why the first error above reads `Symbol` rather than `:not_specified`.

`instance_variable_defined?` synthesizes a `Pin::InstanceVariable` scoped to the guard's true-branch range when nothing in scope assigns the ivar — which is why the four node processors now pass `region.closure`.

Fixes #1296

This PR was written by Claude (Anthropics AI assistant) on behalf of @apiology.


lib/solargraph/parser/flow_sensitive_typing.rb
```diff
@@ -5,6 +5,12 @@ module Parser
     class FlowSensitiveTyping
       include Solargraph::Parser::NodeMethods
 
+      # `kind_of?` is an alias of `is_a?`, with the same semantics on
+      # both paths: a true result proves class membership, and a false
+      # result rules out the class and every subclass of it. Both are
+      # handled by #parse_isa.
+      ISA_METHOD_NAMES = %i[is_a? kind_of?].freeze
```
apiology, line 12: Please mention this addition in PR title and description

spec/parser/flow_sensitive_typing_spec.rb
```diff
@@ -1025,4 +1025,291 @@ def check
...
+          other
+        end
+      end
+  ), 'test.rb')
+    api_map = Solargraph::ApiMap.new.map(source)
+    clip = api_map.clip_at('test.rb', [5, 10])
+    expect(clip.infer.to_s).to eq('Object')
+  end
+
+  it 'uses kind_of? in a simple if() to refine types like is_a?' do
```
apiology: "like is_a?" turns this from description into a changelog, which belongs in git commit messages and PR descriptions, not in code names and comments.

spec/parser/flow_sensitive_typing_spec.rb
```diff
@@ -1025,4 +1025,291 @@ def check
...
+  it 'keeps the guarded arm on the false path of instance_of?, unlike is_a?' do
+    # a Repro value declared as ReproBase fails instance_of?(ReproBase)
+    # while still being a ReproBase, so the false path cannot exclude
+    # the ReproBase arm the way is_a? can
+    isa_source = %(
+      class ReproBase; end
+      class Repro < ReproBase; end
+      # @param repr [ReproBase, String]
+      def verify_repro(repr)
+        if repr.%s(ReproBase)
```
apiology, line 1346: Could you explain the syntax of the above statement to me?

apiology: **Claude:**

`%(...)` is a general-delimited Ruby string literal (same as double
quotes). `%s` inside it is a `Kernel#format`/`sprintf` placeholder, not
Ruby interpolation - it gets filled in later via `format(isa_source,
'instance_of?')` and `format(isa_source, method_name)` in the loop below,
so the same class/method template runs once for instance_of? and once
each for is_a?/kind_of?, instead of duplicating the source three times.

lib/solargraph/parser/flow_sensitive_typing.rb
apiology: **Claude:** Already covered - the PR title ("Narrow literal-equality, respond_to?, kind_of?, and instance_of? guards...") and the Solution section of the description both mention kind_of? explicitly. Nothing further needed here.

spec/parser/flow_sensitive_typing_spec.rb
apiology: **Claude:** Already fixed - the branch tip (d6d2ee834) already dropped the changelog phrasing; the spec description now reads "uses kind_of? in a simple if() to refine types" with no "like is_a?" comparison.

lib/solargraph/complex_type.rb
```diff
@@ -401,14 +401,30 @@ def intersect_with intersection_type, api_map
...
-          if int_type.conforms_to?(api_map, ut, :assignment)
+          if int_type.duck_type?
+            # A duck-type fact (e.g. from a respond_to? guard) selects
+            # the union arms that already provide the method; arms that
+            # don't are excluded by the guard, not replaced by the duck
+            # type. If no arm provides it (an opaque receiver like
+            # Object), the fall-through below keeps the bare duck type.
+            # UniqueType#conforms_to? can't express this test: its
+            # inferred-side duck_type? short-circuit answers true in
+            # the wrong direction.
```
apiology: 2 lines at most please.  Why are we wrapping this in ComplexType?

spec/parser/flow_sensitive_typing_spec.rb
```diff
@@ -1025,4 +1025,437 @@ def check
...
+  end
+
+  # The next two exercise defensive guards in
+  # FlowSensitiveTyping#process_instance_variable_defined that no real
+  # parsed source can reach through its only caller, #process_calls:
+  # that caller already filters to :send nodes before calling it, and
+  # #process_if/#process_while/etc. always hand it a closure built from
+  # a Region (which defaults to an anonymous Pin::Namespace, never
+  # nil). Calling the private method directly is the only way to cover
+  # them.
```
apiology: Is there a restructuring of the code that would make this completely unused defensive code unnecessary?

lib/solargraph/complex_type.rb
```diff
@@ -401,14 +401,24 @@ def intersect_with intersection_type, api_map
       # try to find common types via conformance
       items.each do |ut|
         intersection_type.each do |int_type|
-          if int_type.conforms_to?(api_map, ut, :assignment)
+          if int_type.duck_type?
+            # A duck-type fact selects arms that already provide the
+            # method; #conforms_to?'s own duck-type check matches unconditionally instead.
+            types << ut if duck_types_match?(api_map, int_type, ComplexType.new([ut]))
```
apiology: I think this is a mistake, and we should yank the duck type content from this PR.  The fact that a duck type matches one but not both of the union doesn't mean anything - we could be dealing with a subclass instead that /does/ include the method

apiology: **Claude:**

Yanked. `intersect_with` is byte-identical to master again - both the
duck-type branch and the duck-candidate fall-through are gone.

Taken broadly, on the reading that the subclass argument leaves
`process_respond_to` with nothing sound left to feed: the method, its
`process_calls` dispatch line, and the four respond_to? specs are
removed as well, plus `UniqueType#select` and
`spec/complex_type/intersect_with_spec.rb`, which existed only so the
removed fall-through could call `select` on a bare `UniqueType`.

The other four narrowing forms in this branch - literal equality,
`kind_of?`, `instance_of?` and `instance_variable_defined?` - are
untouched. The narrower reading, dropping only the union-member
selection and keeping `process_respond_to` against master's
`intersect_with`, would be a smaller revert layered on top of this one.

Made in ebd845fb8.

lib/solargraph/complex_type.rb
apiology: **Claude:**

The line is gone with the duck-type revert.

On the wrap itself: it was unnecessary. `duck_types_match?` declares
`inferred` as `[ComplexType, UniqueType]` and only calls `#all?` and
`#any?` on it, both of which `UniqueType` defines, so the bare `ut`
would have worked.

Made in ebd845fb8.

spec/parser/flow_sensitive_typing_spec.rb
apiology: **Claude:**

Yes. `closure` is now a required argument instead of defaulting to nil.
All four call sites - the and/if/or/while node processors - already
pass `region.closure`, and `Region#initialize` guarantees it non-nil by
defaulting to an anonymous `Pin::Namespace`, so `return if
closure.nil?` is unreachable and removed.

The other guard's `node.type == :send` half is removed too:
`#process_calls` establishes that before dispatching.

With both guards gone the two direct-invocation specs and this comment
block are deleted.

Made in fac0656fb.

PR #1299, opened 2026-08-14
Resolve generic type variables against union @param types

### LINKED ISSUE #1298: Generic type variable is not solved when the @param type is a union (generic<A>, nil)
**Claude:** Written by Claude and posted using @apiology's GitHub credentials, while auditing `@sg-ignore` usage in a downstream project.

When a `@param` type is a union containing `generic<A>`, `A` isn't solved against the union — it appears to bind to the whole argument type, so the return stays a union and every downstream call fails. `sorbet-runtime`'s `T.must` is this exact shape. The trigger is the union, not `nil` — `@param arg [generic<A>, Integer]` fails the same way, while a bare `@param arg [generic<A>]` binds correctly.

```ruby
# @generic A
# @param arg [generic<A>, nil]
# @return [generic<A>]
def must_nilable(arg)
  raise if arg.nil?

  arg
end

# @param arg [String, nil]
# @return [Integer]
def via_nilable(arg) = must_nilable(arg).length
```

```
repro.rb:7: Declared return type generic<A> does not match inferred type generic<A>, nil for #must_nilable
repro.rb:15: #via_nilable return type could not be inferred
repro.rb:15: Unresolved call to length on String, nil
3 problems found.
```

Expected 0 problems. The two line-15 errors are this bug. The line-7 error is #1276 (`raise` doesn't narrow), not this — an identical helper with no generics at all (`@param arg [String, nil]` / `@return [String]`) produces the same complaint, so it's independent.

Possibly the argument-side counterpart of #1272; not fixed by #1273.

**Environment:** `castwide/solargraph` master `8fda63384`, `rbs` 4.0.3, Ruby 3.3.12.

apiology:
**Problem:** A generic declared inside a union `@param` makes the call site report an error that isn't real.

```ruby
# @generic A
# @param arg [generic<A>, nil]
# @return [generic<A>]
def must_nilable(arg)
  raise ArgumentError, 'arg must not be nil' if arg.nil?

  arg
end

# @param arg [String, nil]
# @return [Integer]
def via_nilable(arg) = must_nilable(arg).length
#                                       ^ Unresolved call to length on String, nil
```

Any generic method with a nilable parameter hits this, since `generic<A>, nil` is the ordinary way to declare one. Fixes #1298.

**Solution:** `ComplexType#resolve_generics_from_context` subtracts the type's own concrete union members from the context before a generic member binds, leaving the generic only what is not already accounted for.

The three type-checker examples are `pending` on #1276, since `raise` does not narrow a type yet. The `ComplexType` table specs carry the regression guard.

### DIFF
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -50,9 +50,10 @@ module Solargraph
     def resolve_generics_from_context generics_to_resolve, context_type, resolved_generic_values: {}
       return self unless generic?
 
-      ComplexType.new(@items.map do |i|
-        i.resolve_generics_from_context(generics_to_resolve, generic_item_context_type(i, context_type),
-                                        resolved_generic_values: resolved_generic_values)
+      ComplexType.new(@items.map do |unique_type|
+        unique_type.resolve_generics_from_context(generics_to_resolve,
+                                                  destructure_context_type(unique_type, context_type),
+                                                  resolved_generic_values: resolved_generic_values)
       end)
     end
 
@@ -556,11 +557,11 @@ module Solargraph
     # `generic<A>, nil` would bind to the entire `String, nil` context
     # instead of just `String`.
     #
-    # @param item [UniqueType] A member of @items
+    # @param unique_type [UniqueType] A member of @items
     # @param context_type [ComplexType, UniqueType, nil]
     # @return [ComplexType, UniqueType, nil]
-    def generic_item_context_type item, context_type
-      return context_type unless item.generic? && context_type.is_a?(ComplexType) && @items.length > 1
+    def destructure_context_type unique_type, context_type
+      return context_type unless unique_type.generic? && context_type.is_a?(ComplexType) && @items.length > 1
 
       concrete_items = @items.reject(&:generic?)
       return context_type if concrete_items.empty?
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -562,13 +562,9 @@ module Solargraph
 
     private
 
-    # When this type is itself a union (e.g. `generic<A>, nil`) being
-    # matched against a union context type (e.g. `String, nil`), a
-    # generic member should only be bound against the parts of the
-    # context union not already accounted for by this type's other,
-    # concrete (non-generic) members. Otherwise `generic<A>` in
-    # `generic<A>, nil` would bind to the entire `String, nil` context
-    # instead of just `String`.
+    # Narrows what a generic member binds against: context members this
+    # type already matches concretely are dropped, so the generic takes
+    # only what is left over rather than the whole union.
     #
     # @param unique_type [UniqueType] A member of @items
     # @param context_type [ComplexType, UniqueType, nil]
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -51,7 +51,7 @@ module Solargraph
       return self unless generic?
 
       ComplexType.new(@items.map do |i|
-        i.resolve_generics_from_context(generics_to_resolve, context_type,
+        i.resolve_generics_from_context(generics_to_resolve, generic_item_context_type(i, context_type),
                                         resolved_generic_values: resolved_generic_values)
       end)
     end
@@ -547,6 +547,29 @@ module Solargraph
     BOT = ComplexType.parse('bot')
 
     private
+
+    # When this type is itself a union (e.g. `generic<A>, nil`) being
+    # matched against a union context type (e.g. `String, nil`), a
+    # generic member should only be bound against the parts of the
+    # context union not already accounted for by this type's other,
+    # concrete (non-generic) members. Otherwise `generic<A>` in
+    # `generic<A>, nil` would bind to the entire `String, nil` context
+    # instead of just `String`.
+    #
+    # @param item [UniqueType] A member of @items
+    # @param context_type [ComplexType, UniqueType, nil]
+    # @return [ComplexType, UniqueType, nil]
+    def generic_item_context_type item, context_type
+      return context_type unless item.generic? && context_type.is_a?(ComplexType) && @items.length > 1
+
+      concrete_items = @items.reject(&:generic?)
+      return context_type if concrete_items.empty?
+
+      remaining_items = context_type.items.reject { |ct| concrete_items.any? { |ci| ci.name == ct.name } }
+      return context_type if remaining_items.empty?
+
+      ComplexType.new(remaining_items)
+    end
 
     # @todo This is a quick and dirty hack that forces `self` keywords
     #   to reference an instance of their class and never the class itself.
--- a/lib/solargraph/complex_type.rb
+++ b/lib/solargraph/complex_type.rb
@@ -562,9 +562,9 @@ module Solargraph
 
     private
 
-    # Narrows what a generic member binds against: context members this
-    # type already matches concretely are dropped, so the generic takes
-    # only what is left over rather than the whole union.
+    # Narrows what a generic member binds against: context members this type
+    # already matches concretely are dropped. Nothing left over means no value
+    # can reach the generic, which is bot.
     #
     # @param unique_type [UniqueType] A member of @items
     # @param context_type [ComplexType, UniqueType, nil]
@@ -576,7 +576,7 @@ module Solargraph
       return context_type if concrete_items.empty?
 
       remaining_items = context_type.items.reject { |ct| concrete_items.any? { |ci| ci.name == ct.name } }
-      return context_type if remaining_items.empty?
+      return BOT if remaining_items.empty?
 
       ComplexType.new(remaining_items)
     end
--- a/spec/complex_type_spec.rb
+++ b/spec/complex_type_spec.rb
@@ -552,13 +552,12 @@ describe 'YARD type specifier list parsing' do
                                                  { 'K' => 'String', 'V' => 'Integer' }],
         non_discriminating_preserves_already_resolved_generic: ['generic<A>, nil', 'String, nil', { 'A' => 'Integer' },
                                                                 %w[A], 'Integer, nil', { 'A' => 'Integer' }],
-        # Known limitation: no concrete member to discriminate by, so both generics bind to the whole context.
-        no_discriminating_context_member: ['generic<A>, generic<B>', 'String, Integer', {}, %w[A B], 'String, Integer',
-                                           { 'A' => 'String, Integer', 'B' => 'String, Integer' }],
         discriminating_reordered_context_members: ['generic<A>, nil', 'nil, String', {}, %w[A], 'String, nil',
-                                                   { 'A' => 'String' }],
-        # Known limitation: duplicate concrete context members cancel out, leaving nothing to subtract.
-        duplicate_concrete_context_members: ['generic<A>, nil', 'nil, nil', {}, %w[A], 'nil', { 'A' => 'nil' }]
+                                                   { 'A' => 'String' }]
+        # Left out: nothing concrete to subtract, and union members are unordered, so
+        # what these should resolve to is undecided.
+        # no_discriminating_context_member: ['generic<A>, generic<B>', 'String, Integer', {}, %w[A B], ...],
+        # duplicate_concrete_context_members: ['generic<A>, nil', 'nil, nil', {}, %w[A], ...]
       }.freeze
 
       UNION_COMPLEX_TYPE_GENERIC_TESTS.each do |name, (tag, context_type_tag, unfrozen_input_map, generics_to_resolve, expected_to_s, expected_output_map)|
--- a/spec/complex_type_spec.rb
+++ b/spec/complex_type_spec.rb
@@ -533,6 +533,59 @@ describe 'YARD type specifier list parsing' do
           end
         end
       end
+
+      # Exercises ComplexType#resolve_generics_from_context directly
+      # (as opposed to UniqueType#resolve_generics_from_context above)
+      # against union types with more than one member, per
+      # https://github.com/castwide/solargraph/issues/1298
+      #
+      # Note: the resolved_generic_values assertion below must compare
+      # with #to_s, not #tag - ComplexType#tag delegates to the first
+      # union member only (via method_missing), so a binding that's
+      # incorrectly the *entire* multi-member context union - the
+      # #1298 bug - can still report the right #tag while silently
+      # carrying the rest of the union along. #to_s renders every
+      # member and catches that.
+      UNION_COMPLEX_TYPE_GENERIC_TESTS = [
+        # tag, context_type_tag, unfrozen_input_map, generics_to_resolve, expected_to_s, expected_output_map
+        # Discriminating: fails pre-fix with A bound to 'String, nil' (the whole context) instead of 'String'.
+        ['generic<A>, nil', 'String, nil', {}, %w[A], 'String, nil', { 'A' => 'String' }],
+        # Discriminating: fails pre-fix with A bound to 'String, nil, Symbol' instead of 'String'.
+        ['generic<A>, nil, Symbol', 'String, nil, Symbol', {}, %w[A], 'String, nil, Symbol', { 'A' => 'String' }],
+        # Non-discriminating (passes either way): context has only one member, so there's
+        # nothing to subtract - included as a no-regression check for the single-member-context shape.
+        ['generic<A>, nil', 'String', {}, %w[A], 'String, nil', { 'A' => 'String' }],
+        # Non-discriminating: the nested generic<A> is resolved against context.subtypes, and
+        # ComplexType#method_missing's delegation to the first union member already narrows this
+        # correctly with or without the fix. Included to confirm the fix doesn't disturb it.
+        ['Array<generic<A>>, nil', 'Array<String>, nil', {}, %w[A], 'Array<String>, nil', { 'A' => 'String' }],
+        # Non-discriminating, same reason as the Array case above, but for Hash key/value generics.
+        ['Hash{generic<K> => generic<V>}, nil', 'Hash{String => Integer}, nil', {}, %w[K V],
+         'Hash{String => Integer}, nil', { 'K' => 'String', 'V' => 'Integer' }],
+        # Non-discriminating: an already-resolved value is never overwritten by context,
+        # with or without the fix. Included as a no-regression check.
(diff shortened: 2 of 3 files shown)


lib/solargraph/complex_type.rb
```diff
@@ -51,7 +51,7 @@ def resolve_generics_from_context generics_to_resolve, context_type, resolved_ge
       return self unless generic?
 
       ComplexType.new(@items.map do |i|
-        i.resolve_generics_from_context(generics_to_resolve, context_type,
+        i.resolve_generics_from_context(generics_to_resolve, generic_item_context_type(i, context_type),
                                         resolved_generic_values: resolved_generic_values)
       end)
```
apiology, line 57: maybe 'destructure_context_type'?

can we pick a better name for 'i' and the 'item' parameter below?

lib/solargraph/complex_type.rb
```diff
@@ -548,6 +548,29 @@ def try_parse *strings
...
+    # @param context_type [ComplexType, UniqueType, nil]
+    # @return [ComplexType, UniqueType, nil]
+    def generic_item_context_type item, context_type
+      return context_type unless item.generic? && context_type.is_a?(ComplexType) && @items.length > 1
+
+      concrete_items = @items.reject(&:generic?)
+      return context_type if concrete_items.empty?
+
+      remaining_items = context_type.items.reject { |ct| concrete_items.any? { |ci| ci.name == ct.name } }
+      return context_type if remaining_items.empty?
```
apiology: Do we have specs to cover when we try to resolve T<A, nil> with T<nil, String> - i.e., out of order?  Or when there are duplicate types inside - T<A, nil> with T<nil, nil> etc?

lib/solargraph/complex_type.rb
apiology: **Claude:**

Renamed `generic_item_context_type` to `destructure_context_type`, and its `item`/`i` parameter and block-variable names to `unique_type` throughout, in 42a6372.

lib/solargraph/complex_type.rb
apiology: **Claude:**

Added both to `UNION_COMPLEX_TYPE_GENERIC_TESTS` in 42a6372:

- Out-of-order (`generic<A>, nil` against context `nil, String` instead of
  `String, nil`): the concrete-member subtraction matches by name, not
  position, so this resolves correctly - `A` binds to `String`.
- Duplicate concrete member (`generic<A>, nil` against context `nil, nil`):
  this is a known limitation, unchanged by this fix. Both context members
  match and get subtracted, leaving nothing to subtract from, so it falls
  back to the full (deduplicated) context union rather than a meaningful
  remainder - `A` ends up bound to `nil`. Documented as a no-regression
  check, not as evidence this shape is handled well.

spec/complex_type_spec.rb
```diff
@@ -533,6 +533,69 @@
...
+      # against union types with more than one member, per
+      # https://github.com/castwide/solargraph/issues/1298
+      #
+      # Note: the resolved_generic_values assertion below must compare
+      # with #to_s, not #tag - ComplexType#tag delegates to the first
+      # union member only (via method_missing), so a binding that's
+      # incorrectly the *entire* multi-member context union - the
+      # #1298 bug - can still report the right #tag while silently
+      # carrying the rest of the union along. #to_s renders every
+      # member and catches that.
```
apiology: 4 lines at most above, one line at most for each test below in the array.  Consider using a hash so you have a name to attach to help explain why it's there.

spec/complex_type_spec.rb
```diff
@@ -533,6 +533,69 @@
...
+        ['Array<generic<A>>, nil', 'Array<String>, nil', {}, %w[A], 'Array<String>, nil', { 'A' => 'String' }],
+        # Non-discriminating, same reason as the Array case above, but for Hash key/value generics.
+        ['Hash{generic<K> => generic<V>}, nil', 'Hash{String => Integer}, nil', {}, %w[K V],
+         'Hash{String => Integer}, nil', { 'K' => 'String', 'V' => 'Integer' }],
+        # Non-discriminating: an already-resolved value is never overwritten by context,
+        # with or without the fix. Included as a no-regression check.
+        ['generic<A>, nil', 'String, nil', { 'A' => 'Integer' }, %w[A], 'Integer, nil', { 'A' => 'Integer' }],
+        # Known limitation, unchanged by this fix: with no concrete union member to
+        # subtract, there's no positional information to tell which generic should
+        # bind to which context member, so both end up bound to the entire context union.
```
apiology: This should be made obvious in the data structure, with athe real desired result documented.  Or comment it out rather than describe the bad behavior.

spec/type_checker/levels/strong_spec.rb
```diff
@@ -707,6 +707,104 @@ def objects_by_class klass
...
+      expect(messages).not_to include('Unresolved call to length on String, nil')
+    end
+
+    # NOTE: this scenario doesn't distinguish fixed from unfixed
+    # behavior - a union return type of the exact same shape as the
+    # union @param type flattens/dedupes an incorrectly-too-wide
+    # binding back down to the right answer by coincidence (the union
+    # already contains 'nil' as a sibling member either way). It's
+    # included as a no-regression check for union return types, not as
+    # a regression test for #1298 itself.
```
apiology: No changelogging

spec/type_checker/levels/strong_spec.rb
```diff
@@ -707,6 +707,104 @@ def objects_by_class klass
...
+        # @return [generic<A>]
+        def layer2(arg) = layer1(arg)
+
+        # @param arg [String, nil]
+        # @return [Integer]
+        def via_layers(arg) = layer2(arg).length
+      ))
+      # As above, any "Declared return type...does not match inferred
+      # type" problems here are #1276 (`raise` does not narrow the
+      # type), independent of this test.
```
apiology: No changelogging.  Add a pending with a URL if you want to document future desired behavior

spec/complex_type_spec.rb
apiology: **Claude:**

Done in 19095d37f. The array is a hash now, keyed by name
(`discriminating_two_member_union`, `non_discriminating_nested_hash_generic`,
and so on), the header comment is down to four lines, and each entry that
still carries a comment gets one line.

spec/complex_type_spec.rb
apiology: **Claude:**

Took the second option in 6f320355e. Both entries are commented out rather
than asserting the current answer:

    # Left out: nothing concrete to subtract, and union members are unordered, so
    # what these should resolve to is undecided.
    # no_discriminating_context_member: ['generic<A>, generic<B>', 'String, Integer', {}, %w[A B], ...],
    # duplicate_concrete_context_members: ['generic<A>, nil', 'nil, nil', {}, %w[A], ...]

Documenting the real desired result is not available yet. Union members are
unordered, so nothing says that `generic<A>, generic<B>` against
`String, Integer` should bind A to String rather than to Integer, and the
`nil, nil` case has nothing concrete left once it dedupes. Neither has a
settled answer to assert, and a pending would need an issue URL that does not
exist - #1298 closes with this PR.

spec/type_checker/levels/strong_spec.rb

Cut in 19095d37f to the live constraint plus a bare URL:

    # The remaining problem is unrelated to this test:
    # https://github.com/castwide/solargraph/issues/1276

6f320355e puts the other two occurrences into the same form. Those two had
also been pointing at each other positionally, with "as above" and "the
two-member case above".

spec/type_checker/levels/strong_spec.rb

Removed the note in 6f320355e. A pending does not fit here: the example passes
today, so there is no future desired behavior to assert. What the note was
recording is that this example does not discriminate the change under test,
which belongs in the pull request description rather than in the spec.

lib/solargraph/complex_type.rb
```diff
@@ -561,6 +562,29 @@ def try_parse *strings
 
     private
 
+    # When this type is itself a union (e.g. `generic<A>, nil`) being
```
apiology: Let's reduce the size of this an emphasize what it accomplishes and why, less of the details of each type case

spec/type_checker/levels/strong_spec.rb
```diff
@@ -779,6 +779,93 @@ def objects_by_class klass
       end
     end
 
+    it 'resolves a generic type variable against a union @param type' do
+      checker = type_checker(%(
+        # @generic A
+        # @param arg [generic<A>, nil]
+        # @return [generic<A>]
+        def must_nilable(arg)
+          raise if arg.nil?
```
apiology: Make this a real exception being raised, not an invalid re-raise

spec/type_checker/levels/strong_spec.rb
```diff
@@ -779,6 +779,93 @@ def objects_by_class klass
...
+
+          arg
+        end
+
+        # @param arg [String, nil]
+        # @return [Integer]
+        def via_nilable(arg) = must_nilable(arg).length
+      ))
+      # The remaining problem is unrelated to this test:
+      # https://github.com/castwide/solargraph/issues/1276
```
apiology: Explain 'remaining problem' - we should assert what the code should do and add a pending statement with the future issue/PR if they don't work yet

spec/type_checker/levels/strong_spec.rb
```diff
@@ -779,6 +779,93 @@ def objects_by_class klass
...
+
+          arg
+        end
+
+        # @param arg [String, nil, Symbol]
+        # @return [Integer]
+        def via_triple(arg) = must_not_nil_or_symbol(arg).length
+      ))
+      # The remaining problem is unrelated to this test:
+      # https://github.com/castwide/solargraph/issues/1276
```
apiology: ???

spec/type_checker/levels/strong_spec.rb
```diff
@@ -779,6 +779,93 @@ def objects_by_class klass
...
+        # @param arg [generic<A>, nil]
+        # @return [generic<A>]
+        def layer2(arg) = layer1(arg)
+
+        # @param arg [String, nil]
+        # @return [Integer]
+        def via_layers(arg) = layer2(arg).length
+      ))
+      # The remaining problem is unrelated to this test:
+      # https://github.com/castwide/solargraph/issues/1276
```
apiology: ???

spec/complex_type_spec.rb
```diff
@@ -533,6 +533,47 @@
...
+                                                 {}, %w[K V], 'Hash{String => Integer}, nil',
+                                                 { 'K' => 'String', 'V' => 'Integer' }],
+        non_discriminating_preserves_already_resolved_generic: ['generic<A>, nil', 'String, nil', { 'A' => 'Integer' },
+                                                                %w[A], 'Integer, nil', { 'A' => 'Integer' }],
+        discriminating_reordered_context_members: ['generic<A>, nil', 'nil, String', {}, %w[A], 'String, nil',
+                                                   { 'A' => 'String' }]
+        # Left out: nothing concrete to subtract, and union members are unordered, so
+        # what these should resolve to is undecided.
+        # no_discriminating_context_member: ['generic<A>, generic<B>', 'String, Integer', {}, %w[A B], ...],
+        # duplicate_concrete_context_members: ['generic<A>, nil', 'nil, nil', {}, %w[A], ...]
```
apiology: Why are these undecidable?  Should we be using a different algorithm which is able to resolve things?

lib/solargraph/complex_type.rb
apiology: **Claude:**

Cut to three lines in 64d9bd372, dropping the worked example:

    # Narrows what a generic member binds against: context members this
    # type already matches concretely are dropped, so the generic takes
    # only what is left over rather than the whole union.

The YARD tags below it stay - removing those adds strong-typecheck problems.

spec/type_checker/levels/strong_spec.rb

A bare `raise` with no exception in flight re-raises nothing. Fixed in 64d9bd372
in all three fixtures that had it, not just this one:

    raise ArgumentError, 'arg must not be nil' if arg.nil?
    raise ArgumentError, 'arg must not be nil or a Symbol' if arg.nil? || arg.is_a?(Symbol)

spec/type_checker/levels/strong_spec.rb

Done in 64d9bd372. All three examples blocked by #1276 now assert the desired end
state under a pending, and the comment is gone rather than reworded:

    it 'resolves a generic type variable against a union @param type' do
      pending 'https://github.com/castwide/solargraph/issues/1276'
      checker = type_checker(%(
        ...
      ))
      expect(checker.problems.map(&:message)).to be_empty
    end

Each of the three reports as pending rather than passing unexpectedly: 158
examples, 0 failures, 13 pending, up from 10.

One cost worth naming: these three stop guarding #1298, since they now fail as
pending whether or not union resolution regresses. The
UNION_COMPLEX_TYPE_GENERIC_TESTS table and the same-union example, which still
passes outright, carry that guard.

spec/type_checker/levels/strong_spec.rb

That comment named neither the problem nor its scope, so it earned the question.
Removed in 64d9bd372. The three-member example now asserts an empty problem list
under `pending 'https://github.com/castwide/solargraph/issues/1276'`, the same
shape as the two-member one.

spec/type_checker/levels/strong_spec.rb

Same defect and same fix as the two-member and three-member examples: the comment
is gone in 64d9bd372, and the two-layer example asserts an empty problem list
under `pending 'https://github.com/castwide/solargraph/issues/1276'`.

spec/complex_type_spec.rb
apiology: **Claude:**

Correction first. My earlier reply said the second entry was duplicate context
members cancelling out. That is wrong: `ComplexType.parse('nil, nil')` returns a
single `nil` item, so duplicates never reach the algorithm at all. What actually
happens is that the concrete member covers the whole context and leaves nothing
over. Renamed to `context_fully_covered_by_concrete_member` in 64d9bd372, with
the note corrected to match.

The two entries are different kinds of question, and neither is undecidable in
the strict sense - I overstated that.

`generic<A>, generic<B>` against `String, Integer`: items are ordered, so a
positional rule is available and would discriminate. Observed:

    parse('generic<A>, generic<B>') vs parse('String, Integer')  => A = 'String, Integer',  B = 'String, Integer'
    parse('generic<A>, generic<B>') vs parse('Integer, String')  => A = 'Integer, String',  B = 'Integer, String'

Those two context declarations describe the same type, and a positional rule
would return A = String for the first and A = Integer for the second. The
objection is not that no algorithm exists, it is that any discriminating one
reads meaning out of member order that the type system does not put there.

`generic<A>, nil` against `nil`: here a different algorithm really is on the
table. `destructure_context_type` returns the whole context when subtraction
leaves nothing, so A binds to `nil`. The alternative is to leave A unresolved. I
have not implemented that to measure it. What is observed is that an unresolvable
context propagates rather than being dropped:

    parse('generic<A>, nil').resolve_generics_from_context(['A'], parse('undefined'))
    => 'undefined', A => 'undefined'

So the inferred trade-off is a concrete binding that may be wrong against an
undefined return that is definitely useless. That is a change to the resolution
rule rather than to this PR, which only stops the generic swallowing the whole
union when there is something concrete to subtract.

spec/complex_type_spec.rb
```diff
@@ -533,6 +533,48 @@
           end
         end
       end
+
+      # Exercises ComplexType#resolve_generics_from_context against union
+      # types with more than one member. Compares #to_s, not #tag, since
+      # #tag delegates to only the first union member (method_missing).
+      # https://github.com/castwide/solargraph/issues/1298
```
apiology: Why are we listing a URL here?  

Compare rooted_tags please, not .to_s or .tag.  Don't document complextype here please, that's a changelog not a doc.

spec/complex_type_spec.rb
```diff
@@ -533,6 +533,48 @@
...
+                                                 { 'K' => 'String', 'V' => 'Integer' }],
+        non_discriminating_preserves_already_resolved_generic: ['generic<A>, nil', 'String, nil', { 'A' => 'Integer' },
+                                                                %w[A], 'Integer, nil', { 'A' => 'Integer' }],
+        discriminating_reordered_context_members: ['generic<A>, nil', 'nil, String', {}, %w[A], 'String, nil',
+                                                   { 'A' => 'String' }]
+        # Left out as open questions. The first has no concrete member to subtract,
+        # and telling A from B would need union member order, which carries no
+        # meaning. The second leaves nothing over, so A binds to the whole context.
+        # no_discriminating_context_member: ['generic<A>, generic<B>', 'String, Integer', {}, %w[A B], ...],
+        # context_fully_covered_by_concrete_member: ['generic<A>, nil', 'nil', {}, %w[A], ...]
```
apiology: Please ask your open questions.

apiology: **Claude:**

Commented-out entries removed in 0521cca56. Both questions are about what
`resolve_generics_from_context` should do, not what it does:

1. `generic<A>, generic<B>` against context `String, Integer` binds both A and B
   to the whole `String, Integer`. `@items` order is preserved, so a positional
   rule is available and would give A = String, B = Integer - but the same rule
   gives A = Integer for context `Integer, String`, which declares the same type.
   Should this shape resolve at all, or is binding both to the whole union the
   answer you want?

2. `generic<A>, nil` against context `nil` binds A to `nil`, because
   `destructure_context_type` returns the whole context when subtracting the
   concrete members leaves nothing behind. The alternative is to leave A
   unresolved, which makes the return type `undefined`. Which do you prefer?

Neither shape is exercised by the table now, so either answer is a follow-up
rather than part of this PR.

apiology: **Claude:**

Question 1 is answered: bot. 0daa2ecd3 returns `ComplexType::BOT` when
subtraction leaves nothing, since no value can reach the generic there - the
same reason a method whose every path raises returns bot.

Two cases in `UNION_COMPLEX_TYPE_GENERIC_BOT_TESTS` assert the end state, both
pending on #1277:

    context_covered_by_one_member
      'generic<A>, nil' vs 'nil'              => 'nil',         A = bot
    context_covered_by_two_members
      'generic<A>, String, nil' vs 'String, nil'  => 'String, nil', A = bot

They fail today because a union keeps bot as a visible member instead of
absorbing it.

Measured at the call site: the change removes `Unresolved call to length on
String, nil`. `#call_unwrap return type could not be inferred` survives,
because a call on bot yields nothing - the same gap #1277 covers. Full suite
is 1659 examples, 0 failures.

Question 2, the union of two generics with nothing concrete to subtract, is
still open.

PR #1301, opened 2026-08-15
Don't let a bracketed comment in a class body drop its superclass

### LINKED ISSUE #1300: Unrelated comment removes superclass pin?
Commenting out a line of code inside a class body — an ordinary thing to do while iterating on a list of test cases — can make Solargraph lose that class's superclass entirely. Every inherited method then reports `Unresolved call` throughout the class, including on lines *above* the comment, so the reported locations point nowhere near the cause. In our codebase this cost 14 suppression comments in one file and went undiagnosed for months, because "a comment changed type resolution" is not a hypothesis anyone reaches for.

```ruby
# frozen_string_literal: true

class Base
  # @return [String]
  def greet
    'hi'
  end
end

class Sub < Base
  # [:b, { c: :d }]

  # @return [void]
  def call_it
    greet
  end
end
```

```
$ solargraph typecheck --level strong probe.rb
probe.rb:15: Unresolved call to greet
1 problem found.
```

Expected: `0 problems found`. Rewording that line as prose (`# b maps c to d`) passes. Verified on master `8fda6338` ("Release 0.60.3"), ruby 3.3.12, rbs 4.0.3.

The trigger looks specific to **inline RBS parsing of a tuple containing a record type**, in a plain comment that was never marked as an annotation:

| Comment | Result |
|---|---|
| `# [{ c: :d }]` — RBS tuple-of-record | **breaks** |
| `# [{ c: Integer }]` — RBS tuple-of-record | **breaks** |
| `# [{}]` — RBS tuple-of-empty-record | **breaks** |
| `# [{ String => Integer }]` — YARD hash type | clean |
| `# [Integer, String]` — RBS tuple, no record | clean |
| `# { c: Integer }` — RBS record, no brackets | clean |
| `# [:b]`, `# [1]`, `# []`, `# {}`, prose | clean |

It needs brackets *around* a brace-record — neither part alone does it, and YARD's own hash-type syntax (`{K => V}`) is unaffected, which is what points at RBS rather than YARD. The enclosing array literal the comment originally sat in is not required, nor `.freeze`, nor a trailing comma. Note the comment carries no `#:` marker, so whatever is reading it is doing so speculatively.

*Filed by Claude (Anthropic's Claude Code) on behalf of @apiology.*

apiology:
Fixes https://github.com/castwide/solargraph/issues/1300

An ordinary comment anywhere in a class body could silently remove that
class's superclass:

```ruby
class Sub < Base
  # [:b, { c: :d }]

  def call_it
    greet   # Unresolved call to greet
  end
end
```

Every inherited method then reports `Unresolved call`, including on lines
*above* the comment, so the reported locations point nowhere near the cause.

`parameters_from_inline_rbs` regex-scanned the whole class body for `#[...]`,
so any bracketed comment was read as the superclass's inline RBS type
arguments. `# [:b, { c: :d }]` yields the superclass name `Base<:b, { c: :d`,
which resolves to nothing.

The annotation is now only read where [ruby/rbs's inline
syntax](https://github.com/ruby/rbs/blob/master/docs/inline.md) defines it —
directly after the superclass, on the same line, `#[` with no space — and a
closing `]` is required.

Tightening worth noting: `class Foo < Array # [String]` (space after the `#`)
is now an ordinary comment rather than a type annotation.

Also removes a leftover debug `logger.warn` from
https://github.com/castwide/solargraph/pull/1173.

https://claude.ai/code/session_01NtwP1oKRXfJAbPQrPHm5oU

### DIFF
--- a/lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/namespace_node.rb
@@ -27,7 +27,6 @@ module Solargraph
               source: :parser
             )
             pins.push nspin
-            Solargraph.logger.warn "Superclass: #{superclass_name}" if superclass_name&.start_with?('Array')
             if superclass_name
               pins.push Pin::Reference::Superclass.new(
                 location: loc,
@@ -41,12 +40,22 @@ module Solargraph
 
           private
 
-          # @param comments [String]
+          # Type arguments for a generic superclass in inline RBS syntax, e.g.,
+          # `class Foo < Array #[String]`. Only recognized where RBS defines
+          # it: directly after the superclass, on the same line, with no space
+          # between `#` and `[`. Anything else is an ordinary comment.
+          #
           # @return [String, nil]
           def parameters_from_inline_rbs
-            source = region.source.code_for(node)
-            match = source.match(/[^\n]*?#\s?+\[([^\]]*)/)
-            return unless match && match[1]
+            superclass = node.children[1]
+            return unless superclass
+
+            source = region.source.code
+            pos = get_node_end_position(superclass)
+            offset = Position.line_char_to_offset(source, pos.line, pos.character)
+            eol = source.index("\n", offset) || source.length
+            match = source[offset...eol].to_s.match(/\A\s*#\[([^\]]*)\]/)
+            return unless match
 
             code = match[1].strip
             return if code.empty?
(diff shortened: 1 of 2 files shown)


castwide:
I think there's a more efficient way to match the superclass parameter in `parameters_from_inline_rbs`, but the impact is probably trivial.

PR #1308, opened 2026-08-16
Narrow three more flow-sensitive assignment shapes

apiology:
Stacked on #1282 — please review that one first; these are three of its neighbours.

#1282 and its follow-on narrow a parameter after a definite reassignment, and after a nil-guarded default (`x = default if x.nil?`) used past the conditional. Three adjacent shapes still failed. Each is reduced from a real suppression in a consuming codebase, and each is fixed in its own commit.

**1. Narrowing outlived a definite reassignment.** After a guarded `return` plants a downcast, an unconditional reassignment superseded the earlier pin's *assignments* but `combine_with` still unioned `narrowed_return_type`/`exclude_return_type`, so the old value's falsy residue outlived the value it described — `Unresolved call to length on nil, Boolean`. Those two are now taken from the superseding pin alone. Second half: `references_name?` counted a same-named **block parameter** as a self-reference (`find { |specish| specish.name == name }`), blocking the supersede; the walk now descends only into a shadowing block's receiver, which is evaluated outside the block.

**2. A dominating reassignment wasn't treated as definite.** `override_assignments?` already resolved the location-specific case via `definite_reaches?`, but that verdict only reached `combine_assignments` — the combined pin kept `definite: definite || other.definite`, false on both sides, so `Pin::Parameter#typify` fell back to the declared `@param` type. Sound because `ApiMap#var_at_location` is the only caller passing a `location`; without one, `override_assignments?` already requires `other.definite`.

**3. A variable assigned inside an `if` condition.** `if (md = name.match(/…/))` — `process_expression` handled neither the one-child `:begin` that parentheses produce nor `:lvasgn`/`:ivasgn`. Adding those handlers alone changed nothing, because `IfNode#process` ran `FlowSensitiveTyping` *before* processing the condition, so the pin the condition creates didn't exist yet and `find_var` returned nil. The call now runs after the condition.

Every fix ships with a negative control that must keep erroring, and all of them still do: a trailing `if other_thing`; a use site *between* guard and reassignment; a reassignment only in a nested branch; a use before the assignment; and `(md = …) || fallback`, which stays unnarrowed because `process_or` deliberately passes no true ranges to its operands.

Effect on this repo's own suppressions: **two `@sg-ignore` markers removed** from `workspace/gemspecs.rb` and four from `position.rb`, all confirmed Unneeded at strong level rather than deleted on faith. Whole-repo `typecheck --level strong` produces an identical sorted problem set to the base, every delta accounted for by line renumbering. Full suite green.

One marker I restored rather than removed: dropping the `to_spec` one in `gemspecs.rb` surfaced a live `Unresolved call to to_spec`, so its "Unneeded" report was itself wrong — worth knowing that the Unneeded signal is not always reliable.

Not addressed, flagged for a follow-up: `WhileNode#process` has the identical FST-before-condition ordering, so `while (x = f.gets)` still misses shape 3 when `x` has no earlier assignment.

*Authored by Claude (Anthropic's Claude Code) on behalf of @apiology.*

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -706,7 +706,6 @@ module Solargraph
       # @todo If two literals are different values of the same type, it would
       #   make more sense for super_and_sub? to return true, but there are a
       #   few callers that currently expect this to be false.
-      # @sg-ignore flow-sensitive typing should be able to handle redefinition
       return false if sup.literal? && sub.literal? && sup.to_s != sub.to_s
       # @sg-ignore flow sensitive typing should be able to handle redefinition
       sup = sup.simplify_literals.to_s
@@ -714,6 +713,7 @@ module Solargraph
       sub = sub.simplify_literals.to_s
       return true if sup == sub
       sc_fqns = sub
+      # @sg-ignore flow sensitive typing unions rather than overrides types across multiple sequential reassignments
       while (sc = store.get_superclass(sc_fqns))
         # @sg-ignore flow sensitive typing needs to handle "if foo = bar"
         sc_new = store.constants.dereference(sc)
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -416,7 +416,7 @@ module Solargraph
         !pin.visible_at?(closure, location) && !pin.starts_at?(location)
       end
 
-      vars_at_location.inject(&:combine_with)
+      vars_at_location.inject { |acc, pin| acc.combine_with(pin, location: location) }
     end
 
     # Get an array of class variable pins for a namespace.
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -706,11 +706,8 @@ module Solargraph
       # @todo If two literals are different values of the same type, it would
       #   make more sense for super_and_sub? to return true, but there are a
       #   few callers that currently expect this to be false.
-      # @sg-ignore flow sensitive typing unions rather than overrides types across multiple sequential reassignments
       return false if sup.literal? && sub.literal? && sup.to_s != sub.to_s
-      # @sg-ignore https://github.com/castwide/solargraph/pull/1282
       sup = sup.simplify_literals.to_s
-      # @sg-ignore https://github.com/castwide/solargraph/pull/1282
       sub = sub.simplify_literals.to_s
       return true if sup == sub
       sc_fqns = sub
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -706,14 +706,14 @@ module Solargraph
       # @todo If two literals are different values of the same type, it would
       #   make more sense for super_and_sub? to return true, but there are a
       #   few callers that currently expect this to be false.
+      # @sg-ignore flow sensitive typing unions rather than overrides types across multiple sequential reassignments
       return false if sup.literal? && sub.literal? && sup.to_s != sub.to_s
-      # @sg-ignore flow sensitive typing should be able to handle redefinition
+      # @sg-ignore https://github.com/castwide/solargraph/pull/1282
       sup = sup.simplify_literals.to_s
-      # @sg-ignore flow sensitive typing should be able to handle redefinition
+      # @sg-ignore https://github.com/castwide/solargraph/pull/1282
       sub = sub.simplify_literals.to_s
       return true if sup == sub
       sc_fqns = sub
-      # @sg-ignore flow sensitive typing unions rather than overrides types across multiple sequential reassignments
       while (sc = store.get_superclass(sc_fqns))
         # @sg-ignore flow sensitive typing needs to handle "if foo = bar"
         sc_new = store.constants.dereference(sc)
--- a/lib/solargraph/language_server/message/text_document/formatting.rb
+++ b/lib/solargraph/language_server/message/text_document/formatting.rb
@@ -48,7 +48,6 @@ module Solargraph
             return if corrections&.empty?
 
             Solargraph.logger.info('Formatting result:')
-            # @sg-ignore flow sensitive typing should be able to handle redefinition
             corrections.each_line do |line|
               next if line.strip.empty?
               Solargraph.logger.info(line.strip)
--- a/lib/solargraph/language_server/message/text_document/formatting.rb
+++ b/lib/solargraph/language_server/message/text_document/formatting.rb
@@ -48,6 +48,7 @@ module Solargraph
             return if corrections&.empty?
 
             Solargraph.logger.info('Formatting result:')
+            # @sg-ignore flow sensitive typing should be able to handle redefinition
             corrections.each_line do |line|
               next if line.strip.empty?
               Solargraph.logger.info(line.strip)
--- a/lib/solargraph/parser/parser_gem/node_processors/if_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/if_node.rb
@@ -30,12 +30,13 @@ module Solargraph
                 location: get_node_location(then_node),
                 closure: region.closure,
                 compound_statement: region.compound_statement,
+                conditional: true,
                 node: then_node,
                 source: :parser
               )
               pins.push then_cs
               # @sg-ignore Need to add nil check here
-              NodeProcessor.process(then_node, region.update(compound_statement: then_cs, conditional: true), pins, locals, ivars)
+              NodeProcessor.process(then_node, region.update(compound_statement: then_cs), pins, locals, ivars)
             end
 
             else_node = node.children[2]
@@ -45,12 +46,13 @@ module Solargraph
                 location: get_node_location(else_node),
                 closure: region.closure,
                 compound_statement: region.compound_statement,
+                conditional: true,
                 node: else_node,
                 source: :parser
               )
               pins.push else_cs
               # @sg-ignore Need to add nil check here
-              NodeProcessor.process(else_node, region.update(compound_statement: else_cs, conditional: true), pins, locals, ivars)
+              NodeProcessor.process(else_node, region.update(compound_statement: else_cs), pins, locals, ivars)
             end
 
             true
--- a/lib/solargraph/parser/parser_gem/node_processors/if_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/if_node.rb
@@ -22,8 +22,6 @@ module Solargraph
               )
               NodeProcessor.process(condition_node, region, pins, locals, ivars)
             end
-            conditional_region = region.update(conditional: true)
-
             then_node = node.children[1]
             if then_node
               pins.push Solargraph::Pin::CompoundStatement.new(
@@ -32,7 +30,8 @@ module Solargraph
                 node: then_node,
                 source: :parser
               )
-              NodeProcessor.process(then_node, conditional_region, pins, locals, ivars)
+              # @sg-ignore Need to add nil check here
+              NodeProcessor.process(then_node, region.update(conditional_boundary: Range.from_node(then_node)), pins, locals, ivars)
             end
 
             else_node = node.children[2]
@@ -43,7 +42,8 @@ module Solargraph
                 node: else_node,
                 source: :parser
               )
-              NodeProcessor.process(else_node, conditional_region, pins, locals, ivars)
+              # @sg-ignore Need to add nil check here
+              NodeProcessor.process(else_node, region.update(conditional_boundary: Range.from_node(else_node)), pins, locals, ivars)
             end
 
             true
--- a/lib/solargraph/parser/parser_gem/node_processors/if_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/if_node.rb
@@ -35,7 +35,7 @@ module Solargraph
               )
               pins.push then_cs
               # @sg-ignore Need to add nil check here
-              NodeProcessor.process(then_node, region.update(conditional_boundary: Range.from_node(then_node), compound_statement: then_cs), pins, locals, ivars)
+              NodeProcessor.process(then_node, region.update(compound_statement: then_cs, conditional: true), pins, locals, ivars)
             end
 
             else_node = node.children[2]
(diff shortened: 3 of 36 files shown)


lib/solargraph/language_server/message/text_document/formatting.rb
```diff
@@ -48,6 +48,7 @@ def log_corrections corrections
             return if corrections&.empty?
 
             Solargraph.logger.info('Formatting result:')
+            # @sg-ignore flow sensitive typing should be able to handle redefinition
```
apiology: Where is the redefinition?

lib/solargraph/parser/parser_gem/node_processors/resbody_node.rb
```diff
@@ -30,7 +30,21 @@ def process
                 source: :parser
               )
             end
-            NodeProcessor.process(node.children[2], region, pins, locals, ivars)
+            # not pushed onto `pins` - and/or/orasgn/resbody bodies are
+            # too common to warrant a pin per occurrence, so only the
+            # pointer is needed for the compound_statement chain
+            # @sg-ignore Need to add nil check here
+            rescue_body_cs = Solargraph::Pin::CompoundStatement.new(
+              # @sg-ignore Need to add nil check here
```
apiology: So add them

lib/solargraph/pin/local_variable.rb
```diff
@@ -16,9 +16,9 @@ def probe api_map
         super
       end
 
-      def combine_with other, attrs = {}
+      def combine_with other, attrs = {}, location: nil
         # keep this as a parameter
-        return other.combine_with(self, attrs) if other.is_a?(Parameter) && !is_a?(Parameter)
+        return other.combine_with(self, attrs, location: location) if other.is_a?(Parameter) && !is_a?(Parameter)
```
apiology: Why is location passed in specifically rather than being pulled from self or other?

lib/solargraph/pin/parameter.rb
```diff
@@ -45,7 +45,7 @@ def combine_with other, attrs = {}
                         asgn_code: asgn_code
                       }
                     end
-        super(other, new_attrs.merge(attrs))
+        super(other, new_attrs.merge(attrs), location: location)
```
apiology: ditto

lib/solargraph/pin/parameter.rb
```diff
@@ -208,6 +208,15 @@ def index
 
       # @param api_map [ApiMap]
       def typify api_map
+        if definite
+          # flow sensitive typing: this parameter was reassigned by
+          # an assignment guaranteed to have executed, so prefer the
+          # type of the value it was reassigned to over its declared
+          # @param type
```
apiology: don't put YARD tags in comments unless you mean them

lib/solargraph/api_map.rb
```diff
@@ -706,14 +706,14 @@ def super_and_sub? sup, sub
...
       #   few callers that currently expect this to be false.
-      # @sg-ignore flow-sensitive typing should be able to handle redefinition
       return false if sup.literal? && sub.literal? && sup.to_s != sub.to_s
       # @sg-ignore flow sensitive typing should be able to handle redefinition
       sup = sup.simplify_literals.to_s
       # @sg-ignore flow sensitive typing should be able to handle redefinition
       sub = sub.simplify_literals.to_s
       return true if sup == sub
       sc_fqns = sub
+      # @sg-ignore flow sensitive typing unions rather than overrides types across multiple sequential reassignments
```
apiology: Do we have an open PR addressing this?  If yes, use that as the sg-ignore text

lib/solargraph/type_checker.rb
```diff
@@ -652,6 +652,7 @@ def add_to_param_details param_details, param_names, new_param_details
     # @return [Hash{String => Hash{Symbol => String, ComplexType}}]
     def param_details_from_stack signature, method_pin_stack
       signature_type = signature.typify(api_map)
+      # @sg-ignore flow sensitive typing should be able to handle redefinition
```
apiology: Reference open PR URL

lib/solargraph/language_server/message/text_document/formatting.rb
apiology: **Claude:**

Nowhere, as it turns out - and the marker is gone. 6b1240110 removed it, and `formatting.rb` now has no diff against master on this branch.

lib/solargraph/parser/parser_gem/node_processors/resbody_node.rb
apiology: **Claude:**

Added in 6b1240110. Both placeholder markers are gone and the nil check is explicit:

```ruby
rescue_body_node = node.children[2]
rescue_body_cs = Solargraph::Pin::CompoundStatement.new(
  location: rescue_body_node ? get_node_location(rescue_body_node) : nil,
  closure: region.closure,
  compound_statement: region.compound_statement,
  conditional: true,
  node: rescue_body_node,
  source: :parser
)
NodeProcessor.process(rescue_body_node, region.update(compound_statement: rescue_body_cs), pins, locals, ivars) if rescue_body_node
```

lib/solargraph/pin/local_variable.rb
apiology: **Claude:**

`location` is the position being queried, not a property of either pin, so neither side can supply it. It arrives from `ApiMap#var_at_location`, whose own `location` parameter is the cursor the type is being resolved for:

```ruby
vars_at_location.inject { |acc, pin| acc.combine_with(pin, location: location) }
```

`BaseVariable#combine_with` hands it to `override_assignments?` and on to `definite_reaches?`, which asks whether that *reference site* falls inside `other`'s `compound_statement` range. A pin's own `location` is where it was assigned, so `other.location` would instead test whether one assignment sits inside the other's branch - a different question, and not the one that decides whether a conditional assignment dominates the use.

The same two pins have to combine differently depending only on where the value is read:

```ruby
x = 1
if cond
  x = 'foo'   # not globally definite
  x.upcase    # location inside the branch: other overrides
end
x.upcase      # location outside it: union
```

The reasoning is written up on `BaseVariable#combine_with`'s `@param location` tag in this PR; `LocalVariable#combine_with` only forwards.

lib/solargraph/pin/parameter.rb

Same answer as on `LocalVariable#combine_with`: `location` is the query position from `ApiMap#var_at_location`, and `Parameter#combine_with` only forwards it to `super`.

lib/solargraph/pin/parameter.rb

Fixed in 6b1240110 - the comment no longer names a tag:

```ruby
# Reassigned by an assignment guaranteed to have executed:
# prefer that value's type over the declared parameter type.
```

lib/solargraph/api_map.rb

No open PR or issue covers it - and it turned out not to matter, because all three `@sg-ignore` markers in `super_and_sub?` were unneeded. Deleted in 21ab81a11.

CI at dcc707b4c flagged every one of them:

```
api_map.rb:710: Unneeded @sg-ignore comment
api_map.rb:712: Unneeded @sg-ignore comment
api_map.rb:714: Unneeded @sg-ignore comment
```

This branch's own dominance handling resolves the redefinition case they covered, so neither the prose marker nor the two PR 1282 links suppress anything any more.

Worth recording, because it produced a wrong conclusion first: `bin/solargraph typecheck` reports all three as still needed. `bin/solargraph` is a bare script rather than a bundler binstub, so `require 'solargraph'` resolves to the installed 0.60.4 gem instead of `lib/` - an analyzer predating this work. `bundle exec solargraph typecheck --level strong` agrees with CI.

One gap in this method does survive, unsuppressed:

```
api_map.rb:717: Wrong argument type for Solargraph::ApiMap::Store#get_superclass:
fqns expected String, nil, received Solargraph::ComplexType, String
```

`sc_fqns` is `ComplexType, String` there because multiple sequential reassignments union rather than dominate by recency - the shape the deleted wording described. It counts toward the 547 today. Whether to suppress it is a separate call and was not made here.

lib/solargraph/type_checker.rb
apiology: **Claude:**

That one turned out unneeded, so it was dropped in 6b1240110 rather than pointed at a URL. `type_checker.rb` now has no diff against master on this branch.

lib/solargraph/parser/parser_gem/node_processors/if_node.rb
```diff
@@ -8,40 +8,54 @@ class IfNode < Parser::NodeProcessor::Base
...
+            # itself assigns (`if (md = foo.match(...))`) already has a
+            # pin to assert facts about
+            FlowSensitiveTyping.new(locals,
+                                    ivars,
+                                    enclosing_breakable_pin,
+                                    enclosing_compound_statement_pin).process_if(node)
             then_node = node.children[1]
             if then_node
-              pins.push Solargraph::Pin::CompoundStatement.new(
+              # @sg-ignore Need to add nil check here
```
apiology: If you need a nil check, add it now instead of deferring the addition.  Apply elsewhere as well

lib/solargraph/parser/parser_gem/node_processors/lvasgn_node.rb
```diff
@@ -19,6 +19,8 @@ def process
               assignment: node.children[1],
               comments: comments_for(node),
               presence: presence,
+              definite: !region.compound_statement.conditional,
```
apiology: Why not standardize on one predicate - definite or !conditional?

lib/solargraph/parser/parser_gem/node_processors/if_node.rb
apiology: **Claude:**

Done in bfb10e793. This branch added six `Need to add nil check here` markers; none survive.

- `FlowSensitiveTyping#process_if` now guards `conditional_node` before use. CI reported `expected Parser::AST::Node, received Parser::AST::Node, nil` on the `process_guarded_reassignment` call, and the same nil reached `process_expression` one line above, unsuppressed.
- `Pin::Method#infer_from_return_nodes` now calls `Pin::Base#filename`, which already returns nil when `location` is nil, instead of reaching through `location.filename`. That removed two markers, one of which predates this branch.
- The four here in `IfNode` are deleted outright rather than fixed: CI reports all four as `Unneeded @sg-ignore comment`, at `if_node.rb:32`, `42`, `48` and `58`. The `if then_node` / `if else_node` guards were already present.

One thing that cost a couple of wrong attempts, worth knowing before touching `infer_from_return_nodes`: `location` is very often non-nil while `location.filename` is nil, and the loop relies on passing that nil straight to `ApiMap#source_map`. Guarding on the filename rather than on the location returns `undefined` and breaks three `method_spec` examples.

Also unexplained, and it changes what the markers here look like: on my machine `node.children[N]` infers as `Array`, so these four lines report `expected Parser::AST::Node, received Array` and the markers read as needed. CI infers `Parser::AST::Node, nil` for the identical source. I took CI as authoritative and deleted them; if the next run turns them back into real errors rather than clearing them, that inference gap is the reason.

lib/solargraph/parser/parser_gem/node_processors/lvasgn_node.rb
apiology: **Claude:**

They are attributes of two different pins with different lifetimes, so one cannot replace the other.

`conditional` belongs to `Pin::CompoundStatement` and describes a branch: this body may not run. `definite` belongs to `Pin::BaseVariable` and describes an assignment: this one is guaranteed to have run. `lvasgn_node.rb` is the one place that converts the first into the second.

Collapsing to `!compound_statement.conditional` at the point of use would be wrong in both directions:

- `args_node.rb:30` sets `definite: false` for a default-value assignment that sits in no conditional compound statement at all - the derived form would call it definite.
- `BaseVariable#combine_with` promotes it: `definite: definite || other.definite || superseded`. A pin can become definite by being superseded by a later assignment, with its own compound statement unchanged.

Going the other way and dropping `conditional` in favour of `definite` does not work either, since a `CompoundStatement` has no assignment to be definite about.

The one predicate that genuinely bridges them is `definite_reaches?`, which asks whether a given reference falls inside the compound statement of a non-definite assignment - and that needs both.

apiology: Please add a one line comment above this line on its meaning

lib/solargraph/pin/base.rb
```diff
@@ -75,6 +75,7 @@ def assert_location_provided
 
       # @return [Pin::Closure, nil]
       def closure
+        @closure ||= derive_closure_from_compound_statement
```
apiology: What benefit does adding this bring?

lib/solargraph/parser/parser_gem/node_processors/lvasgn_node.rb
apiology: **Claude:**

Added in c1e692b88:

```ruby
# false inside an if/while/rescue body, which may not run
definite: !region.compound_statement.conditional,
```

lib/solargraph/pin/base.rb
apiology: **Claude:**

None that I can measure, so it is removed in c1e692b88 along with the private `derive_closure_from_compound_statement` it called.

All eight `CompoundStatement.new` sites in `lib/` pass `closure:` explicitly, so the derivation never fires in production. With the line removed, the only failing example out of 1685 was the one added alongside it to exercise it - its sole consumer was its own test, so both are gone. The method's two `@sg-ignore flow sensitive typing doesn't narrow self past an is_a? guard` markers went with it.

There is a second benefit to removing it: `#closure` now infers more precisely without the `Pin::Closure, nil` fallback in the method body, which takes the local strong typecheck from 576 problems to 543.

The two remaining examples in `compound_statement_spec` still earn their place - they walk the chain with their own helper and assert it reaches the stored closure, so a node processor that threads `closure:` without `compound_statement:` is still caught. Their header comment described a derivation that no longer exists and has been rewritten.

apiology: **Claude:**

Correcting one claim in my reply above. I said removing the derivation tightens what `#closure` infers, citing a local strong typecheck drop from 576 problems to 543. CI disagrees: 539 both before and after this commit, so there is no measurable inference benefit - only the local run moved.

That is the same local/CI divergence already visible on this branch, where `node.children[N]` infers as `Array` here and `Parser::AST::Node, nil` in CI. Cause still unknown.

The rest of the reply stands, and it is the part that answers the question: the derivation had no production consumer, all eight `CompoundStatement.new` sites pass `closure:`, and its only failing test on removal was the one written to exercise it. CI is green at c1e692b88 with no new problems in `pin/base.rb` - the four that remain there are pre-existing YARD tag and directive ones.

apiology:
**Claude:** Closed in favour of #1338, which combines this with #1282 into one branch.

PR #1310, opened 2026-08-16
Rebind self in Class.new blocks; nested blocks inherit rebound binders

apiology:
Solargraph's strong typecheck reports `Unresolved call` for methods that work at runtime inside a `Class.new` block:

```ruby
Class.new do
  define_method(:foo) { nil }                     # Unresolved call to define_method
  %i[a b].each { |m| define_method(m) { nil } }   # same, one block deeper
end
```

Two changes:

1. **Core fill**: `Class#new` gets `@yieldreceiver [::Class]`. RBS binds only `Class#initialize` (`[self: Class]`, rbs >= 4.1); `Class#new` is `(*untyped, **untyped) -> untyped` in every version, so a `[self: ...]` translator fix never reaches this call site. `@yieldreceiver` can't express the superclass parameter, so the fill stays unparameterized.
2. **`Pin::Block#rebind`/`#binder` cascade**: a block with no rebind of its own inherits its enclosing block's binder. Without it, every existing fill (`class_eval`, ...) stops one nesting level down.

Three new strong-level specs (direct, `.each`-nested, `class_eval`-nested).

🤖 This PR was written by Claude (Claude Code), operated by @apiology.

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -852,11 +852,7 @@ module Solargraph
       # class' responsibility
       methods = store.get_methods(fqns, scope: scope, visibility: visibility).sort { |a, b| a.name <=> b.name }
       if fqns == 'Object' && scope == :instance && visibility.include?(:private)
-        # Methods defined at the top level are private instance methods
-        # of Object at runtime, but are stored under the root namespace,
-        # which is not otherwise part of Object's ancestry. Only requests
-        # that admit private methods see them, which is how a receiverless
-        # call reaches them but 'str'.top_level_helper does not.
+        # Top-level defs are private instance methods of Object at runtime, but stored under the root namespace.
         root_reqstr = "|#{scope}|#{visibility.sort}|#{deep}"
         unless skip.include?(root_reqstr)
           skip.add root_reqstr
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -851,6 +851,18 @@ module Solargraph
       # namespaces; resolving the generics in the method pins is this
       # class' responsibility
       methods = store.get_methods(fqns, scope: scope, visibility: visibility).sort { |a, b| a.name <=> b.name }
+      if fqns == 'Object' && scope == :instance && visibility.include?(:private)
+        # Methods defined at the top level are private instance methods
+        # of Object at runtime, but are stored under the root namespace,
+        # which is not otherwise part of Object's ancestry. Only requests
+        # that admit private methods see them, which is how a receiverless
+        # call reaches them but 'str'.top_level_helper does not.
+        root_reqstr = "|#{scope}|#{visibility.sort}|#{deep}"
+        unless skip.include?(root_reqstr)
+          skip.add root_reqstr
+          methods += store.get_methods('', scope: :instance, visibility: visibility).sort { |a, b| a.name <=> b.name }
+        end
+      end
       logger.info do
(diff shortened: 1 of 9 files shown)


lib/solargraph/pin/block.rb
```diff
@@ -28,12 +28,22 @@ def initialize receiver: nil, args: [], context: nil, node: nil, **splat
       # @param api_map [ApiMap]
       # @return [void]
       def rebind api_map
-        @rebind ||= maybe_rebind(api_map)
+        @rebind ||= begin
+          # An enclosing block's rebind (e.g. a class_eval receiver)
+          # also applies to this block unless overridden here
+          enclosing = closure
+          enclosing.rebind(api_map) if enclosing.is_a?(Block)
+          maybe_rebind(api_map)
```
apiology, line 33: Why here and not in maybe_rebind?

apiology: **Claude:**

`maybe_rebind` only computes this block's own rebind from its own
`receiver`/`@yieldreceiver` tag - it returns `ComplexType::UNDEFINED` when
this block has no receiver of its own (e.g. a bare nested `do...end` with
nothing to key off). It has no knowledge of enclosing blocks at all.

The enclosing-block walk in `rebind` isn't about *applying* the enclosing
block's rebind to this one - `binder` handles that, by falling through to
`enclosing.binder` when this block's own `@rebind` isn't defined. What
`rebind` does by calling `enclosing.rebind(api_map)` first is force that
ancestor's `@rebind` memo to populate *before* `binder`/`context` read it,
since those are the fallback paths `binder` depends on.

So the two methods are doing different jobs: `maybe_rebind` computes this
block's own receiver-based rebind, and `rebind` is the memoization/ancestor
walk that ensures the whole chain's caches are warm before anything reads
them. Folding the ancestor walk into `maybe_rebind` would mix those two
concerns into one method.

apiology: How about extract these two lines to a method with a name that makes it clear?  You can have at most 2 lines of documentation above the new method please.

apiology: **Claude:** Extracted in 533f2aa0a as `warm_enclosing_rebind_cache`, with a 2-line docstring above it. `rebind` now just calls it then `maybe_rebind(api_map)`.

lib/solargraph/source/chain/call.rb
```diff
@@ -71,6 +71,26 @@ def resolve api_map, name_pin, locals
 
         private
 
+        # Which visibilities a call at this position can reach. A
+        # receiverless call - head? - reaches private and protected
+        # methods whatever self is. A call with an explicit receiver
+        # only reaches them from inside the receiver's own namespace or
+        # a subclass of it, mirroring ApiMap#get_complex_type_methods.
```
apiology: 2 lines at most

lib/solargraph/api_map.rb
```diff
@@ -851,6 +851,18 @@ def inner_get_methods rooted_tag, scope, visibility, deep, skip, no_core = false
       # namespaces; resolving the generics in the method pins is this
       # class' responsibility
       methods = store.get_methods(fqns, scope: scope, visibility: visibility).sort { |a, b| a.name <=> b.name }
+      if fqns == 'Object' && scope == :instance && visibility.include?(:private)
+        # Methods defined at the top level are private instance methods
+        # of Object at runtime, but are stored under the root namespace,
+        # which is not otherwise part of Object's ancestry. Only requests
+        # that admit private methods see them, which is how a receiverless
+        # call reaches them but 'str'.top_level_helper does not.
```
apiology: 1 line at most

PR #1311, opened 2026-08-17
Map Forwardable delegations to method pins

apiology:
## Problem

`def_delegators` and `def_delegator` define real methods, but Solargraph maps neither, so every call to a delegated method is unresolved and any method returning one cannot infer its return type:

```ruby
require 'forwardable'

class Records
  # @return [Integer]
  def size = 3
end

class Holder
  extend Forwardable

  def_delegators :@records, :size

  # @return [void]
  def initialize
    # @type [Records]
    @records = Records.new
  end

  # @return [Integer]
  def total = size
end
```

```
test.rb:20: Holder#total return type could not be inferred
test.rb:20: Unresolved call to size
```

Separately, `alias_method` in a class body reports a false positive at strong level today:

```ruby
class Widget
  # @return [String]
  def name = 'x'

  alias_method :title, :name
end
```

```
test.rb:5: Unresolved call to alias_method
```

## Solution

`SendNode` maps both Forwardable forms into `Pin::DelegatedMethod` — the pin class added in #602, until now constructed only by external plugins — which resolves the receiver and the forwarded method lazily, so a delegated method reports the forwarded method's parameters, return type and signatures, and passes arguments and blocks through. Instance-variable, class-variable and method receivers are supported, as is `def_delegator`'s optional alias argument. Expansion fires only in a namespace that extends `Forwardable`; `SingleForwardable`'s class-scope delegation is not covered.

The `alias_method` false positive shares one cause with what would otherwise have been a `def_delegators` false positive: method pins synthesized from a class-body statement cover that statement's range but have no body of their own, so `_locate_pin` treating them as closures makes the declaring statement resolve in instance scope, where `alias_method` and `def_delegators` are not available. `_locate_pin` already skipped attribute pins for this reason, with a `@todo` noting the hack; aliases and delegations now answer the same `Pin::Method#body_less?` predicate.

## Test plan

Six strong-level specs — ivar receiver, method receiver, aliased delegation, argument and block passthrough, no expansion without `extend Forwardable`, and no false positive on the declaring statement. Five fail without this change. Full suite: 1541 examples, 0 failures, 59 pending.

*Opened as a draft. This PR was written by Claude (Anthropic's Claude Code) on behalf of @apiology.*

### DIFF
--- a/lib/solargraph/doc_map.rb
+++ b/lib/solargraph/doc_map.rb
@@ -268,6 +268,19 @@ module Solargraph
         logger.debug { "Using only RBS collection pins for #{gemspec.name}:#{gemspec.version}" }
         combined_pins_in_memory[[gemspec.name, gemspec.version]] = rbs_collection_pins
         combined_pins_in_memory[[gemspec.name, gemspec.version]]
+      elsif (fallback_pins = rbs_map.fallback_pins)
+        # rbs_map's own stdlib cache (stdlib/#{gemspec.name}.ser) already
+        # has these pins - built and serialized by RbsMap::StdlibMap#initialize
+        # a few lines above. What's missing is a *combined* cache entry
+        # for this gemspec, which only `solargraph gems` writes; until
+        # that runs, this method has nowhere else to look.
+        #
+        # Deliberately not written to combined_pins_in_memory: that is
+        # process-wide and keyed only by name and version, so a
+        # provisional set stored there would go on being served after
+        # the build that supersedes it.
+        logger.debug { "Using #{gemspec.name}:#{gemspec.version}'s stdlib RBS pins (no combined cache entry yet)" }
+        fallback_pins
       else
         logger.debug { "Pins not yet cached for #{gemspec.name}:#{gemspec.version}" }
         nil
--- a/lib/solargraph/doc_map.rb
+++ b/lib/solargraph/doc_map.rb
@@ -268,19 +268,6 @@ module Solargraph
         logger.debug { "Using only RBS collection pins for #{gemspec.name}:#{gemspec.version}" }
         combined_pins_in_memory[[gemspec.name, gemspec.version]] = rbs_collection_pins
         combined_pins_in_memory[[gemspec.name, gemspec.version]]
-      elsif (fallback_pins = rbs_map.fallback_pins)
-        # rbs_map's own stdlib cache (stdlib/#{gemspec.name}.ser) already
-        # has these pins - built and serialized by RbsMap::StdlibMap#initialize
-        # a few lines above. What's missing is a *combined* cache entry
-        # for this gemspec, which only `solargraph gems` writes; until
-        # that runs, this method has nowhere else to look.
-        #
-        # Deliberately not written to combined_pins_in_memory: that is
-        # process-wide and keyed only by name and version, so a
-        # provisional set stored there would go on being served after
-        # the build that supersedes it.
-        logger.debug { "Using #{gemspec.name}:#{gemspec.version}'s stdlib RBS pins (no combined cache entry yet)" }
-        fallback_pins
       else
         logger.debug { "Pins not yet cached for #{gemspec.name}:#{gemspec.version}" }
         nil
--- a/lib/solargraph/parser/parser_gem/node_processors/send_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/send_node.rb
@@ -59,6 +59,7 @@ module Solargraph
 
           # True when the enclosing namespace extends Forwardable, which is
           # what makes def_delegator and def_delegators available there.
+          # Matched by name only, so a same-named local module would fool this.
           #
           # @return [Boolean]
           def extends_forwardable?
--- a/lib/solargraph/parser/parser_gem/node_processors/send_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/send_node.rb
@@ -40,7 +40,7 @@ module Solargraph
                 process_private_constant
               elsif method_name == :alias_method && node.children[2] && node.children[2] && node.children[2].type == :sym && node.children[3] && node.children[3].type == :sym
                 process_alias_method
-              elsif %i[def_delegator def_delegators].include?(method_name) && forwardable_extended?
+              elsif %i[def_delegator def_delegators].include?(method_name) && extends_forwardable?
                 process_def_delegators
               elsif method_name == :private_class_method && node.children[2].is_a?(AST::Node)
                 # Processing a private class can potentially handle children on its own
@@ -61,7 +61,7 @@ module Solargraph
           # what makes def_delegator and def_delegators available there.
           #
           # @return [Boolean]
-          def forwardable_extended?
+          def extends_forwardable?
             pins.any? do |pin|
               pin.is_a?(Pin::Reference::Extend) &&
                 pin.closure == region.closure &&
@@ -76,10 +76,15 @@ module Solargraph
           def symbol_word subject
             return nil unless subject.is_a?(::Parser::AST::Node)
 
-            # @sg-ignore is_a? guard above is not narrowed through a return
+            # @sg-ignore is_a? guard above narrows subject to a
+            # workspace-local type but not to an external gem-sourced
+            # type like Parser::AST::Node - "Unresolved call to type on
+            # Object" without this. No upstream issue filed yet.
             return nil unless %i[sym str].include?(subject.type)
 
-            # @sg-ignore is_a? guard above is not narrowed through a return
+            # @sg-ignore Same is_a?-vs-external-type narrowing gap as
+            # above - "Unresolved call to children on Object" without
+            # this. No upstream issue filed yet.
             subject.children.first.to_s
           end
 
--- a/lib/solargraph/parser/parser_gem/node_processors/send_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/send_node.rb
@@ -77,15 +77,10 @@ module Solargraph
           def symbol_word subject
             return nil unless subject.is_a?(::Parser::AST::Node)
 
-            # @sg-ignore is_a? guard above narrows subject to a
-            # workspace-local type but not to an external gem-sourced
-            # type like Parser::AST::Node - "Unresolved call to type on
-            # Object" without this. No upstream issue filed yet.
+            # @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
             return nil unless %i[sym str].include?(subject.type)
 
-            # @sg-ignore Same is_a?-vs-external-type narrowing gap as
-            # above - "Unresolved call to children on Object" without
-            # this. No upstream issue filed yet.
+            # @sg-ignore flow sensitive typing needs to narrow down type with an if is_a? check
             subject.children.first.to_s
           end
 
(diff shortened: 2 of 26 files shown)


lib/solargraph/doc_map.rb
```diff
@@ -268,6 +268,19 @@ def deserialize_combined_pin_cache gemspec
         logger.debug { "Using only RBS collection pins for #{gemspec.name}:#{gemspec.version}" }
         combined_pins_in_memory[[gemspec.name, gemspec.version]] = rbs_collection_pins
         combined_pins_in_memory[[gemspec.name, gemspec.version]]
+      elsif (fallback_pins = rbs_map.fallback_pins)
```
apiology: What's the purpose of assigning to a new local here?

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
```diff
@@ -55,6 +57,90 @@ def process
...
+          # @param subject [Object]
+          # @return [String, nil]
+          def symbol_word subject
+            return nil unless subject.is_a?(::Parser::AST::Node)
+
+            # @sg-ignore is_a? guard above is not narrowed through a return
+            return nil unless %i[sym str].include?(subject.type)
+
+            # @sg-ignore is_a? guard above is not narrowed through a return
+            subject.children.first.to_s
```
apiology: What's the actual error we get from typechecking

lib/solargraph/type_checker.rb
```diff
@@ -141,6 +141,12 @@ def method_tag_problems
     # @return [Array<Problem>]
     def method_return_type_problems_for pin
       return [] if pin.is_a?(Pin::MethodAlias)
+      # A def_delegators whose receiver doesn't resolve statically - one
+      # reaching a method_missing dispatcher, say - has no declaration site a
+      # @return tag could be written at. The type belongs to the delegation
+      # target, and where that can't be found the @!method directive written
+      # above the def_delegators supplies it instead.
```
apiology: Please cut the filler from this comment

lib/solargraph/pin/delegated_method.rb
```diff
@@ -5,6 +5,11 @@ module Pin
     # A DelegatedMethod is a more complicated version of a MethodAlias that
     # allows aliasing a method from a different closure (class/module etc).
     class DelegatedMethod < Pin::Method
+      # @return [Boolean]
+      def body_less?
+        true
+      end
```
apiology: Let's use "has_body?" and invert

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
```diff
@@ -55,6 +57,90 @@ def process
 
           private
 
+          # True when the enclosing namespace extends Forwardable, which is
+          # what makes def_delegator and def_delegators available there.
+          #
+          # @return [Boolean]
+          def forwardable_extended?
```
apiology: Maybe `extends_forwardable?`

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
apiology: **Claude:**

```
lib/solargraph/parser/parser_gem/node_processors/send_node.rb:80: Unresolved call to type on Object
lib/solargraph/parser/parser_gem/node_processors/send_node.rb:83: Unresolved call to children on Object
```

Isolated it to one variable: `is_a?` narrowing works for a workspace-local
class but not for an external gem-sourced one. This passes clean:

```ruby
module Foo
  class Node
    # @return [Symbol]
    def type; :sym; end
  end

  # @param subject [Object]
  # @return [Symbol, nil]
  def symbol_word(subject)
    return nil unless subject.is_a?(Node)

    subject.type
  end
end
```

This does not, with `Node` swapped for the real `::Parser::AST::Node`
(from the `parser` gem) and everything else identical:

```ruby
module Foo
  # @param subject [Object]
  # @return [Symbol, nil]
  def symbol_word(subject)
    return nil unless subject.is_a?(::Parser::AST::Node)

    subject.type
  end
end
```

-> `Unresolved call to type on Object`

No upstream issue currently tracks this. Recorded the literal errors on
both markers in 3f62fc8; whether to file an issue for it is on the
decision list I'm keeping in the Claude Code session.

lib/solargraph/type_checker.rb
apiology: **Claude:**

Trimmed in 3f62fc8:

```
- # A def_delegators whose receiver doesn't resolve statically - one
- # reaching a method_missing dispatcher, say - has no declaration site a
- # @return tag could be written at. The type belongs to the delegation
- # target, and where that can't be found the @!method directive written
- # above the def_delegators supplies it instead.
+ # A DelegatedMethod whose receiver doesn't resolve statically has no
+ # declaration site for a @return tag - the @!method directive above
+ # the def_delegators call supplies the type instead.
```

lib/solargraph/pin/delegated_method.rb
apiology: **Claude:**

Renamed `body_less?` to `has_body?` (inverted) across `Method`,
`DelegatedMethod`, and `MethodAlias`, and the `source_map.rb` call site,
in 3f62fc8. RuboCop's `Naming/PredicatePrefix` wants `body?` instead of
`has_body?`; disabled that cop inline on the three definitions rather than
rename away from what you asked for.

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
apiology: **Claude:**

Renamed `forwardable_extended?` to `extends_forwardable?` in 3f62fc8.

lib/solargraph/rbs_map.rb
```diff
@@ -153,6 +153,16 @@ def resolved?
       @resolved
     end
 
+    # Pins this map can hand back directly when a caller has nowhere else
+    # cached to look (no combined gem cache entry yet, no RBS collection
+    # entry). Base RbsMap has none; a subclass whose own pins are safe to
+    # serve standalone (see RbsMap::StdlibMap) overrides this.
```
apiology: 1 line at most

lib/solargraph/source_map.rb
```diff
@@ -221,9 +221,11 @@ def _locate_pin line, character, *klasses
       position = Position.new(line, character)
       found = nil
       pins.each do |pin|
-        # @todo Attribute pins should not be treated like closures, but
-        #   there's probably a better way to handle it
-        next if pin.is_a?(Pin::Method) && pin.attribute?
+        # Method pins synthesized from a class-body statement - attributes,
+        # aliases, delegations - cover that statement's range but have no
+        # body of their own, so treating them as closures would make the
+        # declaring statement resolve in the wrong scope.
```
apiology: 1 line at most

lib/solargraph/type_checker.rb
```diff
@@ -141,6 +141,10 @@ def method_tag_problems
     # @return [Array<Problem>]
     def method_return_type_problems_for pin
       return [] if pin.is_a?(Pin::MethodAlias)
+      # A DelegatedMethod whose receiver doesn't resolve statically has no
+      # declaration site for a @return tag - the @!method directive above
+      # the def_delegators call supplies the type instead.
```
apiology: 1 line at most

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
```diff
@@ -55,6 +57,95 @@ def process
...
 
+          # True when the enclosing namespace extends Forwardable, which is
+          # what makes def_delegator and def_delegators available there.
+          #
+          # @return [Boolean]
+          def extends_forwardable?
+            pins.any? do |pin|
+              pin.is_a?(Pin::Reference::Extend) &&
+                pin.closure == region.closure &&
+                ['Forwardable', '::Forwardable'].include?(pin.name)
```
apiology: Acknowledge in a comment that we could be fooled by a more local module named Forwardable please

lib/solargraph/pin/method_alias.rb
```diff
@@ -7,6 +7,12 @@ module Pin
     # methods from included modules.
     #
     class MethodAlias < Method
+      # @return [Boolean]
+      # rubocop:disable Naming/PredicatePrefix
+      def has_body?
+        false
+      end
+      # rubocop:enable Naming/PredicatePrefix
```
apiology: How about just body?

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
```diff
@@ -55,6 +57,96 @@ def process
...
+            # @sg-ignore is_a? guard above narrows subject to a
+            # workspace-local type but not to an external gem-sourced
+            # type like Parser::AST::Node - "Unresolved call to type on
+            # Object" without this. No upstream issue filed yet.
+            return nil unless %i[sym str].include?(subject.type)
+
+            # @sg-ignore Same is_a?-vs-external-type narrowing gap as
+            # above - "Unresolved call to children on Object" without
+            # this. No upstream issue filed yet.
+            subject.children.first.to_s
```
apiology: Look to see if you see an appropriate line in rules.rb matching this sg-ignore issue; if not create one and log there.  One line only.

lib/solargraph/rbs_map.rb
apiology: **Claude:** Done in 92a30e273, now one line:

`# Pins a subclass can serve standalone with no gem cache entry yet; see RbsMap::StdlibMap.`

lib/solargraph/source_map.rb
apiology: **Claude:** Done in 92a30e273, now one line:

`# A bodyless method pin covers its declaring statement's range but isn't a closure.`

lib/solargraph/type_checker.rb
apiology: **Claude:** Done in 92a30e273, now one line:

`# An unresolvable DelegatedMethod's type comes from its @!method directive, not a @return tag.`

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
apiology: **Claude:** Added in 92a30e273, as the third line of the docstring:

`# Matched by name only, so a same-named local module would fool this.`

lib/solargraph/pin/method_alias.rb
apiology: **Claude:** Renamed to `body?` in 92a30e273. That also dropped the `Naming/PredicatePrefix` disable/enable pair the `has_body?` name needed, and the call site in `SourceMap#_locate_pin` now reads `next if pin.is_a?(Pin::Method) && !pin.body?`.

lib/solargraph/parser/parser_gem/node_processors/send_node.rb
apiology: **Claude:** `rules.rb` already had a matching entry, so no new one was needed:

`# @todo 21: flow sensitive typing needs to narrow down type with an if is_a? check`

Both markers now read as that single line, in 32e451e8d. It is the same gap: `node_methods.rb` uses that entry 15 times for the identical shape, guarding `.type` and `.children` calls on a `::Parser::AST::Node` behind a guard that does not narrow. Bumped the entry from 19 to 21 and its group total from 96 to 98.

PR #1312, opened 2026-08-17
pin CLI: resolve paths through method lookup, unsilence messages, fix bare arrows

apiology:
## Problem

Three `solargraph pin` behaviors made diagnosis harder than it needs to be.

**A path that names no pin of its own described nothing**, even when Ruby would happily call the method. Given:

```ruby
module Mixin
  # @return [String]
  def helper
    'from mixin'
  end
end

class Child
  include Mixin
end
```

```
$ solargraph pin 'Child#helper'
$ echo $?
1
```

Nothing printed at all — which is the second bug. `bin/solargraph` sets `$VERBOSE = nil` to silence Ruby diagnostics, and that also turns `Kernel#warn` into a no-op, so the command's "Pin not found" message never reached the user. The same suppression applied to every warn-based CLI message: gem-not-found and caching progress in `gems`/`cache`/`uncache`, scan errors, and the invalid-`SOLARGRAPH_LOG` notice.

**An unparameterized `generic` rendered no return type at all**, because `UniqueType#to_rbs` returns nil for that tag and callers interpolate it away. The result is not valid RBS — `RBS::Parser` raises `ParsingError` on it:

```
$ solargraph pin 'Container#items' --rbs
def items: () ->
```

## Solution

1. **Resolve the path through Ruby method lookup by default**, describing the definition a call would actually reach:

   ```
   $ solargraph pin 'Child#helper'
   #<Solargraph::Pin::Method `name="helper" return_type=Mixin#helper def helper: () -> String, context=Mixin, closure="Mixin", binder=Mixin` at ...>
   $ echo $?
   0
   ```

   A new `--resolve` (default true, with Thor's `--no-resolve`) controls it; `--no-resolve` restores exact-path-only lookup. Lookup runs only when the path names no pin of its own, so output for a path that resolves exactly is byte-identical to before — verified across method and namespace paths against this base. Nothing extra is printed when resolution happens: the description names its own owner (`closure="Mixin"` above), so stdout stays a single clean record and `--rbs` still emits one parseable signature. Resolved-via-lookup exits 0; nothing found anywhere still exits 1. `--stack` already walks the ancestry and lists the whole chain, so it supersedes resolution.

2. **Print user-facing messages via `$stderr.puts`**, which `$VERBOSE = nil` cannot silence — the pin miss plus 13 other sites. One developer-diagnostic dump in `Pin::Base#choose` moves to `logger.warn` instead. `Style/StderrPuts` is already disabled repo-wide citing this hazard.

3. **Render an unparameterized `generic` as `untyped`**, which restores valid output: `def items: () -> untyped` parses as RBS; the bare arrow does not.

## Test plan

Eight new specs. They pin that an exact-path hit is byte-identical and never consults method lookup, that resolution happens by default and prints only the pin, that `--no-resolve` restores the old failure, that a total miss still exits 1, that pin-miss and gems-miss messages are visible under `$VERBOSE = nil` as `bin/solargraph` runs, and that `generic` renders as `untyped` both at the ComplexType level and through a method signature. Full suite: no new failures.

*Opened as a draft. This PR was written by Claude (Anthropic's Claude Code) on behalf of @apiology.*

### DIFF
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -312,7 +312,10 @@ module Solargraph
         elsif name.downcase == 'nil'
           'nil'
         elsif name == GENERIC_TAG_NAME
-          all_params.first&.name
+          # A generic with no parameter would otherwise render nil, which
+          # callers interpolate into signatures as an empty string
+          # ("def foo: () -> ")
+          all_params.first&.name || 'untyped'
         elsif %w[Class Module].include?(name)
           rbs_name
         elsif %w[Tuple Array].include?(name) && fixed_parameters?
--- a/lib/solargraph/logging.rb
+++ b/lib/solargraph/logging.rb
@@ -16,8 +16,8 @@ module Solargraph
               LOG_LEVELS.fetch(configured_level)
             else
               if configured_level
-                warn "Invalid value for SOLARGRAPH_LOG: #{configured_level.inspect} - " \
-                     "valid values are #{LOG_LEVELS.keys}"
+                $stderr.puts "Invalid value for SOLARGRAPH_LOG: #{configured_level.inspect} - " \
+                             "valid values are #{LOG_LEVELS.keys}"
               end
               DEFAULT_LOG_LEVEL
             end
--- a/lib/solargraph/logging.rb
+++ b/lib/solargraph/logging.rb
@@ -4,24 +4,30 @@ require 'logger'
 
 module Solargraph
   module Logging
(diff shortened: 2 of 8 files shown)


apiology:
Please push up an empty commit to trigger a fresh CI run now that GitHub is up again

apiology:
**Claude:**

`gh run rerun --failed` on the three failed runs was rejected for lack of
admin rights on castwide/solargraph, so pushed an empty commit (d6fbd0a5)
to retrigger CI instead.

PR #1313, opened 2026-08-17
Name where inference stopped in typecheck inference errors

apiology:
**Problem**

A local whose expression calls a method that does not exist reports only the variable's name:

```ruby
# @type [String]
v = 'a'.no_such.upcase
```

```
example1.rb:5: Variable type could not be inferred for v
```

Nothing says whether inference stopped at `'a'`, `no_such`, or `upcase`, so diagnosing it means bisecting the expression by hand. The same gap affects `... return type could not be inferred`, which names the method but not the part of the expression responsible.

**Solution**

Both messages now name where inference stopped and what it was called on:

```
example1.rb:5: Variable type could not be inferred for v: `no_such` could not be resolved on String
example2.rb:3: Report#rows return type could not be inferred: could not determine the return type of `fetch` on Report
```

`Chain#first_undefined_link` re-walks the chain, on the error path only.

A bare-variable return walks into that variable's own assignment, so the cause can name a call appearing nowhere in the return expression itself.

Messages stay byte-identical when no cause can be named, and when the failing call is the one already named.

*Written by Claude (Anthropic's Claude Code) on behalf of @apiology.*

### DIFF
--- a/lib/solargraph/pin/base_variable.rb
+++ b/lib/solargraph/pin/base_variable.rb
@@ -180,14 +180,15 @@ module Solargraph
       # @param api_map [ApiMap]
       # @return [Source::Chain::LinkResolution, nil]
       def probe_blame api_map
-        return nil if closure.nil?
+        return nil if closure.nil? || location.nil?
 
         assignments.each do |parent_node|
           value_position_nodes_only(parent_node).each do |node|
             next if node.nil? || node.type == :NIL || node.type == :nil
             rng = Range.from_node(node)
             next if rng.nil?
-            # @sg-ignore Need to add nil check here
+            # @sg-ignore location nil-guarded at method entry; return-guard
+            #   narrowing not tracked - https://github.com/castwide/solargraph/issues/1254
             clip = api_map.clip_at(location.filename, rng.ending)
             chain = Parser.chain(node, nil, nil)
             # @sg-ignore closure nil-guarded at method entry; return-guard
--- a/lib/solargraph/pin/base_variable.rb
+++ b/lib/solargraph/pin/base_variable.rb
@@ -172,6 +172,33 @@ module Solargraph
         types
       end
 
+      # Locate where inference broke for this variable's assignment: the
+      # first link of the assigned expression's chain whose type resolved
+      # undefined. Diagnostic-only; returns nil when the assignment infers
+      # cleanly (or there is no assignment to walk).
+      #
+      # @param api_map [ApiMap]
+      # @return [Source::Chain::LinkResolution, nil]
+      def probe_blame api_map
+        return nil if closure.nil?
+
+        assignments.each do |parent_node|
+          value_position_nodes_only(parent_node).each do |node|
+            next if node.nil? || node.type == :NIL || node.type == :nil
+            rng = Range.from_node(node)
+            next if rng.nil?
+            # @sg-ignore Need to add nil check here
+            clip = api_map.clip_at(location.filename, rng.ending)
+            chain = Parser.chain(node, nil, nil)
+            # @sg-ignore closure nil-guarded at method entry; return-guard
+            #   narrowing not tracked - https://github.com/castwide/solargraph/issues/1254
+            failure = chain.first_undefined_link(api_map, closure, clip.locals)
+            return failure unless failure.nil?
+          end
+        end
+        nil
+      end
+
       # @param api_map [ApiMap]
       # @return [ComplexType, ComplexType::UniqueType]
       def probe api_map
--- a/lib/solargraph/pin/base_variable.rb
+++ b/lib/solargraph/pin/base_variable.rb
@@ -187,12 +187,10 @@ module Solargraph
             next if node.nil? || node.type == :NIL || node.type == :nil
             rng = Range.from_node(node)
             next if rng.nil?
-            # @sg-ignore location nil-guarded at method entry; return-guard
-            #   narrowing not tracked - https://github.com/castwide/solargraph/issues/1254
+            # @sg-ignore https://github.com/castwide/solargraph/issues/1254
             clip = api_map.clip_at(location.filename, rng.ending)
             chain = Parser.chain(node, nil, nil)
-            # @sg-ignore closure nil-guarded at method entry; return-guard
-            #   narrowing not tracked - https://github.com/castwide/solargraph/issues/1254
+            # @sg-ignore https://github.com/castwide/solargraph/issues/1254
             failure = chain.first_undefined_link(api_map, closure, clip.locals)
             return failure unless failure.nil?
           end
--- a/lib/solargraph/pin/method.rb
+++ b/lib/solargraph/pin/method.rb
@@ -452,6 +452,37 @@ module Solargraph
         api_map.get_method_stack(method_namespace, method_name, scope: scope).reject { |pin| pin.path == path }
       end
 
+      # Locate where inference broke for this method's return value: the
+      # first failing link among its return-node chains. A failing
+      # bare-variable return delegates to that variable's own assignment,
+      # since the real failure is usually there. Diagnostic-only; nil when
+      # nothing failed or there is nothing to walk.
+      #
+      # @param api_map [ApiMap]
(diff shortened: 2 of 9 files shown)


lib/solargraph/pin/base_variable.rb
```diff
@@ -172,6 +172,33 @@ def return_types_from_node parent_node, api_map
...
+      # @return [Source::Chain::LinkResolution, nil]
+      def probe_blame api_map
+        return nil if closure.nil?
+
+        assignments.each do |parent_node|
+          value_position_nodes_only(parent_node).each do |node|
+            next if node.nil? || node.type == :NIL || node.type == :nil
+            rng = Range.from_node(node)
+            next if rng.nil?
+            # @sg-ignore Need to add nil check here
```
apiology: What's the actual typecheck error here?  If it's related to the above line's nil check, is there an open PR or issue addressing that?

lib/solargraph/pin/base_variable.rb
```diff
@@ -172,6 +172,33 @@ def return_types_from_node parent_node, api_map
...
+            clip = api_map.clip_at(location.filename, rng.ending)
+            chain = Parser.chain(node, nil, nil)
+            # @sg-ignore closure nil-guarded at method entry; return-guard
+            #   narrowing not tracked - https://github.com/castwide/solargraph/issues/1254
```
apiology: Ditto on actual error

lib/solargraph/pin/base_variable.rb
apiology: The actual error, from `solargraph typecheck --level strong lib/solargraph/pin/base_variable.rb` with only this line removed:

    lib/solargraph/pin/base_variable.rb:190: Unresolved call to filename on Solargraph::Location, nil

That is not a tool limitation ��� the nil check really was missing. `Pin::Base#location` is typed `Location, nil`, and `probe_blame` here guarded only `closure.nil?` before calling `location.filename`. `Pin::Method#probe_blame`, added in this same PR, does guard it: `return nil if node.nil? || method_body_node.nil? || location.nil?`.

The guard is now extended to `return nil if closure.nil? || location.nil?`.

Measured after adding the guard: removing the `@sg-ignore` again still reports the identical `Unresolved call to filename on Solargraph::Location, nil` at that line, so the suppression is still needed and its cause is now the same return-guard narrowing gap as the line below ��� https://github.com/castwide/solargraph/issues/1254. Retagged to match, in the same wording used in `Pin::Method#probe_blame`.

lib/solargraph/pin/base_variable.rb
apiology:
```
lib/solargraph/pin/base_variable.rb:193: Wrong argument type for Solargraph::Source::Chain#first_undefined_link: name_pin expected Solargraph::Pin::Base, received Solargraph::Pin::Closure, nil
```

`closure` is `Pin::Closure, nil` and is guarded by `return nil if closure.nil?` at method entry, but the narrowing is not carried past the return guard ��� https://github.com/castwide/solargraph/issues/1254 ("Flow-sensitive typing doesn't narrow past a raise/return-based nil guard"). No code change here; the existing tag matches the observed error.

apiology: So we can track impact locally of bugs/missing features, please make this just `@sg-ignore <PR or issue URL>` - folks can go to the URL to understand more

lib/solargraph/pin/method.rb
```diff
@@ -452,6 +452,37 @@ def rest_of_stack api_map
...
+      # @param api_map [ApiMap]
+      # @return [Source::Chain::LinkResolution, nil]
+      def probe_blame api_map
+        return nil if node.nil? || method_body_node.nil? || location.nil?
+        returns_from_method_body(method_body_node).each do |n|
+          next if n.nil? || %i[NIL nil].include?(n.type)
+          rng = Range.from_node(n)
+          next unless rng
+          # @sg-ignore location nil-guarded at method entry; return-guard
+          #   narrowing not tracked - https://github.com/castwide/solargraph/issues/1254
```
apiology: Ditto

lib/solargraph/pin/method.rb
```diff
@@ -452,6 +452,37 @@ def rest_of_stack api_map
...
+          clip = api_map.clip_at(location.filename, rng.ending)
+          # @sg-ignore see above
```
apiology: Ditto

spec/type_checker/levels/strong_spec.rb
```diff
@@ -925,5 +927,213 @@ def baz(bases)
...
+
+    it 'reports a diagnostic for a nonexistent method called on an untagged local' do
+      pending 'A plain typo passes the typechecker: call_problems suppresses the ' \
+              'unresolved-call report when the walk\'s `found` pin is an internal method ' \
+              'whose type is undefined, and with no declared tag there is no ' \
+              'variable-validation path to fire instead, so nothing is reported. The ' \
+              'boundary is not yet mapped - the six-call chain in the spec below ' \
+              '(xs.first.to_s.no_such.upcase.size, also untagged) does report ' \
+              '"Unresolved call to no_such on String", so something about the receiver ' \
+              'decides it, and that difference is unexplained.'
```
apiology: This is very hard to read as-is; maybe use a shorter version and turn the long form into a comment, or maybe use a heredoc.

lib/solargraph/pin/base_variable.rb
apiology: **Claude:**

Collapsed to a single-line `@sg-ignore <issue URL>` in 55a9538.

lib/solargraph/pin/method.rb
apiology: **Claude:**

Same, both lines collapsed to `@sg-ignore <issue URL>` in 55a9538.

lib/solargraph/pin/method.rb
apiology: **Claude:**

Same, both lines collapsed to `@sg-ignore <issue URL>` in 55a9538.

spec/type_checker/levels/strong_spec.rb

Moved the long explanation into a comment above the `it` block, and
reduced the `pending` message itself to a one-line heredoc, in 55a9538.

lib/solargraph/type_checker.rb
```diff
@@ -676,6 +676,47 @@ def internal? pin
...
+      blame = pin.probe_blame(api_map)
+      cause = blame_description(blame) unless blame.nil?
+      message += ": #{cause}" unless cause.nil?
+      message
+    end
+
+    # A short plain-language cause for the first failing link of a chain.
+    # The pin count picks the wording: zero pins means the name did not
+    # resolve at all; otherwise a definition exists but its return type
+    # could not be determined.
```
apiology: 3 lines at most

lib/solargraph/type_checker.rb
```diff
@@ -676,6 +676,47 @@ def internal? pin
...
+    #
+    # @param blame [Source::Chain::LinkResolution]
+    # @return [String, nil] nil when the failing link has no name to report
+    def blame_description blame
+      # The parser produced an undefined placeholder rather than a named
+      # link, so there is no name to report - say nothing rather than
+      # guessing. (Numbered block parameters land here.)
```
apiology: 1 line at most - and not in love with the magic string.  Isn't there a link.undefined? or something to call?

spec/source/chain_spec.rb
```diff
@@ -435,4 +435,36 @@ def obj(foo); end
...
+      expect(failure.index).to eq(1)
+      expect(failure.total).to eq(3)
+      expect(failure.pin_count).to eq(0)
+      expect(failure.receiver_type.to_s).to include('String')
+    end
+  end
+
+  it 'chains a numbered-parameter block like an ordinary block' do
+    pending 'NodeChainer#generate_links has no :numblock branch, so a `_1` block produces ' \
+            'no links and the whole expression resolves to the undefined-call placeholder'
```
apiology: Don't we have a PR out on that?  Replace with just that URL

PR #1314, opened 2026-08-17
Resolve a self return type per union member

apiology:
A method that declares a `self` return type, called on a union receiver, infers the whole union instead of the member that supplied the method.

```ruby
class Counter
  # @return [self]
  def to_sym
    self
  end
end

# @param x [String, Counter]
# @return [Symbol, Counter]
def probe_union(x)
  x.to_sym
end
```

```
before:  probe.rb:10: Declared return type ::Symbol, ::Counter does not match
                      inferred type ::Symbol, ::String, ::Counter for #probe_union
after:   0 problems found.
```

Any method declaring `self` is affected, in any position — `@return [Array<self>]` on an `Alpha, Beta` receiver collapsed the same way.

`Chain::Call#resolve` used to split the binder only to collect method pins, then flatten them and call `#inferred_pins` once with `name_pin` still bound to the whole union. It now resolves one unique type at a time against a `name_pin` bound to that type alone, so `self` can only mean the type that supplied the pin. Results dedupe on path and resolved return type, so a `self`-returning method shared by every member still yields each one: `(String | Symbol)#itself` stays `String, Symbol`.

It also removes an `@sg-ignore`: `Diagnostics::TypeCheck#diagnose` calls `level.to_sym` where `level` is `String, Symbol`, which needed suppressing and now resolves to `Symbol`.

### DIFF
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -55,18 +55,31 @@ module Solargraph
           # chain.rb#maybe_nil will add the nil type later, we just
           # need to worry about the not-nil case
 
-          # @sg-ignore Need to handle duck-typed method calls on union types
           binder = binder.without_nil if nullable?
-          # @sg-ignore Need to handle duck-typed method calls on union types
-          pin_groups = binder.each_unique_type.map do |context|
+          # Resolve each arm of a union receiver against that arm alone. A
+          # method whose declared return type is `self` must resolve to the
+          # arm that supplied the method pin, not to the entire union - e.g.
+          # for a String, Symbol receiver, Symbol#to_sym (RBS `-> self`) is
+          # Symbol, not String, Symbol.
+          # @type [::Array<Pin::Base>]
+          resolved = []
+          unresolved_arm = false
+          binder.each_unique_type do |context|
             ns_tag = context.namespace == '' ? '' : context.namespace_type.tag
             stack = api_map.get_method_stack(ns_tag, word, scope: context.scope)
-            [stack.first].compact
+            pin = stack.first
+            if pin.nil?
+              unresolved_arm = true
+              next
+            end
+            resolved.concat inferred_pins([pin], api_map, name_pin, locals, context)
           end
-          pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
-          return [] if pins.empty?
-          inferred_pins(pins, api_map, name_pin, locals)
+          return [] if unresolved_arm && !api_map.loose_unions
+          return [] if resolved.empty?
+          # Different arms can resolve to the same pin path (a method
+          # inherited from a shared ancestor); dedup on the resolved return
+          # type too, so a real union doesn't lose every arm but the first.
+          resolved.uniq { |pin| [pin.path, pin.return_type.tag] }
         end
 
         private
@@ -75,8 +88,13 @@ module Solargraph
         # @param api_map [ApiMap]
         # @param name_pin [Pin::Base]
         # @param locals [::Array<Solargraph::Pin::LocalVariable, Solargraph::Pin::Parameter>]
+        # @param self_binder [ComplexType, ComplexType::UniqueType, nil] The
+        #   type that `self` refers to in the resolved pins' declarations. For a
+        #   union receiver this is the single arm which supplied `pins`, not the
+        #   whole union.
         # @return [::Array<Pin::Base>]
-        def inferred_pins pins, api_map, name_pin, locals
+        def inferred_pins pins, api_map, name_pin, locals, self_binder = nil
+          self_binder ||= name_pin.binder
           result = pins.map do |p|
             next p unless p.is_a?(Pin::Method)
             overloads = p.signatures
@@ -131,11 +149,11 @@ module Solargraph
                                                                                     blocktype)
                 # @todo It shouldn't be necessary to choose either generics or macros
                 new_return_type = if new_signature_pin.return_type.defined?
-                  new_signature_pin.return_type
-                else
-                  named_types = p.parameter_names.zip(arguments.map { |arg| ComplexType.try_parse(simple_convert(arg.node).to_s) }).to_h
-                  p.typify(api_map).expand(named_types)
-                end
+                                    new_signature_pin.return_type
+                                  else
+                                    named_types = p.parameter_names.zip(arguments.map { |arg| ComplexType.try_parse(simple_convert(arg.node).to_s) }).to_h
+                                    p.typify(api_map).expand(named_types)
+                                  end
                 self_type = if head?
                               # If we're at the head of the chain, we called a
                               # method somewhere that marked itself as returning
@@ -149,7 +167,7 @@ module Solargraph
                               # declaration - we can't just use the type of the
                               # method pin, as this might be a subclass of the
                               # place where the method is defined
-                              name_pin.binder
+                              self_binder
                             end
                 # This same logic applies to the YARD work done by
                 # 'with_params()'.
@@ -179,7 +197,7 @@ module Solargraph
               # @sg-ignore Need to add nil check here
               next pin if pin.return_type.undefined?
               # @sg-ignore Need to add nil check here
-              selfy = pin.return_type.self_to_type(name_pin.binder)
+              selfy = pin.return_type.self_to_type(self_binder)
               # @sg-ignore Need to add nil check here
               selfy == pin.return_type ? pin : pin.proxy(selfy)
             end
--- a/lib/solargraph/source/chain/call.rb
+++ b/lib/solargraph/source/chain/call.rb
@@ -56,11 +56,7 @@ module Solargraph
           # need to worry about the not-nil case
 
           binder = binder.without_nil if nullable?
-          # Resolve each arm of a union receiver against that arm alone. A
-          # method whose declared return type is `self` must resolve to the
-          # arm that supplied the method pin, not to the entire union - e.g.
-          # for a String, Symbol receiver, Symbol#to_sym (RBS `-> self`) is
-          # Symbol, not String, Symbol.
+          # Resolve each arm alone, so a `self` return type narrows to that arm, not the whole union.
           # @type [::Array<Pin::Base>]
           resolved = []
           unresolved_arm = false
@@ -72,10 +68,7 @@ module Solargraph
               unresolved_arm = true
               next
             end
-            # Give this arm's resolution a name_pin whose #binder is this
-            # arm alone, not the whole union - every name_pin.binder read
-            # inside #inferred_pins (including the Class#new special case)
-            # then sees only this arm, with nothing extra to keep in sync.
+            # This arm's name_pin binds only this arm, so #inferred_pins (incl. Class#new) sees just it.
             arm_name_pin = Pin::ProxyType.anonymous(name_pin.context,
                                                     closure: name_pin.closure,
                                                     gates: name_pin.gates,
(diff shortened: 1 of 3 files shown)


apiology:
Yes ��� YARD reaches this too, and the fix covers it without a second change.

`self_to_type` runs on the pin's parsed `ComplexType`, downstream of whichever syntax produced the declaration, so `@return [self]` and an RBS `-> self` arrive at the same call in `#inferred_pins`. Measured rather than reasoned about ��� this file has no RBS in it at all:

```ruby
class Alpha
  # @return [Symbol]
  def to_thing; end
end

class Beta
  # @return [self]
  def to_thing; end
end

# @param x [Alpha, Beta]
# @return [Symbol, Beta]
def probe_union(x)
  x.to_thing
end
```

`solargraph typecheck --level strong`, master at `8fda63384`:

```
probe.rb:17: Declared return type ::Symbol, ::Beta does not match inferred type ::Symbol, ::Alpha, ::Beta for #probe_union
1 problem found.
```

Same file at `4931a9554`: `0 problems found.`

Two further shapes, both pure YARD, inferred type before and after:

| shape | master | branch |
| --- | --- | --- |
| `@return [self]` inherited by both arms | `::Alpha, ::Beta` | `::Alpha, ::Beta` |
| `@return [Array<self>]` inherited by both arms | `::Array<::Alpha, ::Beta>` | `::Array<::Alpha>, ::Array<::Beta>` |

The first is the dedup guard holding ��� a method shared by every arm is not over-split. The second is the generic case distributing.

lib/solargraph/source/chain/call.rb
```diff
@@ -55,18 +55,40 @@ def resolve api_map, name_pin, locals
...
 
-          # @sg-ignore Need to handle duck-typed method calls on union types
           binder = binder.without_nil if nullable?
-          # @sg-ignore Need to handle duck-typed method calls on union types
-          pin_groups = binder.each_unique_type.map do |context|
+          # Resolve each arm of a union receiver against that arm alone. A
+          # method whose declared return type is `self` must resolve to the
+          # arm that supplied the method pin, not to the entire union - e.g.
+          # for a String, Symbol receiver, Symbol#to_sym (RBS `-> self`) is
+          # Symbol, not String, Symbol.
```
apiology: One line at most

lib/solargraph/source/chain/call.rb
```diff
@@ -55,18 +55,40 @@ def resolve api_map, name_pin, locals
...
-            [stack.first].compact
+            pin = stack.first
+            if pin.nil?
+              unresolved_arm = true
+              next
+            end
+            # Give this arm's resolution a name_pin whose #binder is this
+            # arm alone, not the whole union - every name_pin.binder read
+            # inside #inferred_pins (including the Class#new special case)
+            # then sees only this arm, with nothing extra to keep in sync.
```
apiology: One line at most

lib/solargraph/source/chain/call.rb
```diff
@@ -55,18 +55,40 @@ def resolve api_map, name_pin, locals
...
           end
-          pin_groups = [] if !api_map.loose_unions && pin_groups.any?(&:empty?)
-          pins = pin_groups.flatten.uniq(&:path)
-          return [] if pins.empty?
-          inferred_pins(pins, api_map, name_pin, locals)
+          return [] if unresolved_arm && !api_map.loose_unions
+          return [] if resolved.empty?
+          # Different arms can resolve to the same pin path (a method
+          # inherited from a shared ancestor); dedup on the resolved return
+          # type too, so a real union doesn't lose every arm but the first.
```
apiology: One line at most

spec/source/chain/call_spec.rb
```diff
@@ -361,8 +361,132 @@ def yielder(&blk)
...
+    api_map = Solargraph::ApiMap.new
+    api_map.map source
+
+    chain = Solargraph::Source::SourceChainer.chain(source, Solargraph::Position.new(5, 13))
+    type = chain.infer(api_map, Solargraph::Pin::ROOT_PIN, api_map.source_map('test.rb').locals)
+    # Class<Beta>'s own `.new` resolves separately to Beta; the bare `Class`
+    # arm falls back to the generic Class#new, which cannot know the
+    # constructed type. The gate for that fallback must read only this
+    # arm's binder - a Class<Beta> arm elsewhere in the union must not leak
+    # Beta into the bare Class arm's result.
```
apiology: 1 line at most

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Done in [d6928dff2](https://github.com/castwide/solargraph/commit/d6928dff2511d435ad45f5372d7fabb8bb9254b1) — the five-line block is now one line:

```ruby
# Resolve each arm alone, so a `self` return type narrows to that arm, not the whole union.
```

The worked `String, Symbol` / `Symbol#to_sym` example that filled the other four lines is now carried by the spec example instead of repeated here.

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Done in [d6928dff2](https://github.com/castwide/solargraph/commit/d6928dff2511d435ad45f5372d7fabb8bb9254b1) — the four-line block is now one line:

```ruby
# This arm's name_pin binds only this arm, so #inferred_pins (incl. Class#new) sees just it.
```

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Done in [d6928dff2](https://github.com/castwide/solargraph/commit/d6928dff2511d435ad45f5372d7fabb8bb9254b1) — the three-line block is now one line:

```ruby
# Dedup on return type too, so arms sharing an inherited pin path don't collapse to just the first.
```

spec/source/chain/call_spec.rb
apiology: **Claude:** Done in [f6c873469](https://github.com/castwide/solargraph/commit/f6c873469). Every comment in the new union-arm examples is now one line, and the one on the arm-order example is gone entirely — it restated the example name rather than giving a reason.

The five that remain each say why the expected tag is what it is, rather than narrating the steps. The one anchored here:

```ruby
# The bare `Class` arm cannot know what its #new constructs, and Beta must not leak in from the other arm.
```

`spec/source/chain/call_spec.rb` still passes: 38 examples, 0 failures, 8 pending.

lib/solargraph/source/chain/call.rb
```diff
@@ -55,18 +55,31 @@ def resolve api_map, name_pin, locals
           # chain.rb#maybe_nil will add the nil type later, we just
           # need to worry about the not-nil case
 
-          # @sg-ignore Need to handle duck-typed method calls on union types
           binder = binder.without_nil if nullable?
-          # @sg-ignore Need to handle duck-typed method calls on union types
-          pin_groups = binder.each_unique_type.map do |context|
+          # Resolve each arm alone, so a `self` return type narrows to that arm, not the whole union.
```
apiology: There's a LOT of unexamined assumptions of things being unions here.  There will also be intersections very soon, duck types, generics, etc etc.  This code (and certainly the comment!) will very quickly break - can we think how to treat these in a kind-independent way?

lib/solargraph/source/chain/call.rb
```diff
@@ -162,7 +175,10 @@ def match_overload_type overload, pin, api_map, name_pin, locals, type, new_sign
 
         # @param pins [::Enumerable<Pin::Base>]
         # @param api_map [ApiMap]
-        # @param name_pin [Pin::Base]
+        # @param name_pin [Pin::Base] name_pin.binder resolves `self` in the
+        #   resolved pins' declarations. For a union receiver, callers pass a
+        #   name_pin bound to the single arm that supplied `pins`, not the
+        #   whole union.
```
apiology: Is this focused on how inferred_pins uses this, or how /this PR/ changes it?

spec/source/chain/call_spec.rb
```diff
@@ -361,8 +361,119 @@ def yielder(&blk)
 
     chain = Solargraph::Source::SourceChainer.chain(source, Solargraph::Position.new(4, 11))
     type = chain.infer(api_map, Solargraph::Pin::ROOT_PIN, api_map.source_map('test.rb').locals)
-    # @todo It would be more accurate to return `Enumerator<Array<Integer>>` here
-    expect(type.tag).to eq('Enumerator<Integer, String, Array<Integer>>')
+    # Array#each's `-> Enumerator[Elem, self]` narrows to the Array<Integer> arm, not the union.
```
apiology: Is 'arm' a term first used in this PR?  If so, don't introduce it.

lib/solargraph/source/chain/call.rb
apiology: **Claude:** Split the question in two, because the answers differ.

**Enumeration is already kind-independent.** The loop never asks what kind the binder is — `each_unique_type` is implemented per kind and each one answers for itself:

```ruby
# ComplexType — walks its own items
def each_unique_type &block
  @items.each { |item| item.each_unique_type(&block) }
end

# ComplexType::TypeMethods, mixed into UniqueType — yields itself
def each_unique_type &block
  yield self
end
```

So the body receives one `UniqueType` at a time and treats every one the same way. A single-type binder yields exactly one and the loop degenerates to the old single pass. A new kind participates by implementing `each_unique_type`; this loop needs no change for that.

**Composition is not.** Accumulating every type's pins into one result and deduping is union semantics. For an intersection it would be wrong — you would want the conjunct that supplies the method, not the union of what all of them return. `api_map.loose_unions`, which decides whether one unresolved type kills the whole result, is union-shaped for the same reason; that one predates this PR.

I have made that assumption explicit rather than building an abstraction for a kind that does not exist yet, in [91af940df](https://github.com/apiology/solargraph/commit/91af940df):

```ruby
# Accumulating every type's result treats the binder as a union, which is the only
# multi-type binder ComplexType builds; dedup on return type as well as path so that
# types sharing an inherited pin do not collapse to just the first.
resolved.uniq { |pin| [pin.path, pin.return_type.tag] }
```

The locals are now `unresolved_type` and `type_name_pin`, and the loop comment reads "one unique type at a time" rather than naming unions.

The next step, when intersections land, is to move that accumulation onto the type itself — the enumerate-then-combine pair becomes one call that a union implements as concat-and-dedup and an intersection implements as select-one. Doing it now would mean guessing the intersection half.

lib/solargraph/source/chain/call.rb
apiology: **Claude:** It was the second one — the last three lines described what this branch changed about the callers, not what `inferred_pins` requires. Cut to the contract alone in [91af940df](https://github.com/apiology/solargraph/commit/91af940df):

```ruby
# @param name_pin [Pin::Base] its #binder resolves `self` in the declarations of `pins`
```

spec/source/chain/call_spec.rb
apiology: **Claude:** Yes — `git grep -i "\barm\b"` over `lib` and `spec` returned 15 hits, all of them lines this branch added: 4 in `call.rb`, 11 here. Gone in [91af940df](https://github.com/apiology/solargraph/commit/91af940df); the same grep now returns nothing.

In `lib` the term is **unique type**, matching what `each_unique_type` yields, which also stops the code reading as if the binder were always a union. In the specs, which really are about unions, it is **union member**. This line now reads:

```ruby
# Array#each's `-> Enumerator[Elem, self]` narrows to the Array<Integer> member, not the union.
```

Six example names changed with it, e.g. `resolves an RBS self return type to the union member that supplied the method`.

PR #1315, opened 2026-08-17
Re-opened classes should not reset superclass to 'Object'

apiology:
With activesupport in the bundle, Solargraph reports the wrong superclass for `DateTime`:

```ruby
DateTime.superclass                # => Date, at runtime
get_superclass('DateTime')         # => Object
```

`DateTime` is then cut off from `Date` and from everything above it.

The `Object` comes from activesupport reopening the class:

```ruby
# active_support/core_ext/date_time/blank.rb
class DateTime # :nodoc:
  def blank?
    false
  end
end
```

A reopening with no superclass clause asserts nothing about ancestry — and it cannot, because Ruby fixes a class's superclass at its first definition and refuses to let anything restate it differently:

```ruby
class DateTime < Object; end       # TypeError: superclass mismatch for class DateTime
```

A class with no superclass reference already resolves to `Object` via `Store#try_special_superclasses`, so the mapper now skips recording an `Object` reference at all, leaving only the specific one that carries a claim. `get_superclass` is untouched — it stays a plain `.first` with no tie-break, because there is no longer a conflict to arbitrate. `BasicObject` references are kept, since the only class YARD defaults to `BasicObject` is `Object` itself, where the reference is correct either way.

Suite: 1625 examples, 0 failures, 60 pending. Opening as a draft.

*This PR was written by Claude (Anthropic's Claude Code) on behalf of @apiology.*

### DIFF
--- a/lib/solargraph/yard_map/mapper.rb
+++ b/lib/solargraph/yard_map/mapper.rb
@@ -50,7 +50,16 @@ module Solargraph
                          else
                            code_object.superclass.to_s
                          end
-            result.push Solargraph::Pin::Reference::Superclass.new(name: superclass, closure: nspin, source: :yard_map)
+            # YARD fills in `Object` for every class whose definition it never
+            # saw a superclass clause on - including a bare `class Foo`
+            # reopening, which asserts nothing about ancestry - and keeps no
+            # record of which case it was. A class with no superclass reference
+            # already resolves to Object, so recording this one adds nothing and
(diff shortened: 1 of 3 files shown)


castwide:
`ClassDefinition` is not actually a convention. It's just being used as a namespace for a node processor that doesn't require a convention. The same is true of `DataDefinition` and `StructDefinition`.

I'll go ahead and merge this PR as is and address refactoring in a separate issue.

PR #1319, opened 2026-08-19
Keep a block's reassignment of an outer local visible after the block

apiology:
**Problem:** A local reassigned inside a block loses that assignment from its type once the block ends, so a call that is safe at runtime cannot be inferred.

```ruby
# @param lines [Array<String>]
# @return [Integer, nil]
def foo lines       # foo return type could not be inferred
  x = nil
  lines.each { |line| x = [line] }
  x&.length
end
```

The same assignment inside an `if` infers `nil, Array` and checks clean.

**Solution:** `LvasgnNode#process` attached the pin to the `Pin::Block`, whose presence ends at the block's closing token; it now walks outward to the enclosing closure already holding that name.

A name the block introduces itself — parameter, shadow arg, or first assignment — stays block-local.

Inference becomes `nil, Array`, not `Array`: the block is not assumed to run, so the pre-block value stays in the union and genuine nil errors survive.

`Call#inferred_pins` is here because its `pins.map` block shadowed the outer `result=` with two inner `result` assignments the new tracking now picks up; both are renamed.

### DIFF
--- a/lib/solargraph/parser/parser_gem/node_processors/lvasgn_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/lvasgn_node.rb
@@ -28,14 +28,8 @@ module Solargraph
 
           private
 
-          # The closure the assigned variable belongs to.
-          #
-          # An assignment inside a block writes through to a variable of the
-          # same name already in scope outside the block, so the pin belongs
-          # to that outer closure and stays visible after the block ends. A
-          # name the block introduces itself - a block parameter, a shadow
-          # arg, or an earlier assignment in the same block - is block-local
-          # and shadows anything outside.
+          # The closure `name` is already declared in, searching outward
+          # through enclosing blocks; defaults to the current block if none.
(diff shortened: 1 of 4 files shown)


lib/solargraph/parser/parser_gem/node_processors/lvasgn_node.rb
```diff
@@ -9,20 +9,60 @@ class LvasgnNode < Parser::NodeProcessor::Base
...
+          private
+
+          # The closure the assigned variable belongs to.
+          #
+          # An assignment inside a block writes through to a variable of the
+          # same name already in scope outside the block, so the pin belongs
+          # to that outer closure and stays visible after the block ends. A
+          # name the block introduces itself - a block parameter, a shadow
+          # arg, or an earlier assignment in the same block - is block-local
+          # and shadows anything outside.
```
apiology: 2 lines at most

PR #1320, opened 2026-08-19
Support blocks with implicit parameters (_1, it) and backtick commands

apiology:
Three block and literal forms have no type at `--level strong`, so calls on them are unresolved and enclosing methods lose their return type.

```ruby
# @return [Array<String>]
def numbered
  'a,b'.split(',').map { _1.upcase }   # Unresolved call to _1
end

# @return [Array<String>]
def implicit_it
  'a,b'.split(',').map { it.upcase }   # Unresolved call to it
end

# @return [String]
def backticks
  `hostname`.strip                     # Unresolved call to strip
end
```

All are clean in explicit form. A wrong tag also passes silently on the implicit block form: `@return [Array<Integer>]` on `'a,b'.split(',').map { it }` reports nothing.

`_1` produces a `numblock` node, so the call was chained as though no block were passed. `it` arrives as an ordinary block missing only its parameter — hence the silent pass. Backticks produce an `xstr` node, absent from the literal type table. Neither block form records parameters as an args node does, so both are synthesized; a local named `it` still wins, matching Ruby.

### DIFF
--- a/lib/solargraph/parser/parser_gem/node_processors/block_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/block_node.rb
@@ -28,10 +28,35 @@ module Solargraph
               source: :parser
             )
             pins.push block_pin
+            add_numbered_parameters block_pin if node.type == :numblock
             process_children region.update(closure: block_pin)
           end
 
           private
+
+          # A numblock node stores the highest numbered parameter used in
+          # its body (e.g., 2 for `_2`) instead of an args node, so the
+          # parameter pins have to be synthesized here.
+          #
+          # @param block_pin [Pin::Block]
+          # @return [void]
+          def add_numbered_parameters block_pin
+            count = node.children[1]
+            return unless count.is_a?(::Integer)
+
+            (1..count).each do |number|
+              locals.push Solargraph::Pin::Parameter.new(
+                location: block_pin.location,
+                closure: block_pin,
+                name: "_#{number}",
+                # @sg-ignore Need to add nil check here
+                presence: block_pin.location.range,
+                decl: :arg,
+                source: :parser
+              )
+              block_pin.parameters.push locals.last
+            end
+          end
 
           def other_class_eval?
             node.children[0].type == :send &&
--- a/lib/solargraph/parser/parser_gem/node_processors/block_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/block_node.rb
@@ -28,34 +28,107 @@ module Solargraph
               source: :parser
             )
             pins.push block_pin
-            add_numbered_parameters block_pin if node.type == :numblock
+            add_implicit_parameters block_pin
             process_children region.update(closure: block_pin)
           end
 
           private
 
-          # A numblock node stores the highest numbered parameter used in
-          # its body (e.g., 2 for `_2`) instead of an args node, so the
-          # parameter pins have to be synthesized here.
+          # Blocks written with implicit parameters carry no args node for
+          # ArgsNode to work from, so their parameter pins are synthesized
+          # here instead.
           #
           # @param block_pin [Pin::Block]
           # @return [void]
-          def add_numbered_parameters block_pin
-            count = node.children[1]
-            return unless count.is_a?(::Integer)
-
-            (1..count).each do |number|
-              locals.push Solargraph::Pin::Parameter.new(
-                location: block_pin.location,
-                closure: block_pin,
-                name: "_#{number}",
-                # @sg-ignore Need to add nil check here
-                presence: block_pin.location.range,
-                decl: :arg,
-                source: :parser
-              )
-              block_pin.parameters.push locals.last
+          def add_implicit_parameters block_pin
+            if node.type == :numblock
+              numbered_parameter_count.times { |idx| add_parameter block_pin, "_#{idx + 1}" }
+            elsif implicit_it_parameter?
+              add_parameter block_pin, 'it'
             end
+          end
+
+          # A numblock stores the highest numbered parameter its body uses
+          # (2 for `_2`) where an ordinary block stores its args node.
+          #
+          # @sg-ignore flow sensitive typing does not narrow the return value through the is_a? guard
+          # @return [Integer]
+          def numbered_parameter_count
+            count = node.children[1]
+            return 0 unless count.is_a?(::Integer)
+
+            count
+          end
+
+          # An `it` block reaches us as an ordinary block with an empty args
+          # node, so the node itself records nothing about the parameter and
+          # the only available signal is an `it` reference in the body.
+          #
+          # Mixing `it` with ordinary parameters is a syntax error, so an
+          # empty args node is a precondition rather than a guess. A local
+          # variable named `it` in an enclosing scope does take precedence
+          # over the implicit parameter, and is parsed identically, so it has
+          # to be excluded here.
+          #
+          # @return [Boolean]
+          def implicit_it_parameter?
+            args = node.children[1]
+            return false unless Parser.is_ast_node?(args)
+            # @sg-ignore Translate to something flow sensitive typing understands
+            return false unless args.type == :args && args.children.empty?
+            return false if shadowed_it_local?
+            references_it?(node.children[2])
+          end
+
+          # Only a local variable declared outside any block shadows the
+          # implicit parameter. A block's own `it` must not stop a nested
+          # block from having one: in `xs.map { it.map { it.upcase } }` each
+          # `it` belongs to the block it appears in.
+          #
+          # @return [Boolean]
+          def shadowed_it_local?
+            loc = get_node_location(node)
+            locals.any? do |pin|
+              pin.name == 'it' && !pin.closure.is_a?(Pin::Block) && pin.visible_at?(region.closure, loc)
+            end
+          end
+
+          # `it` inside a nested block belongs to that block, so the search
+          # skips a nested block's body. Its receiver and arguments are still
+          # searched, since those sit in the enclosing block: the `it` in
+          # `xs.map { it.map { |y| y } }` is the outer block's parameter.
+          #
+          # @param subject [BasicObject, nil]
+          # @return [Boolean]
+          def references_it? subject
+            return false unless Parser.is_ast_node?(subject)
+            # @sg-ignore Translate to something flow sensitive typing understands
+            return true if subject.type == :lvar && subject.children[0] == :it
+            # @sg-ignore Translate to something flow sensitive typing understands
+            children = if %i[block numblock].include?(subject.type)
+                         # @sg-ignore Translate to something flow sensitive typing understands
+                         subject.children[0..1]
+                       else
+                         # @sg-ignore Translate to something flow sensitive typing understands
+                         subject.children
+                       end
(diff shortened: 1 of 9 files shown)


lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,116 @@ def process
...
+                         # @sg-ignore Translate to something flow sensitive typing understands
+                         subject.children
+                       end
+            children.any? { |child| references_it?(child) }
+          end
+
+          # @param block_pin [Pin::Block]
+          # @param name [String]
+          # @return [void]
+          def add_parameter block_pin, name
```
apiology: How about "add_implicit_parameter"

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,116 @@ def process
...
+            end
+          end
+
+          # A numblock stores the highest numbered parameter its body uses
+          # (2 for `_2`) where an ordinary block stores its args node.
+          #
+          # @sg-ignore flow sensitive typing does not narrow the return value through the is_a? guard
+          # @return [Integer]
+          def numbered_parameter_count
+            count = node.children[1]
```
apiology, line 54: We document elsewhere what a Node structure looks like for a given situation - could you add that documentation to the docstring?

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,116 @@ def process
...
           private
 
+          # Blocks written with implicit parameters carry no args node for
+          # ArgsNode to work from, so their parameter pins are synthesized
+          # here instead.
+          #
+          # @param block_pin [Pin::Block]
+          # @return [void]
+          def add_implicit_parameters block_pin
+            if node.type == :numblock
```
apiology: Maybe note that numblocks are regular blocks that don't have a parameter block?

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,116 @@ def process
...
+          #
+          # Mixing `it` with ordinary parameters is a syntax error, so an
+          # empty args node is a precondition rather than a guess. A local
+          # variable named `it` in an enclosing scope does take precedence
+          # over the implicit parameter, and is parsed identically, so it has
+          # to be excluded here.
+          #
+          # @return [Boolean]
+          def implicit_it_parameter?
+            args = node.children[1]
```
apiology: args or params?

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,116 @@ def process
...
+            return false unless Parser.is_ast_node?(args)
+            # @sg-ignore Translate to something flow sensitive typing understands
```
apiology: Why postpone this work?  If this is a solargraph typechecking limitation that should be fixed, choose a different slug from rules.rb

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,116 @@ def process
...
+          def references_it? subject
+            return false unless Parser.is_ast_node?(subject)
+            # @sg-ignore Translate to something flow sensitive typing understands
+            return true if subject.type == :lvar && subject.children[0] == :it
+            # @sg-ignore Translate to something flow sensitive typing understands
+            children = if %i[block numblock].include?(subject.type)
+                         # @sg-ignore Translate to something flow sensitive typing understands
+                         subject.children[0..1]
+                       else
+                         # @sg-ignore Translate to something flow sensitive typing understands
```
apiology: See above

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
apiology: **Claude:**

Renamed `add_parameter` to `add_implicit_parameter` in 925cc75.

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
apiology: **Claude:**

Added a sexp example to `numbered_parameter_count`'s docstring in 925cc75,
matching the `s(:type, ...)` convention used in `masgn_node.rb`:

```
# [1, 2].each { _1 + _2 } parses as:
#
#   s(:numblock,
#     s(:send, s(:array, s(:int, 1), s(:int, 2)), :each),
#     2,
#     s(:send, s(:lvar, :_1), :+, s(:lvar, :_2)))
```

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
apiology: **Claude:**

Added in 925cc75: "A numblock is a regular block that has no args node -
it stores its highest numbered parameter where an ordinary block stores
that node instead."

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
apiology: **Claude:**

`args` - it's the parser gem's own node type for this position (`:args`),
matching the existing `ArgsNode` processor class name in this codebase.
Left it as `args`.

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
apiology: **Claude:**

Not postponed work - it's a real Solargraph typechecking limitation, and
"Translate to something flow sensitive typing understands" was a
placeholder that undersold that. Isolated the mechanism: flow-sensitive
typing only recognizes a literal `.is_a?`/`.kind_of?`/`.instance_of?` call
directly on the guarded variable. It does not narrow through a
user-defined predicate method that wraps one internally, like
`Parser.is_ast_node?` here (`node.is_a?(::Parser::AST::Node)` under the
hood). Confirmed with a minimal repro - inlining the `is_a?` check narrows
fine, wrapping it in a one-line predicate method breaks it.

Retagged both markers in 925cc75 to
`tool-limitation:is_a-narrowing:predicate-wrapper`, with the mechanism
explained inline. No upstream issue filed yet - "Translate to..." isn't
one of the `rules.rb` categories, so there wasn't a real slug to fall back
to; this new one describes the actual cause instead. Whether to file an
issue for it is on the decision list I'm keeping in the Claude Code
session, separate from the narrower external-gem-class narrowing gap
already on that list from PR 1311.

This pattern (`# @sg-ignore Translate to something flow sensitive typing
understands`) appears 30 times across the codebase, not just the 5 in
this file - retagging all of them is more than this PR's scope, so I only
touched the two you flagged plus the other three in the same method
(`references_it?`) that share the identical cause.

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
apiology: **Claude:**

Not postponed work - it's a real Solargraph typechecking limitation, and
"Translate to something flow sensitive typing understands" was a
placeholder that undersold that. Isolated the mechanism: flow-sensitive
typing only recognizes a literal `.is_a?`/`.kind_of?`/`.instance_of?` call
directly on the guarded variable. It does not narrow through a
user-defined predicate method that wraps one internally, like
`Parser.is_ast_node?` here (`node.is_a?(::Parser::AST::Node)` under the
hood). Confirmed with a minimal repro - inlining the `is_a?` check narrows
fine, wrapping it in a one-line predicate method breaks it.

Retagged both markers in 925cc75 to
`tool-limitation:is_a-narrowing:predicate-wrapper`, with the mechanism
explained inline. No upstream issue filed yet - "Translate to..." isn't
one of the `rules.rb` categories, so there wasn't a real slug to fall back
to; this new one describes the actual cause instead. Whether to file an
issue for it is on the decision list I'm keeping in the Claude Code
session, separate from the narrower external-gem-class narrowing gap
already on that list from PR 1311.

This pattern (`# @sg-ignore Translate to something flow sensitive typing
understands`) appears 30 times across the codebase, not just the 5 in
this file - retagging all of them is more than this PR's scope, so I only
touched the two you flagged plus the other three in the same method
(`references_it?`) that share the identical cause.

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,130 @@ def process
...
           end
 
           private
 
+          # Blocks written with implicit parameters carry no args node for
+          # ArgsNode to work from, so their parameter pins are synthesized
+          # here instead. A numblock is a regular block that has no args
+          # node - it stores its highest numbered parameter where an
+          # ordinary block stores that node instead (see
+          # #numbered_parameter_count).
```
apiology: 3 lines at most

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,130 @@ def process
...
+
+          # An `it` block reaches us as an ordinary block with an empty args
+          # node, so the node itself records nothing about the parameter and
+          # the only available signal is an `it` reference in the body.
+          #
+          # Mixing `it` with ordinary parameters is a syntax error, so an
+          # empty args node is a precondition rather than a guess. A local
+          # variable named `it` in an enclosing scope does take precedence
+          # over the implicit parameter, and is parsed identically, so it has
+          # to be excluded here.
```
apiology: 2 lines at most

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,130 @@ def process
...
+          # implicit parameter. A block's own `it` must not stop a nested
+          # block from having one: in `xs.map { it.map { it.upcase } }` each
+          # `it` belongs to the block it appears in.
+          #
+          # This diverges from Ruby for an enclosing block with an explicit
+          # parameter named `it`. In `xs.map { |it| ys.map { it } }` Ruby
+          # reads the inner `it` as the outer parameter; this gives the
+          # inner block its own. Treating a block parameter as a shadow
+          # instead would break every nested implicit `it`, which is the
+          # far more common shape.
```
apiology: I don't follow - treating this like Ruby does would break something?  Maybe an example would help?

lib/solargraph/parser/parser_gem/node_processors/block_node.rb
```diff
@@ -28,11 +28,122 @@ def process
...
+          #
+          # @return [Boolean]
+          def implicit_it_parameter?
+            args = node.children[1]
+            return false unless Parser.is_ast_node?(args)
+            # @sg-ignore tool-limitation:is_a-narrowing:predicate-wrapper -
+            #   flow-sensitive typing only recognizes a literal is_a? call
+            #   on the guarded variable, not one wrapped in a helper
+            #   predicate method like Parser.is_ast_node? above. No
+            #   upstream issue filed yet.
```
apiology: This is not the sg-ignore naming standard for solargraph

PR #1324, opened 2026-08-21
Narrow is_a? checks joined by || to their union type

apiology:

**Problem:** `solargraph typecheck --level strong` reports a call as accepting the full, unnarrowed union of a variable's declared type, even inside an `if` branch that has already ruled out some of those types via `||`-joined `is_a?` checks.

```ruby
class Repro
  # @param e [Array<Symbol>, String]
  def eval_function_call(e); end

  # @param e [Array<Symbol>, String, Integer]
  def eval(e)
    if e.is_a?(Array) || e.is_a?(String)
      eval_function_call(e)  # reports e as Array<Symbol>, String, Integer
    end
  end
end
```

This produces a false `Wrong argument type` problem for any code relying on this common idiom to narrow a variable before use.

**Solution:** When both sides of an `||` are plain `is_a?` checks on the same variable, narrow that variable to the union of the two checked types inside the `if` branch; any other shape - different variables, or a side that isn't a plain `is_a?` call - is left unnarrowed, same as before.

**Known limitation:** a chain of three or more `is_a?` checks (`a.is_a?(X) || a.is_a?(Y) || a.is_a?(Z)`) still isn't narrowed, since the AST nests as `(X || Y) || Z` and the outer left side is an `:or` node rather than a `:send`. This degrades to the pre-existing no-narrowing behavior rather than narrowing incorrectly.

### DIFF
--- a/lib/solargraph/parser/flow_sensitive_typing.rb
+++ b/lib/solargraph/parser/flow_sensitive_typing.rb
@@ -65,7 +65,11 @@ module Solargraph
         # false, so provide false ranges to assert facts on
 
         # can't assume if an or is true that every single condition is
-        # true, so don't provide true ranges to assert facts on
+        # true, so don't provide true ranges to assert facts on to
+        # each side individually - however, if both sides are
+        # `is_a?` checks on the same variable, we can narrow that
+        # variable to the union of the checked types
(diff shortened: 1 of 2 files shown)


apiology:
Opened against the wrong repo (castwide/solargraph instead of apiology/solargraph). Reopening correctly.

PR #1325, opened 2026-08-21
Fix splat target in multiple assignment losing its element type

apiology:

A splat target in a multiple assignment loses its inferred type entirely:

```ruby
class Repro
  # @param mutator [Array<Symbol, BasicObject>]
  # @return [void]
  def call(mutator)
    command, *args = mutator
    args.fetch(0)  # Unresolved call to fetch
  end
end
```

**Cause:** `MasgnNode#process` looks up the pin for each left-hand-side entry by matching its node location against existing local/instance-variable pins. For a splat entry (`*args`, parsed as `s(:splat, s(:lvasgn, :args))`), the code used the *splat node's* location (`*args`) instead of the inner assignment target's location (`args`). These never match, so `args` never gets a `mass_assignment` set on it and stays permanently untyped.

**Fix:**
- Unwrap `:splat` nodes to their inner assignment target before computing the lookup location (and skip anonymous splats, which have no target).
- Track whether an entry is a splat alongside its right-hand-side node and index.
- In `BaseVariable#probe`, a splat entry now collects the union of the source array's element parameter types and binds `Array<...>` of that union, rather than binding a single (and previously entirely absent) element type.

Non-splat multiple assignment (`a, b = pair`) is untouched — this repo's `ComplexType::UniqueType#tuple?` is currently disabled (hard-coded `false`, added deliberately in `#1201` "Ignore literal values in type inference"), so positional/tuple-accurate element typing isn't implemented for any multi-assignment position, splat or not; that limitation is out of scope here.

**Test plan:**
- [x] `bundle exec rspec spec/pin/base_variable_spec.rb` — new specs for splat and non-splat masgn cases
- [x] `bundle exec rspec` — full suite, 0 failures
- [x] `bin/solargraph typecheck --level strong` on repro — 0 problems (previously "Unresolved call to fetch")
- [x] `bundle exec rubocop` on changed files — clean

### DIFF
--- a/lib/solargraph/parser/parser_gem/node_processors/masgn_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/masgn_node.rb
@@ -33,11 +33,26 @@ module Solargraph
             process_children
 
             lhs_arr.each_with_index do |lhs, i|
-              location = get_node_location(lhs)
-              pin = if lhs.type == :lvasgn
+              # A splat entry (e.g. `*args`) wraps the actual
+              # assignment target - `s(:splat, s(:lvasgn, :args))` -
+              # and the splat node's own location (which includes
+              # the leading `*`) never matches the location of the
+              # pin created for the inner assignment target. Unwrap
+              # it so the pin lookup below finds the right pin, and
+              # remember that this entry captures the "rest" of the
+              # right-hand side as an array rather than a single
+              # element.
+              splat = lhs.type == :splat
+              target = splat ? lhs.children[0] : lhs
+              # An anonymous splat (`a, * = arr`) has no assignment
+              # target to bind a type to.
+              next if target.nil?
+
+              location = get_node_location(target)
+              pin = if target.type == :lvasgn
                       # lvasgn is a local variable
                       locals.find { |l| l.location == location }
-                    elsif lhs.type == :ivasgn
+                    elsif target.type == :ivasgn
                       # ivasgn is an instance variable assignment
                       ivars.find { |iv| iv.location == location }
                     else
@@ -47,11 +62,11 @@ module Solargraph
               #   when a non-existant method is called on 'l'
               if pin.nil?
                 Solargraph.logger.debug do
-                  "Could not find local for masgn= value in location #{location.inspect} in #{lhs_arr} - masgn = #{masgn}, lhs.type = #{lhs.type}"
+                  "Could not find local for masgn= value in location #{location.inspect} in #{lhs_arr} - masgn = #{masgn}, target.type = #{target.type}"
                 end
                 next
               end
-              pin.mass_assignment = [mass_rhs, i]
+              pin.mass_assignment = [mass_rhs, i, splat]
             end
           end
         end
--- a/lib/solargraph/parser/parser_gem/node_processors/masgn_node.rb
+++ b/lib/solargraph/parser/parser_gem/node_processors/masgn_node.rb
@@ -33,15 +33,8 @@ module Solargraph
             process_children
 
             lhs_arr.each_with_index do |lhs, i|
-              # A splat entry (e.g. `*args`) wraps the actual
-              # assignment target - `s(:splat, s(:lvasgn, :args))` -
-              # and the splat node's own location (which includes
-              # the leading `*`) never matches the location of the
-              # pin created for the inner assignment target. Unwrap
-              # it so the pin lookup below finds the right pin, and
-              # remember that this entry captures the "rest" of the
-              # right-hand side as an array rather than a single
-              # element.
+              # A splat (`*args`) wraps its target - `s(:splat, s(:lvasgn, :args))` -
+              # at a location that won't match the inner pin's, so unwrap it.
               splat = lhs.type == :splat
               target = splat ? lhs.children[0] : lhs
               # An anonymous splat (`a, * = arr`) has no assignment
--- a/lib/solargraph/pin/base_variable.rb
+++ b/lib/solargraph/pin/base_variable.rb
@@ -191,7 +191,7 @@ module Solargraph
             # take the source array's full parameter union rather
             # than a single indexed/first parameter.
             types = types.flat_map do |type|
-              if type.tuple? || ['::Array', '::Set', '::Enumerable'].include?(type.rooted_name)
+              if type.tuple? || splattable?(api_map, type)
                 type.all_params
               else
                 []
@@ -201,7 +201,7 @@ module Solargraph
             types.map! do |type|
               if type.tuple?
                 type.all_params[index]
-              elsif ['::Array', '::Set', '::Enumerable'].include?(type.rooted_name)
+              elsif splattable?(api_map, type)
                 type.all_params.first
               end
             end.compact!
@@ -317,6 +317,21 @@ module Solargraph
       attr_writer :presence
 
       private
+
+      # True for a type whose elements can be destructured by a splat
+      # or positional multiple-assignment target - Array/Set/Enumerable
+      # by name, or any type that structurally conforms to Enumerable
+      # (a user-defined class that includes it, Hash, Range, etc).
+      #
+      # @param api_map [ApiMap]
+      # @param type [ComplexType::UniqueType]
+      # @return [Boolean]
+      def splattable? api_map, type
+        return true if ['::Array', '::Set', '::Enumerable'].include?(type.rooted_name)
+
+        enumerable = ComplexType.parse('::Enumerable').qualify(api_map)
+        type.qualify(api_map).conforms_to?(api_map, enumerable, :assignment)
+      end
 
       # @param api_map [ApiMap]
       # @param raw_return_type [ComplexType, ComplexType::UniqueType]
--- a/lib/solargraph/pin/base_variable.rb
+++ b/lib/solargraph/pin/base_variable.rb
@@ -19,7 +19,7 @@ module Solargraph
       #   that was made to this variable
       # @param assignments [Array<Parser::AST::Node>] Possible
       #   assignments that may have been made to this variable
-      # @param mass_assignment [::Array(Parser::AST::Node, Integer), nil]
+      # @param mass_assignment [::Array(Parser::AST::Node, Integer, Boolean), nil]
       # @param assignment [Parser::AST::Node, nil] First assignment
       #   that was made to this variable
       # @param assignments [Array<Parser::AST::Node>] Possible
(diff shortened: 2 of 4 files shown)


spec/pin/base_variable_spec.rb
```diff
@@ -47,6 +47,44 @@ def bar
...
+          command, *args = mutator
+        end
+      end
+    ), 'test.rb')
+    api_map = Solargraph::ApiMap.new
+    api_map.map source
+    locals = api_map.source_map('test.rb').locals
+    command_pin = locals.find { |l| l.name == 'command' }
+    args_pin = locals.find { |l| l.name == 'args' }
+    expect(command_pin.probe(api_map).tag).to eq('Symbol')
```
apiology, line 65: Why just Symbol here but not below?  If the other type is there is expected to be in the union, we should look for it; if it's not, we should assert it isn't there.

lib/solargraph/pin/base_variable.rb
```diff
@@ -183,19 +183,42 @@ def probe api_map
...
-              type.all_params[index]
-            elsif ['::Array', '::Set', '::Enumerable'].include?(type.rooted_name)
-              type.all_params.first
+          if splat
+            # A splat target (`*args`) captures every remaining
+            # element of the right-hand side, not just one position -
+            # take the source array's full parameter union rather
+            # than a single indexed/first parameter.
+            types = types.flat_map do |type|
+              if type.tuple? || ['::Array', '::Set', '::Enumerable'].include?(type.rooted_name)
```
apiology: If I create a class implementing Enumerable, can I splat it?

lib/solargraph/pin/base_variable.rb
apiology, line 171: The last element of the tuple in the return type is Boolean - is that what 'result' is here?

spec/pin/base_variable_spec.rb
apiology: **Claude:**

Real bug, not just a test inconsistency. `Array<Integer, String>` (no
tuple parens) declares a union of possible element types, not a
positional tuple - so `a` and `b` should both resolve to `Integer,
String`, not distinct per-position types. The code was taking
`all_params.first` unconditionally instead of the whole union.

Fixed in 6c302b28: the non-splat branch now takes the full union for a
non-tuple Array-like type, keeping the positional lookup only for an
actual tuple. Rewrote this spec (renamed to "gives every non-splat target
the full element union of a non-tuple Array") asserting `a` and `b` both
resolve to `Integer, String`, via `#to_s` rather than `#tag` -
`ComplexType#tag` delegates to the first union member only, so it
would have silently passed even with the union collapsed to `Integer`.

One caveat found while fixing this: `ComplexType::UniqueType#tuple?` is
currently dead code - `return false` unconditionally, before its real
logic - from upstream commit
[8c40692](https://github.com/castwide/solargraph/commit/8c406923760c908748968db4d87bbac00758b2cf)
(#1201), unrelated to this PR. So an actual positional tuple
(`::Array(Integer, String)`) can't be distinguished from a union-typed
Array right now; both take the union path. Not fixing that here - it is
outside the scope of this PR's own change and needs its own fix upstream.
On the decision list.

lib/solargraph/pin/base_variable.rb
apiology: **Claude:**

No, not before this change - the check was an exact-name allowlist
(`['::Array', '::Set', '::Enumerable'].include?(type.rooted_name)`), so a
custom class that merely includes Enumerable was never recognized, and
splatting it silently produced `undefined`.

Fixed in efdd3bc: added a `splattable?` helper that keeps the fast-path
name check and falls back to `type.conforms_to?(api_map, Enumerable,
:assignment)`, so a user-defined Enumerable class (or Hash, Range, etc)
is now recognized too. Added a spec with a generic-parameterized custom
Enumerable class confirming both the splat and non-splat targets resolve
correctly.

lib/solargraph/pin/base_variable.rb
apiology: **Claude:**

No - `result` here is whatever `chain.infer(...).self_to_type(...)`
resolves to for one value-position node in the assignment's RHS (a plain
`ComplexType`, not a tuple). Neither `Chain#infer` nor `#self_to_type` is
tuple-typed.

The Boolean-ending tuple in this file is a different thing: the
`@mass_assignment` accessor a few lines up is typed `[::Array(Parser::AST::Node,
Integer, Boolean), nil]` - that's `[rhs_node, index, splat_flag]`, set by
`MasgnNode#process`. Its Boolean is whether this target is a splat
(`*args`), not a value type - unrelated to what `result`/`chain.infer`
returns for a value node. If that is what the comment meant, the answer
is: no, they are unrelated - `result` is a value type, the tuple's
Boolean is a parsing-time flag about the assignment target's shape.

spec/pin/base_variable_spec.rb
```diff
@@ -47,6 +47,95 @@ def bar
...
+    expect(rest_pin.probe(api_map).tag).to eq('Array<String>')
+  end
+
+  it 'gives every non-splat target the full element union of a non-tuple Array' do
+    # Array<Integer, String> (no tuple parens) declares a union of
+    # possible element types, not a positional tuple, so every
+    # position - a and b alike - can hold either member. Note: #tag
+    # delegates to the first union member only, so a multi-member
+    # union must be asserted with #to_s instead, or a regression that
+    # collapses the union to its first member would still pass here.
```
apiology: 1 line at most

lib/solargraph/parser/parser_gem/node_processors/masgn_node.rb
```diff
@@ -33,11 +33,26 @@ def process
...
-              pin = if lhs.type == :lvasgn
+              # A splat entry (e.g. `*args`) wraps the actual
+              # assignment target - `s(:splat, s(:lvasgn, :args))` -
+              # and the splat node's own location (which includes
+              # the leading `*`) never matches the location of the
+              # pin created for the inner assignment target. Unwrap
+              # it so the pin lookup below finds the right pin, and
+              # remember that this entry captures the "rest" of the
+              # right-hand side as an array rather than a single
+              # element.
```
apiology: 2 lines at most

lib/solargraph/pin/base_variable.rb
```diff
@@ -19,7 +19,7 @@ class BaseVariable < Base
       #   that was made to this variable
       # @param assignments [Array<Parser::AST::Node>] Possible
       #   assignments that may have been made to this variable
-      # @param mass_assignment [::Array(Parser::AST::Node, Integer), nil]
+      # @param mass_assignment [::Array(Parser::AST::Node, Integer, Boolean), nil]
```
apiology: Please add 2 lines of documentation on this - seems complex

lib/solargraph/pin/base_variable.rb
```diff
@@ -183,19 +183,59 @@ def probe api_map
...
+            types = types.flat_map do |type|
+              if type.tuple?
+                # A true tuple's positions are heterogeneous, so only
+                # the position actually being assigned applies here.
+                [type.all_params[index]].compact
+              elsif splattable?(api_map, type)
+                # A non-tuple Array-like type (e.g. Array<Integer,
+                # String>) declares a union of possible element types,
+                # not a positional tuple - every position can hold any
+                # member of that union, so the whole union applies.
```
apiology: 1 line at most

lib/solargraph/pin/base_variable.rb
```diff
@@ -183,19 +183,59 @@ def probe api_map
...
+              else
+                []
+              end
             end
-          end.compact!
+          else
+            types = types.flat_map do |type|
+              if type.tuple?
+                # A true tuple's positions are heterogeneous, so only
+                # the position actually being assigned applies here.
```
apiology: Reword?  I don't understand

lib/solargraph/pin/base_variable.rb
```diff
@@ -295,6 +335,21 @@ def visible_at? other_closure, other_loc
 
       private
 
+      # True for a type whose elements can be destructured by a splat
+      # or positional multiple-assignment target - Array/Set/Enumerable
+      # by name, or any type that structurally conforms to Enumerable
+      # (a user-defined class that includes it, Hash, Range, etc).
```
apiology: 2 lines at most

spec/pin/base_variable_spec.rb
apiology: **Claude:** Cut to one line in 5d1ab4495:

```ruby
    # #to_s, not #tag, so a regression collapsing the union to one member fails here.
```

lib/solargraph/parser/parser_gem/node_processors/masgn_node.rb
apiology: **Claude:** Cut to two lines in 5d1ab4495:

```ruby
              # A splat (`*args`) wraps its target - `s(:splat, s(:lvasgn, :args))` -
              # at a location that won't match the inner pin's, so unwrap it.
```

lib/solargraph/pin/base_variable.rb
apiology: **Claude:** Cut from four lines to one in b79463d03:

```ruby
              elsif splattable?(api_map, type)
                # Array<Integer, String> is a union: any position holds either.
```

lib/solargraph/pin/base_variable.rb
apiology: **Claude:** Reworded in b79463d03 to name the type form it is about, so it pairs with the union comment on the `elsif` two lines down:

```ruby
              if type.tuple?
                # Array(Integer, String) gives each position its own type.
```

One thing worth knowing while reading this branch: it cannot execute today. `ComplexType#tuple?` returns before its body, a shim added by 8c406923 ("Ignore literal values in type inference", #1201) and removed again by the open #1223:

```ruby
def tuple?
  return false
  @tuple ||= (name == 'Tuple') || (name == 'Array' && subtypes.length >= 1 && fixed_parameters?)
end
```

Measured on this branch: a parameter typed `Array(Integer, String)` parses with `fixed_parameters?` true and `all_params == ["Integer", "String"]`, yet `tuple?` is false, so all three targets fall through to the union branch. That is also why undercover reports this block at 83.33 percent, with `[type.all_params[index]].compact` at zero hits.

94772eb42 adds a pending spec asserting the positional result. It fails today with `got: "Integer, String"` and will break the build the day #1223 lands, which is the signal to delete the `pending` line.

lib/solargraph/pin/base_variable.rb
apiology: **Claude:** Cut to two lines in 5d1ab4495:

```ruby
        # True for a type destructurable by a splat/multi-assignment target -
        # Array/Set/Enumerable by name, or anything structurally Enumerable.
```

PR #1326, opened 2026-08-21
Fix unresolved method-level generic block parameters

apiology:

## Problem

Enumerable#each_with_object is typed in core RBS as:

    [U] (obj U) { [U] (arg_0 Elem, obj U) -> untyped } -> U

`U` is a method-level generic bound by the `obj` argument passed to `each_with_object`. Inside the block, `arg_0` (typed `Elem`, the receiver's own generic) resolves correctly, but `obj` (typed `U`) resolves to no type at all:

```ruby
class Repro
  # @return [void]
  def scratch
    # @type [Array<Integer>]
    arr = [1, 2, 3]
    arr.each_with_object({}) { |e, memo| memo[e] = e }
  end
end
```

`solargraph typecheck --level strong` reports `Unresolved call to []=` on `memo`. The same shape of bug affects any method with a method-level generic bound via an argument passed alongside a block, e.g. `Enumerable#inject`.

## Root cause

`Pin::Callable#resolve_generics_from_context` resolves a method-level generic from the method's own argument types into a shared `resolved_generic_values` hash, then recurses into the yielded block's parameter list to resolve that signature too. That recursive call is made with no per-parameter argument types available (`nil`), and the code took a shortcut for that case: it `dup`'d each block parameter instead of running it through `resolve_generics_from_context`. So even though `resolved_generic_values["U"]` was already populated from the method's own `obj` argument, the block's own `obj` parameter (typed `generic<U>`) never picked it up.

## Fix

Always route block parameters through `resolve_generics_from_context`, passing `arg_types&.[](i)` instead of skipping resolution when `arg_types` is nil. `ComplexType::UniqueType#resolve_generics_from_context` already handles a nil context type by falling back to `resolved_generic_values[type_param] || self`, so this is a pure bug fix with no behavior change for the case where argument types are available.

## Testing

Added spec cases in `spec/source_map/clip_spec.rb` covering `each_with_object`, `inject`, and a negative control (`each_with_index`, which has no method-level generic in its block signature and must be unaffected). Full suite: 0 failures.

### DIFF
--- a/lib/solargraph/pin/callable.rb
+++ b/lib/solargraph/pin/callable.rb
@@ -143,13 +143,17 @@ module Solargraph
                                         resolved_generic_values: {}
         callable = super(generics_to_resolve, return_type_context, resolved_generic_values: resolved_generic_values)
         callable.parameters = callable.parameters.each_with_index.map do |param, i|
-          if arg_types.nil?
-            param.dup
-          else
-            param.resolve_generics_from_context(generics_to_resolve,
-                                                arg_types[i],
-                                                resolved_generic_values: resolved_generic_values)
-          end
+          # Even when we have no argument type to bind a generic
+          # *from* at this level (arg_types is nil, e.g. a yielded
+          # block's own parameter list), a method-level generic used
+          # here may already have been bound from context elsewhere
+          # (e.g. the enclosing method call's own arguments) and
+          # recorded in resolved_generic_values. Still route through
+          # resolve_generics_from_context so that value gets applied,
+          # rather than leaving the parameter as an unresolved generic.
+          param.resolve_generics_from_context(generics_to_resolve,
+                                              arg_types&.[](i),
+                                              resolved_generic_values: resolved_generic_values)
         end
         if callable.block?
           callable.block = block.resolve_generics_from_context(generics_to_resolve,
--- a/lib/solargraph/pin/callable.rb
+++ b/lib/solargraph/pin/callable.rb
@@ -143,14 +143,12 @@ module Solargraph
                                         resolved_generic_values: {}
         callable = super(generics_to_resolve, return_type_context, resolved_generic_values: resolved_generic_values)
         callable.parameters = callable.parameters.each_with_index.map do |param, i|
-          # Even when we have no argument type to bind a generic
-          # *from* at this level (arg_types is nil, e.g. a yielded
-          # block's own parameter list), a method-level generic used
-          # here may already have been bound from context elsewhere
-          # (e.g. the enclosing method call's own arguments) and
-          # recorded in resolved_generic_values. Still route through
-          # resolve_generics_from_context so that value gets applied,
-          # rather than leaving the parameter as an unresolved generic.
+          # arg_types[i] may be nil (e.g. a yielded block's own
+          # parameter list has no argument type to bind from), but a
+          # method-level generic can already be bound in
+          # resolved_generic_values from elsewhere (the enclosing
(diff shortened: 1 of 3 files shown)


lib/solargraph/pin/callable.rb
```diff
@@ -143,13 +143,17 @@ def resolve_generics_from_context generics_to_resolve,
...
-                                                resolved_generic_values: resolved_generic_values)
-          end
+          # Even when we have no argument type to bind a generic
+          # *from* at this level (arg_types is nil, e.g. a yielded
+          # block's own parameter list), a method-level generic used
+          # here may already have been bound from context elsewhere
+          # (e.g. the enclosing method call's own arguments) and
+          # recorded in resolved_generic_values. Still route through
+          # resolve_generics_from_context so that value gets applied,
+          # rather than leaving the parameter as an unresolved generic.
```
apiology: Is this helpful documentation or changelog?  Seems very wordy for the former.

spec/source_map/clip_spec.rb
```diff
@@ -1856,6 +1856,56 @@ def foo; end
...
+      a = [1, 2, 3]
+      a.each_with_index do |e, i|
+        e
+      end
+    ), 'test.rb')
+    api_map = Solargraph::ApiMap.new.map(source)
+    clip = api_map.clip_at('test.rb', [4, 8])
+    type = clip.infer
+    expect(type.to_s).to eq('Integer')
+  end
```
apiology: Instead of adding more to a very long clip_spec.rb, can we find a more specific please for these - maybe callable_spec.rb?

lib/solargraph/pin/callable.rb
apiology: **Claude:**

Trimmed to the constraint itself in 8d0192b:

```
- # Even when we have no argument type to bind a generic
- # *from* at this level (arg_types is nil, e.g. a yielded
- # block's own parameter list), a method-level generic used
- # here may already have been bound from context elsewhere
- # (e.g. the enclosing method call's own arguments) and
- # recorded in resolved_generic_values. Still route through
- # resolve_generics_from_context so that value gets applied,
- # rather than leaving the parameter as an unresolved generic.
+ # arg_types[i] may be nil (e.g. a yielded block's own
+ # parameter list has no argument type to bind from), but a
+ # method-level generic can already be bound in
+ # resolved_generic_values from elsewhere (the enclosing
+ # call's own arguments) - resolve_generics_from_context
+ # still applies that binding.
```

spec/source_map/clip_spec.rb
apiology: **Claude:**

Moved all 3 to a new spec/pin/callable_spec.rb in 8d0192b.

lib/solargraph/pin/callable.rb
```diff
@@ -143,13 +143,15 @@ def resolve_generics_from_context generics_to_resolve,
...
-            param.resolve_generics_from_context(generics_to_resolve,
-                                                arg_types[i],
-                                                resolved_generic_values: resolved_generic_values)
-          end
+          # arg_types[i] may be nil (e.g. a yielded block's own
+          # parameter list has no argument type to bind from), but a
+          # method-level generic can already be bound in
+          # resolved_generic_values from elsewhere (the enclosing
+          # call's own arguments) - resolve_generics_from_context
+          # still applies that binding.
```
apiology: 1 line at most - better yet, explain it better by assigning it to a local variable with a better name.

PR #1328, opened 2026-08-22
Synthesize a Struct.new constructor pin for gem-sourced constants

apiology:
## Summary

`Foo = Struct.new(:bar, :baz)` (and `class Foo < Struct.new(...)`) sourced from a gem, not workspace code -- YARD's Ruby handler registers `Foo` as a real `ClassObject` with accessor pins for each member, but never documents an `initialize`. `Solargraph::YardMap::Mapper` has no constructor pin to combine, so `Foo.new(...)` falls back to resolving against `Struct.new`'s own RBS signature -- the nearest ancestor method it can find -- and reports a wrong-argument-type error for ordinary positional construction.

`Convention::StructDefinition` already covers this for workspace source, building an `initialize` pin from the parsed `Struct.new` node. This adds the gem/yardoc equivalent (`ToStructInitializer`), where there's no such node -- only the member names YARD kept in `ClassObject#attributes`.

`keyword_init: true` is detected by rereading the source line YARD points at (YARD itself discards the flag), first detecting the file's declared magic-comment encoding the way Ruby's own parser does, so a gem source line declared e.g. `ISO-8859-1` decodes correctly rather than being misread as the default external encoding.

## Test plan
- [x] 8 specs in `spec/yard_map/mapper/to_struct_initializer_spec.rb`, including the `ISO-8859-1` encoding case and one genuinely-unrecoverable case (declared encoding doesn't match actual bytes)
- [x] `bundle exec rspec` (full suite): 1633 examples, 1 pre-existing failure (confirmed unrelated via stash A/B)
- [x] `bundle exec rubocop`: no offenses

**Claude:** filed on behalf of the repo owner while auditing `@sg-ignore` suppressions in a downstream project.

### DIFF
--- a/lib/solargraph/yard_map/mapper.rb
+++ b/lib/solargraph/yard_map/mapper.rb
@@ -42,7 +42,6 @@ module Solargraph
           nspin = ToNamespace.make(code_object, @spec, @namespace_pins[code_object.namespace.to_s])
           @namespace_pins[code_object.path] = nspin
           result.push nspin
-          # See ToStructInitializer for why a Struct.new definition needs this.
           result.concat ToStructInitializer.make(code_object, nspin, @spec)
           if code_object.is_a?(YARD::CodeObjects::ClassObject) && !code_object.superclass.nil?
             # This method of superclass detection is a bit of a hack. If
--- a/lib/solargraph/yard_map/mapper.rb
+++ b/lib/solargraph/yard_map/mapper.rb
@@ -6,6 +6,7 @@ module Solargraph
       autoload :ToMethod, 'solargraph/yard_map/mapper/to_method'
       autoload :ToNamespace, 'solargraph/yard_map/mapper/to_namespace'
       autoload :ToConstant, 'solargraph/yard_map/mapper/to_constant'
+      autoload :ToStructInitializer, 'solargraph/yard_map/mapper/to_struct_initializer'
 
       # @param code_objects [Array<YARD::CodeObjects::Base>]
       # @param spec [Gem::Specification, nil]
@@ -41,6 +42,8 @@ module Solargraph
           nspin = ToNamespace.make(code_object, @spec, @namespace_pins[code_object.namespace.to_s])
           @namespace_pins[code_object.path] = nspin
           result.push nspin
+          # See ToStructInitializer for why a Struct.new definition needs this.
+          result.concat ToStructInitializer.make(code_object, nspin, @spec)
           if code_object.is_a?(YARD::CodeObjects::ClassObject) && !code_object.superclass.nil?
             # This method of superclass detection is a bit of a hack. If
             # the superclass is a Proxy, it is assumed to be undefined in its
--- a/lib/solargraph/yard_map/mapper/to_struct_initializer.rb
+++ b/lib/solargraph/yard_map/mapper/to_struct_initializer.rb
@@ -1,0 +1,127 @@
+# frozen_string_literal: true
+
+module Solargraph
+  class YardMap
+    class Mapper
+      # Synthesizes the constructor pins YARD never generates for a
+      # `Foo = Struct.new(:bar, :baz)` (or `class Foo < Struct.new(...)`)
+      # definition it documents.
+      #
+      # YARD's own Ruby handlers for `Struct.new`
+      # (`YARD::Handlers::Ruby::ConstantHandler#process_structclass`,
+      # `YARD::Handlers::Ruby::ClassHandler`) register the struct as a real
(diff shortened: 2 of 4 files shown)


lib/solargraph/yard_map/mapper.rb
```diff
@@ -41,6 +42,8 @@ def generate_pins code_object
           nspin = ToNamespace.make(code_object, @spec, @namespace_pins[code_object.namespace.to_s])
           @namespace_pins[code_object.path] = nspin
           result.push nspin
+          # See ToStructInitializer for why a Struct.new definition needs this.
```
apiology: Does this comment meet the bar of adding clarity?

lib/solargraph/yard_map/mapper/to_struct_initializer.rb
```diff
@@ -0,0 +1,127 @@
...
+            Solargraph.logger.info "Could not check #{path}:#{line} for keyword_init: [#{e.class}] #{e.message}"
+            false
+          end
+
+          # @param path [String]
+          # @return [Encoding]
+          # @sg-ignore `Encoding.find`'s RBS signature allows a nil return
+          #   that Ruby's own docs say never happens in practice; the `||`
+          #   below already covers it, but flow-sensitive typing doesn't
+          #   narrow it out of this method's inferred overall return type.
```
apiology: Add reference URLs?

lib/solargraph/yard_map/mapper.rb
apiology: **Claude:** No - removed in 81abbae62. "See ToStructInitializer for why..." adds nothing at the call site a reader wouldn't already get from the method call itself; ToStructInitializer.make already says what it does, and its own module docstring already covers the why.

lib/solargraph/yard_map/mapper/to_struct_initializer.rb
apiology: **Claude:** Traced to ground truth, no existing upstream issue found to link. Confirmed precisely: `YARD::CodeObjects::Base#line` is documented `@return [Fixnum]` in YARD's own source (predates Ruby 2.4's Fixnum/Integer merge), and `solargraph pin 'Fixnum'` resolves to a namespace pin sourced from `diff-lcs-1.6.2/lib/diff/lcs/backports.rb` - a legacy compat shim with no methods and no registered Integer relationship, so `Fixnum#-` has no pin at all. This predates this PR (both the YARD typing and diff-lcs's shim already exist independent of it); this PR only newly exercises it via `line - 1`.

Not filing a GitHub issue unilaterally - added to the decision list instead, since it's worth deciding whether to file against Fixnum specifically or against the broader pattern (any gem documenting a legacy-Ruby-compat shim can shadow a core RBS type this way). Will add the URL to the ignore once that's decided.

spec/yard_map/mapper/to_struct_initializer_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  it 'falls back to positional when the declared encoding does not match the actual bytes' do
+    Dir.mktmpdir do |dir|
+      path = File.join(dir, 'struct_def.rb')
+      # The comment claims US-ASCII, but \xE9 is not a valid US-ASCII byte --
+      # a genuinely unrecoverable case, not merely one File.readlines
+      # defaults its way past. Exercised directly against #keyword_init?,
+      # since a file this malformed is unlikely to reach YARD as a real
+      # ClassObject in the first place.
```
apiology: Shouldn't the point of the test be documented in the 'it' statement above, not in a buried comment?

spec/yard_map/mapper/to_struct_initializer_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+  it 'honors a magic encoding comment when rereading the definition line' do
+    Dir.mktmpdir do |dir|
+      path = File.join(dir, 'struct_def.rb')
+      # File.readlines defaults to Encoding.default_external (UTF-8) and
+      # does not itself honor a magic encoding comment the way compiling
+      # the file would -- \xE9 is only valid as "e" line data under
+      # ISO-8859-1. YARD parses this fine (it resolves the same comment),
+      # so #keyword_init? has to detect and honor it too, not just survive
+      # not honoring it.
```
apiology: See below

lib/solargraph/yard_map/mapper/to_struct_initializer.rb
apiology: >>> that Ruby's own docs say never happens in practice

>> Add reference URLs?

> no existing upstream issue found to link

who said issue?

PR #1329, opened 2026-08-23
Fix generic arguments lost through inherited methods

apiology:
**Claude:**

A class reference with a concrete generic argument doesn't propagate that argument through a method inherited from its superclass:

```
$ bundle exec solargraph pin 'NoMethodError<String>#receiver' --rbs
untyped
```

`NoMethodError<T> < NameError<T>`, and `NameError#receiver` is declared to return `T`. Given `NoMethodError<String>`, `#receiver` should resolve to `String` (or `String | nil`) — instead it erases to `untyped`.

Root cause: `UniqueType#resolve_generics` erased a class's own generic type parameter to `ComplexType::UNDEFINED` whenever the querying context supplied no concrete value and no default existed, rather than leaving it as an unbound placeholder. `resolve_generics` is called twice on this path (`ApiMap#inner_get_methods_from_reference`) — once resolving the superclass reference itself, once resolving the returned method pins — and the fallback nuked the intermediate value before the second call could use it.

After the fix:

```
$ bundle exec solargraph pin 'NoMethodError<String>#receiver' --rbs
(::String | nil)
```

A bare, unrooted query (`NoMethodError#receiver`, no type args) still shows `untyped` via the CLI `pin` command — that comes from `ApiMap#erase_generics`, a separate, deliberate policy (see `spec/source_map/clip_spec.rb`'s "erases unresolvable class generics" example), and is unaffected by this change.

Full spec suite: 1624 examples, 0 failures. Rubocop: 2 pre-existing offenses, both outside this diff (`Lint/UnreachableCode` at lines 151 and 379, unrelated to the change at line ~466).

### DIFF
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -463,14 +463,8 @@ module Solargraph
             elsif idx.zero? && !context_type.all_params.empty?
               ComplexType.new(context_type.all_params)
             else
-              # No concrete value was supplied for this type param (e.g.
-              # a bare reference to a generic class/superclass with no
-              # type arguments) and it has no default. Leave it as an
-              # unbound generic placeholder rather than erasing it to
-              # untyped -- the same type param queried directly on its
-              # declaring class (with no reference/superclass involved)
-              # is never substituted at all and stays unbound the same
-              # way.
+              # No concrete value was supplied; leave unbound rather than untyped,
+              # matching a direct (non-substituted) lookup on the declaring class.
               t
             end
           else
--- a/lib/solargraph/complex_type/unique_type.rb
+++ b/lib/solargraph/complex_type/unique_type.rb
@@ -463,7 +463,15 @@ module Solargraph
             elsif idx.zero? && !context_type.all_params.empty?
               ComplexType.new(context_type.all_params)
             else
-              ComplexType::UNDEFINED
+              # No concrete value was supplied for this type param (e.g.
+              # a bare reference to a generic class/superclass with no
+              # type arguments) and it has no default. Leave it as an
+              # unbound generic placeholder rather than erasing it to
+              # untyped -- the same type param queried directly on its
+              # declaring class (with no reference/superclass involved)
+              # is never substituted at all and stays unbound the same
+              # way.
+              t
             end
           else
             t
--- a/spec/source_map/clip_spec.rb
+++ b/spec/source_map/clip_spec.rb
@@ -1904,10 +1904,7 @@ describe Solargraph::SourceMap::Clip do
   end
 
   it 'propagates a concrete generic argument through an inherited method' do
-    # NoMethodError<T> < NameError<T>, and NameError#receiver
-    # returns T - given a concrete argument on the subclass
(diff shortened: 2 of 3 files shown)


lib/solargraph/complex_type/unique_type.rb
```diff
@@ -463,7 +463,15 @@ def resolve_generics definitions, context_type
...
-              ComplexType::UNDEFINED
+              # No concrete value was supplied for this type param (e.g.
+              # a bare reference to a generic class/superclass with no
+              # type arguments) and it has no default. Leave it as an
+              # unbound generic placeholder rather than erasing it to
+              # untyped -- the same type param queried directly on its
+              # declaring class (with no reference/superclass involved)
+              # is never substituted at all and stays unbound the same
+              # way.
+              t
```
apiology, line 468: Add a spec please that would show a reviewer the exact situation this assists with.

apiology: **Claude:**

Added a spec at `spec/complex_type_spec.rb`, right after the existing `resolves generic namespace parameters` example. It builds a `Box` namespace with two `@generic` params (`K`, `V`), references it bare (`Box`, no type args), and resolves `generic<V>`. Before this PR fix that returned `undefined`; with the fix it stays `generic<V>`.

Verified the spec actually discriminates by swapping in master's `unique_type.rb` and confirming it fails there with `expected: "generic<V>", got: "undefined"`, then restoring the fix and confirming it passes.

apiology: Higher level spec as well please.  You added a good example in the PR description - why not something closer to that?

apiology: **Claude:** Added in 9afc266da - a spec going through real workspace source and ApiMap with the exact NoMethodError<String>#receiver shape from the PR description, alongside the existing resolve_generics-level one.

lib/solargraph/complex_type/unique_type.rb
```diff
@@ -463,7 +463,15 @@ def resolve_generics definitions, context_type
...
             else
-              ComplexType::UNDEFINED
+              # No concrete value was supplied for this type param (e.g.
+              # a bare reference to a generic class/superclass with no
+              # type arguments) and it has no default. Leave it as an
+              # unbound generic placeholder rather than erasing it to
+              # untyped -- the same type param queried directly on its
+              # declaring class (with no reference/superclass involved)
+              # is never substituted at all and stays unbound the same
+              # way.
```
apiology: limit to 2 lines

spec/source_map/clip_spec.rb
```diff
@@ -1903,6 +1903,22 @@ def bar; baz; end
     expect(type.to_s).to eq('undefined')
   end
 
+  it 'propagates a concrete generic argument through an inherited method' do
+    # NoMethodError<T> < NameError<T>, and NameError#receiver
+    # returns T - given a concrete argument on the subclass
+    # reference, the inherited method should resolve it rather
+    # than erasing it to untyped/undefined.
```
apiology: 0-1 lines - no filler

PR #1330, opened 2026-08-25
Fix @!override with @overload not dispatching per call site

apiology:
**Claude:**

**Problem:** An `@!override` with two `@overload` tags only ever applies the first one, so Solargraph rejects calls that actually match the second.

```ruby
# @!override Passthrough#identity
#   @overload identity(arguments)
#     @param arguments [Array<Hash>]
#     @return [Array<Hash>]
#   @overload identity(arguments)
#     @param arguments [Array<String>]
#     @return [Array<String>]

Passthrough.identity(['str'])
# Solargraph flags this: expected Array<Hash>, got Array<String>
# Wrong -- the second @overload matches, but only the first ever applies
```

On RBS-sourced pins like `Array#sum`, the override doesn't apply at all — Solargraph keeps using the original untyped signature.

**Solution:** Apply all of an override's tags to the docstring before invalidating `@signatures`/`@overloads` once, gated on the override actually carrying `@overload` tags, since an unconditional invalidate broke 34 unrelated specs that only redefine a return type via bare `@return`/`@param`.

Full suite green; one pre-existing test-order flake on `master`, unrelated to this change.

### DIFF
--- a/lib/solargraph/api_map/index.rb
+++ b/lib/solargraph/api_map/index.rb
@@ -195,14 +195,14 @@ module Solargraph
             end
             # pin.reset_generated! always runs (matches prior behavior) so
             # non-overload tags (e.g. @param) still re-typify against the
-            # updated docstring. The @signatures/@overloads ivars are only
-            # cleared in addition when the override carries @overload tags
-            # -- see #reset_overridden_signatures.
+            # updated docstring. #invalidate_signatures! is only called in
+            # addition when the override carries @overload tags -- see its
+            # doc comment on Pin::Method for why that's conditional.
             pin.reset_generated!
             new_pin&.reset_generated!
             if ovr.tags.any? { |tag| tag.tag_name == 'overload' }
-              reset_overridden_signatures pin
-              reset_overridden_signatures new_pin
+              pin.invalidate_signatures!
+              new_pin&.invalidate_signatures!
             end
             ovr.tags.each do |tag|
               redefine_return_type pin, tag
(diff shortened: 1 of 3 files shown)


spec/api_map/index_spec.rb
```diff
@@ -61,4 +61,116 @@
...
+      method_pin = output_pins.find { |pin| pin.path == 'Benchmark.measure' }
+      expect(method_pin.return_type.tag).to eq('Benchmark::Tms')
+    end
+
+    context 'when overriding a method that already had its signatures computed' do
+      it 'still redefines the return type' do
+        # Force memoization before the override applies, matching how a
+        # real consumer (e.g. solargraph-rspec's @!override annotations
+        # for ActionController::TestCase::Behavior#request/#response)
+        # can hit a pin whose signatures were already resolved.
```
apiology: 1 line at most

spec/api_map/index_spec.rb
```diff
@@ -61,4 +61,116 @@
...
+          passthrough,
+          identity,
+          identity_override
+        ]
+      end
+
+      it 'still applies every @overload tag, not zero of them' do
+        # Force memoization of the original (un-overridden) signatures,
+        # simulating an earlier pass over the ApiMap having already
+        # read #signatures before the override was applied.
```
apiology: 1 line at most

lib/solargraph/pin/method.rb
```diff
@@ -101,6 +101,21 @@ def reset_generated!
...
       end
 
+      # Discards memoized #signatures/#overloads so the next read rebuilds
+      # from the docstring -- needed when a caller mutates the docstring
+      # directly (e.g. applying `@!override`'s `@overload` tags to an
+      # already-cached pin) and needs the new tags to take effect.
+      #
+      # Only call this when `@overload` tags were added or removed: with
+      # none present, #signatures regenerates from this pin's own
+      # #parameters, which is empty or stale for many RBS/combined pins.
```
apiology: 3 lines at most

lib/solargraph/api_map/index.rb
```diff
@@ -183,22 +183,32 @@ def map_overrides
...
+            # @overload tag at once, so invalidating mid-loop would freeze
+            # it against a half-applied docstring and drop the rest.
             ovr.tags.each do |tag|
               pin.docstring.add_tag(tag)
+              new_pin&.docstring&.add_tag(tag)
+            end
+            # reset_generated! always runs, so non-overload tags (e.g.
+            # @param) still re-typify; invalidate_signatures! only runs in
+            # addition when the override has @overload tags -- see its doc
+            # comment on Pin::Method for why.
```
apiology: 1 line at most

lib/solargraph/api_map/index.rb
```diff
@@ -183,22 +183,32 @@ def map_overrides
               pin.docstring.delete_tags tag
               new_pin&.docstring&.delete_tags tag
             end
+            # Add all tags before invalidating: #signatures rebuilds every
+            # @overload tag at once, so invalidating mid-loop would freeze
+            # it against a half-applied docstring and drop the rest.
```
apiology: 1 line at most

apiology:
**Claude:**

Absorbed into #1104, which now carries these commits along with the docstring work they turned out to depend on.

The two changes are one change. An `@!override` can only refine an RBS signature set rather than replace it once `combine_with` is the join between type sources — landing this separately would have shipped a provenance check that #1104 makes unnecessary.

PR #1332, opened 2026-08-25
Resolve a module's ivar types from its includers

apiology:
**Claude:**

## Summary

`get_instance_variable_pins('Loggable')` returned `[]` for a module whose methods read an ivar only assigned by classes that include it — it only walked `ApiMap::Store#get_superclass`, never the include direction. This made a `@return`-declared return type on a mixin method that reads an instance variable uninferrable, even when every includer assigns that ivar consistently.

```ruby
module Loggable
  # @return [Logger]
  def log_message
    @logger  # return type could not be inferred
  end
end

class Widget
  include Loggable
  def initialize
    @logger = Logger.new($stdout)
  end
end
```

`Store#includer_map` inverts the existing `include_references` index (includer fqns -> included module refs) into module fqns -> includer fqns, closed transitively so a module included by another included module still resolves back to the original includer. `get_instance_variable_pins` now also walks each includer's own ivar pins (and their superclass chains) for the queried module, gated by `ApiMap#merge_includer_instance_variable_pins`: if two includers assign the same ivar name with conflicting probed types, both are excluded rather than picking one arbitrarily.

Cost: the gate probes every candidate pin's type on each query, so a module with many includers pays O(includers) probes per ivar lookup — did not move suite timing in testing.

## Test plan

- Repro above: return type could not be inferred before, resolves to `Logger` after.
- 3 new specs added to `spec/api_map_spec.rb` (single includer, transitive includer, conflicting-type exclusion) — all pass.
- `spec/api_map_spec.rb`: 70 examples, 0 failures, 4 pending.
- Rubocop clean on the diff (3 pre-existing offenses on `master` in these files, unrelated to this change, confirmed unchanged before/after).

### DIFF
--- a/lib/solargraph/api_map.rb
+++ b/lib/solargraph/api_map.rb
@@ -382,8 +382,27 @@ module Solargraph
     # @param scope [Symbol] :instance or :class
     # @return [Array<Solargraph::Pin::InstanceVariable>]
     def get_instance_variable_pins namespace, scope = :instance
+      result = own_instance_variable_pins(namespace, scope)
+
+      includer_fqns_list = store.get_includers(namespace)
+      unless includer_fqns_list.empty?
+        per_includer = includer_fqns_list.map { |includer_fqns| own_instance_variable_pins(includer_fqns, scope) }
+        result.concat merge_includer_instance_variable_pins(per_includer)
+      end
+
+      result
+    end
+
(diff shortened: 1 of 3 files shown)


apiology:
Closing ��� opened without review first, which shouldn't have happened. Not resurrecting without going through the sign-off step properly.

PR #1340, opened 2026-09-06
Serve RBS stdlib pins before the gem cache builds

apiology:
**Problem:** Until `solargraph gems` has run for a workspace, a gem with no combined pin cache entry contributes no pins at all, so completion and typechecking see none of its API — even where RBS already ships the types.

```ruby
# fresh workspace, logger not yet cached
dm = Solargraph::DocMap.new(["logger"], workspace, out: nil)
dm.pins.map(&:path).grep(/\ALogger#/).length
# => 0        # 34 are available from RBS, withheld until a build runs
```

On this machine that is 15 of 82 bundled gems, 1,305 pins, withheld until a build finishes.

**Solution:** Adds `RbsMap#fallback_pins` — nil on the base class, since a combined entry merges RBS pins with separately-cached YARD pins and dropping the YARD half is not safe in general, with `RbsMap::StdlibMap` overriding it to return its own pins, which need no merge — and serves it from `DocMap#deserialize_combined_pin_cache` where that previously returned nil, without caching it in the process-wide index or marking the gem built.

### DIFF
--- a/lib/solargraph/doc_map.rb
+++ b/lib/solargraph/doc_map.rb
@@ -231,7 +231,7 @@ module Solargraph
     end
 
     # @param gemspec [Gem::Specification]
-    # @return [void]
+    # @return [Array<Pin::Base>, nil]
     def deserialize_combined_pin_cache gemspec
       unless combined_pins_in_memory[[gemspec.name, gemspec.version]].nil?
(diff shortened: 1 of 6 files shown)


apiology:
**Claude:** Reopening this in [apiology/solargraph#98](https://github.com/apiology/solargraph/pull/98).

PR #1343, opened 2026-09-06
Pass the require path to find_by_path

apiology:
**Problem:** `Gem::Specification.find_by_path` matches a requirable path — the same string that appears in a `require` statement. `resolve_require` was handing it a lib-relative filename instead, which it can never match, so that lookup always returned nil.

```ruby
Gem::Specification.find_by_path("yard")         # => the yard gemspec
Gem::Specification.find_by_path("lib/yard.rb")  # => nil
```

Resolution then fell through to guessing the gem name from the require path and, failing that, scanning every bundled gemspec and reading its file list.

**Solution:** Pass the require path itself, and move the lib-relative filename down to the manual `spec.files` scan, which is the only remaining reader of it.

### DIFF
--- a/lib/solargraph/workspace/gemspecs.rb
+++ b/lib/solargraph/workspace/gemspecs.rb
@@ -44,8 +44,7 @@ module Solargraph
         return auto_required_gemspecs_from_bundler if require == 'bundler/require'
 
         # Determine gem name based on the require path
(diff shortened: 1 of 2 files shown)


castwide:
Closed as not planned. Path resolution is changing in https://github.com/castwide/solargraph/pull/1369.

PR #1346, opened 2026-09-07
Fix crash in solargraph gems core

apiology:
**Problem:** `solargraph gems core` crashes instead of caching core pins.

```
$ solargraph gems core
lib/solargraph/shell.rb:193:in `block in gems': undefined method `cache_core'
  for Solargraph::PinCache:Module (NoMethodError)
```

`cache_core` is an instance method on `RbsMap::CoreMap`; nothing defines it, or `core?`, on `PinCache`. The `@sg-ignore` above the call asserting that both are dynamically defined is simply wrong, so the marker was suppressing a genuine error rather than a tooling gap.

**Solution:** Call `RbsMap::CoreMap.new.cache_core` and drop the false marker.

### DIFF
--- a/lib/solargraph/shell.rb
+++ b/lib/solargraph/shell.rb
@@ -189,8 +189,7 @@ module Solargraph
         warn("Caching these gems: #{names}")
         names.each do |name|
           if name == 'core'
-            # @sg-ignore cache_core and core? are dynamically defined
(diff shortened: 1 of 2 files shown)


castwide:
Closed as not planned. This issue will be superseded by the transition to `Collection::Core` in https://github.com/castwide/solargraph/pull/1369.

